PR #2, opened 2019-09-20
Add dart support

Almoullim:
- No Flutter specific code included but technically, it should work for Flutter, dart-vm, and probably dart2js too

### Missing
- Testing for every API to make sure all of them function as expected
- Example of implementation
- Parameter types. Currently, all parameter types are set to `dynamic`
- `@required` for the required parameters of methods. Currently, all parameters are set to be optional.
- Automated tests?


eldadfux:
@Almoullim really great job, everything seems great! Would be great to get your feedback about how hard was it to onboard the project and whether the docs were good enough or should some thing be made clearer.

PR #12, opened 2019-10-01
Update phpunit.xml file

joaorobertopb:
This **PR** will:

- Remove legacy phpunit flag, **syntaxCheck="false"**:
  - That configuration setting has not had an effect in years. It was removed long ago.
- Rename **phpunit.xml** to **phpunit.xml.dist** and ignore **phpunit.xml** file:
  - This way we allow developers to keep a custom phpunit.xml file locally if they want to. This is quite common in PHP packages.


peter279k:
This PR contains conflict.

It should be concerned and resolved before merging PR.

Issue #9, opened 2019-09-29
Add C# Templates to the SDK Generator


eldadfux:
We need help adding C# templates to this project to allow easy generation of C# SDKs for Appwrite server.

A full guide on how to add new language support to the Appwrite SDK generator is located here:
https://github.com/appwrite/sdk-generator/blob/master/CONTRIBUTING.md

Kerl1310:
I'd like to get involved with this

eldadfux:
Hey @Kerl1310 that will be amazing! Please reach out to me on Gitter for any question or help!

Joe118:
I can also help.

Kerl1310:
Think this is finished, awaiting review

Joe118:
Cool. Where can I take a peek?

Joe118:
Actually I meant where can I take a look at the generated C# SDK for AppWrite lol

eldadfux:
Hey @Kerl1310 were the c# files written manually or using the twig templates? If manually I can help creating the twig templates, just let me know.

Also, could you please add a snippet example to show how to init the SDK and make an HTTP call?

eldadfux:
So there's some good news about this issue. @komemi have been working on it for the last few days and it seems close to a PR. We'll add more update to this issue when more progress is made.

eldadfux:
The .NET SDK first *experimental* release is out: https://github.com/appwrite/sdk-for-dotnet we'd love help with testing!

All credits to @komemi and @TorstenDittmann for working on this! Great work guys!!

PR #13, opened 2019-10-04
Adding CSharp template

Kerl1310:
#9 Adding CSharp template to the generator


CONTRIBUTING.md
```diff
@@ -42,7 +42,7 @@ Notice: The easiest way to get started is to copy an existing language class clo
 
 Add your new templates as listed in your language class **getFiles** method. Make sure to follow the [checklist](https://github.com/appwrite/sdk-generator#sdk-checklist) when building the language templates.
 
-Make sure to follow the objects structure and service separation architecture. We aim to keep developer experience as consisted as possible across different SDKs to make the learning curve as small as possible.
+Make sure to follow the objects structure and service separation architecture. We aim to keep developer experience as consistant as possible across different SDKs to make the learning curve as small as possible.
```
MellenIO: `consistant` -> `consistent`

Kerl1310: Nice catch! Shall update this shortly and commit

Kerl1310:
Fixed - renamed the REDAME.md files in js and ruby as README.md too

Kerl1310:
@eldadfux This is now ready

Kerl1310:
Ah! Missed that folder, going through and writing them now

Kerl1310:
@eldadfux Adding examples folder and pulled upstream master into my branch

eldadfux:
Hey @Kerl1310 could you please add a snippet example to show how to init the SDK and make an HTTP call?

examples/csharp/src/client.cs
```diff
@@ -0,0 +1,136 @@
...
+                }
+            }
+
+            foreach (KeyValuePair<string, string> kvp in _globalHeaders)
+            {
+                request.AddHeader(kvp.Key, kvp.Value);
+            }
+        }
+
+        if (paramsIn == null)
```
Joe118: So if the caller of Call() passes a null paramsIn, do you **have** to create an object and pass that to request.AddJsonBody? Does AddJsonBody cause an exception if paramsIn is null? Why even call it if paramsIn is null?

examples/csharp/src/services/account.cs
```diff
@@ -0,0 +1,91 @@
+using System.Collections.Generic;
+using Service;
+
+public class Account : Service
+{
+    public Get()
+    {
+        paramsIn = new Dictionary<string, string>();
```
Joe118: Where are these instance variables (paramsIn and path) defined? I don't see them in Account, or in Service, and Service has no base class, so... ?

examples/csharp/src/client.cs
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    public SetEndpoint(string endpoint)
+    {
+        _endpoint = endpoint;
+    }
+
+    public AddHeader(string key, string value)
+    {
+        if (_globalHeaders.ContainsKey(key.ToLower()))
```
Joe118: Maybe avoid calling ToLower() 1+ times? You could store the result in a local variable or overwrite the value of key?

examples/csharp/src/client.cs
```diff
@@ -0,0 +1,136 @@
...
+        }
+        else
+        {
+            _globalHeaders.Add(key, value.ToLower());
+        }
+    }
+
+    public SetProject(string value)
+    {
+        if (_globalHeaders.ContainsKey(key.ToLower()))
```
Joe118: Why not use AddHeader here and in every method below?

examples/csharp/src/client.cs
```diff
@@ -0,0 +1,136 @@
...
+                request.AddHeader(kvp.Key, kvp.Value);
+            }
+        }
+
+        if (paramsIn == null)
+        {
+            paramsIn = new Dictionary<string, string>();
+        }
+        
+        if (headers["Content-Type"] == "application/json")
```
Joe118: Does this header always exist?

examples/csharp/src/client.cs
```diff
@@ -0,0 +1,136 @@
...
+    {
+        var restClient = new RestClient(_endpoint);
+        var request = new RestRequest(path)
+        {
+            Method = method
+        };
+
+        if (headers != null && headers.Count > 0)
+        {
+            foreach (KeyValuePair<string, string> kvp in headers)
```
Joe118: So if this Client is reused many times it will keep adding headers to the globalHeaders and send these in subsequent requests? Don't you need to reset the headers to a known state when a REST request is finished?

examples/csharp/src/services/account.cs
```diff
@@ -0,0 +1,91 @@
...
+    {
+        path = "/account/prefs";
+        paramsIn = new Dictionary<string, string>();
+
+        return _client.Call("get", path, new Dictionary<string, string>(), paramsIn);
+    }
+
+    public UpdatePrefs(string prefs)
+    {
+        path = "/account/prefs";
```
Joe118: Elsewhere you use $"some string", here you use "another string" w.o the $... which one is right for this use case?

examples/csharp/src/client.cs
```diff
@@ -0,0 +1,136 @@
...
+        {
+            _globalHeaders["x-appwrite-mode"] = value.ToLower();
+        }
+        else
+        {
+            _globalHeaders.Add("x-appwrite-mode", value.ToLower());
+        }
+    }
+
+    public Call(string method, string path="", Dictionary<string, string> headers = null, Dictionary<string, string> paramsIn = null)
```
Joe118: As far as I can tell, all invocations of Call() pass an empty Dictionary for headers. Maybe get rid of this parameter entirely?

examples/csharp/src/services/database.cs
```diff
@@ -0,0 +1,127 @@
...
+            "limit" = limit,
+            "offset" = offset,
+            "orderType" = orderType
+        };
+        path = $"/database";
+
+        return _client.Call("get", path, new Dictionary<string, string>(), paramsIn);
+    }
+
+    public CreateCollections(string name = "", string[] read = "", string[] write = "", string[] rules = "")
```
Joe118: Should this be called CreateCollection() ? I think this method only creates a single collection, correct? Same for UpdateCollections, GetCollections, DeleteCollections. The operations for documents have correct names i.e. DeleteDocuments and ListDocuments :)

PR #15, opened 2019-10-11
Dart optimizations

olexale:
With this pull request, I propose to make the Dart code correspond more to coding standard from `Effective Dart`, and make API look more 'dart'-ish.

I have an issue running `php example.php`, and I'm not really familiar with neither PHP nor Twig templates, so I modified the dart code, and then backported it to templates. I hope everything went fine, but if there is an issue - please feel free to get back to me.

Thanks!

### DIFF
--- a/templates/dart/lib/client.dart.twig
+++ b/templates/dart/lib/client.dart.twig
@@ -4,9 +4,10 @@ import 'package:cookie_jar/cookie_jar.dart';
 
 class Client {
     Client({
-        this.selfSigned = false,
+        bool selfSigned = false,
         String endPoint = '{{spec.endpoint}}',
     }) : http = Dio(),
+         _selfSigned = selfSigned,
          _endPoint = endPoint,
          headers = {} {
         headers.addAll({
@@ -41,7 +42,7 @@ class Client {
 
 {% endfor %}
     Client setSelfSigned({bool status = true}) {
-        selfSigned = status;
+        _selfSigned = status;
 
         return this;
     }
--- a/templates/dart/lib/client.dart.twig
+++ b/templates/dart/lib/client.dart.twig
@@ -21,9 +21,11 @@ class Client {
 
     String _endPoint;
     String get endPoint => _endPoint;
+    
+    bool _selfSigned;
+    bool get selfSigned => _selfSigned;
 
     final Map<String, String> headers;
-    bool selfSigned;
     final Dio http;
 
 {% for header in spec.global.headers %}
--- a/templates/dart/lib/client.dart.twig
+++ b/templates/dart/lib/client.dart.twig
@@ -3,24 +3,28 @@ import 'package:dio_cookie_manager/dio_cookie_manager.dart';
 import 'package:cookie_jar/cookie_jar.dart';
 
 class Client {
-    String endPoint;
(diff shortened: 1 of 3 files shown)


Almoullim:
What is the issue you are facing?

olexale:
When I run `php example.php` I get:
```
Error: Passed variable is not an array or object on /Users/ole/Development/Pets/sdk-generator/src/Spec/Spec.php:44
```

The result is the same without my changes, so I think there is something wrong with my dev environment. I have macOS 10.15, PHP 7.3.8.

Thanks!

eldadfux:
Hey @olexale seems like the problem is in this line:
https://github.com/appwrite/sdk-generator/blob/e9f394f5505c55874753a68333b5c18bd09afd6a/example.php#L32

Just comment line 32, and uncomment line 31.

I'll make sure to add the spec locally in the project for demos, that way there will be no unnecessary dependencies.

olexale:
Thanks, it works with that change!
I noticed some issues with the proposed changes. Fixed them, so the generated code should work now.

eldadfux:
@olexale That's Great! Going to review and merge them soon.

I am working on some improvements to make the project more friendly for new contributors. It would be really great to get some feedback about your experience working on the repo.

I am also working on adding some kind of a testing strategy that will be cross-platform, to help us make sure the output templates are stable for production usage.

olexale:
Here are some of my thoughts about SDK repositories:
1) At first I was confused about the `sdk-for-dart` repository. I started with the code modification, and only then read a README where I found that the whole repo is the result of code generation, and the original template is in another place. My fault, but this might be highlighted even more.

2) `php example.php` command is not listed in README. Maybe it's obvious for PHP devs, but it took me some time to understand that it is the "entry point" of the generator and not the part of some website.

3) I have some concerns about SDK generation. As a consumer, I would love to get the most optimized SDK, which looks native to the ecosystem. It is hard to achieve this with code generation, because sometimes `path` might be a constant, sometimes `params` initialization is not needed, etc. Most importantly, it is not possible to implement model deserialization, so users will be forced to deal with `dynamic` type. It is not a best practice for statically typed languages.

Don't get me wrong, I think this is a really good project, and I want to see it growing! I'd be glad to offer my help with SDK redevelopment without code generation. It is not that big. :-)

eldadfux:
@olexale thank you so much! great feedback!

1. Totally agree, did tried to make it more clear after I saw a couple of pull requests by developers. I will think of more ways to make it more obvious.

2. The `php example.php` isn't something obvious for PHP devs as well. I omitted it from the readme because it's only relevant for projects maintainers and not users of the library. I will try to make it more clear in the contribution guide. I have also replace it with a docker command to make it easier for maintainers of other languages & ecosystems.

3. Your concerns about SDK generation are totally in place. This is not a magic solution and it is not suitable to every project at every stage of it. I do think it can be a great starting point, and the fact that our templates are extendable is very powerful for project in need for more customisation.

The static type support is actually something we do have in the generator and is currently missing from the dart templates. Would be great if you or @Almoullim will have some time to complete it.

You can see a good example for this in our new go templates:
https://github.com/appwrite/sdk-generator/blob/master/examples/go/avatars.go

And thanks again, great job & great feedback!!

PR #21, opened 2019-10-27
Enhance Travis CI setting

peter279k:
# Changed log
- Add `php-7.1` and `php-7.2` tests on Travis CI build.
- The `PHPUnit` will read `phpunit.xml ` configuration by default.
It's fine to remove this command option.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,15 @@
+dist: xenial
+
 language: php
 
 php:
-- 7.3
+  - 7.3
 
 notifications:
   email:
-  - team@appwrite.io
+    - team@appwrite.io
 
 before_script: composer install
 
 script:
-- vendor/bin/phpunit --configuration phpunit.xml
+  - vendor/bin/phpunit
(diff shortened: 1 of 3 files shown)


peter279k:
It looks like the Travis CI build has unexpected error when executing assertion work.

That's why the Travis CI build is failed.

.travis.yml
```diff
@@ -1,13 +1,17 @@
+dist: xenial
+
 language: php
 
 php:
-- 7.3
+  - 7.1
+  - 7.2
```
eldadfux: This was actually intentional. The tests in this project are run inside docker containers. This is done because this project is cross-platform and we don't have the capacity to implement multiple testing frameworks in multiple languages.

The PHP part in this CI is only to trigger the tests that are executed inside the containers. Using only the latests PHP version is enough and by running the tests only on one env we save a lot of travis CPU time.

peter279k: @eldadfux, thanks for your reply and explanation.

Revert this change and using only latest stable `PHP-7.3` version :).

eldadfux:
Thanks again @peter279k, You're AWESOME! ❤️

Issue #4, opened 2019-09-20
Add Swift support


eldadfux:
Add Swift templates to allow easy generation of Swift SDKs for iOS, macOS and other platform that rely on the language.

A full guide on how to add new language support to the Appwrite SDK generator is located here:
https://github.com/appwrite/sdk-generator/blob/master/CONTRIBUTING.md

lbm:
If this isn't being worked on, I'd be happy to give it a try @eldadfux.

eldadfux:
@lbm This would be great! This is one of our most requested features and we have been looking for a Swift contributor for quite some time now. I would be happy to assist with any question here or on our Discord server (https://discord.gg/GSeTUeA). Feel free to reach out!

lbm:
No worries, I’ll fork it and start looking into it as soon as possible. I’ll join the discord then too 🙂

armino-dev:
Hi guys, I hope you don't mind. I had some spare time and made the first version of the Swift SDK generator

PR #28, opened 2020-01-29
Added the swift sdk generator

armino-dev:
In ref to #4 here it's the first version of the Swift SDK.


eldadfux:
Sorry for the delay! Great work as usual! gonna try it out!

eldadfux:
@armino-dev any idea what are the most popular package managers for swift packages and SDKs?

armino-dev:
I think The Swift Package Manager is the way as it works on both Linux and Apple platforms. When I'll have a little more time I'll make some examples and a tutorial on how to integrate with Appwrite.

PR #30, opened 2020-04-03
Dart SDK generator code refactor

bartektartanus:
Simplified some code, added assertions and test. Cleanup, reformatting and OrderType enums instead of Strings.


eldadfux:
Our approach today is to avoid adding native unit-tests the traditional way. This is done because the repo has become extremely complicated when we had to add so many dependencies from so many different eco-systems and languages and it was really hard to test them all using traditional CI tools. It was also really hard to enforce all SDKs are being tested equally.

Our new approach is to to make end-to-end like tests where we create a test file for each language and in that file we send requests to a mock server, the server checks the requests structure and returns the result. If all output is as expected the test passes. All the tests run in a docker isolated env using the a single test suit (phpunit validates the output using docker exec), this way we can also run tests to each SDK vs multiple versions of each language.

Example Node.js test:
https://github.com/appwrite/sdk-generator/blob/master/tests/languages/node/test.js

Example Ruby test:
https://github.com/appwrite/sdk-generator/blob/master/tests/languages/ruby/tests.rb

Main test file with Docker settings:
https://github.com/appwrite/sdk-generator/blob/master/tests/SDKTest.php

In this approach we test the SDK generator being able to produce code that work rather than testing the code itself. At the end of the day, the result is pretty similar -> a working SDK.

bartektartanus:
These tests only verify if Client can properly perform GET/POST/etc request - not if all the required parameters are sent when you want for example get all files and you need to pass offset and limit.

examples/dart/lib/services/account.dart
```diff
@@ -141,74 +140,74 @@ class Account extends Service {
...
-       return await this.client.call(HttpMethod.delete, path: path, params: params);
+        return this.client.call(HttpMethod.delete, path: path, params: params);
     }
      /// Allow the user to login to his account using the OAuth2 provider of his
      /// choice. Each OAuth2 provider should be enabled from the Appwrite console
      /// first. Use the success and failure arguments to provide a redirect URL's
      /// back to your app when login is completed.
-    Future<Response> createOAuth2Session({@required String provider, @required String success, @required String failure}) async {
-       String path = '/account/sessions/oauth2/{provider}'.replaceAll(RegExp('{provider}'), provider);
+    Future<Response> createOAuth2Session({@required String provider, @required String success, @required String failure}) {
```
eldadfux, line 191: On the web client (JS SDK) we use this method to just return the GET url required for login, like this:
https://github.com/appwrite/sdk-for-js/blob/master/src/sdk.js#L773

This is done because of the way OAuth protocol works.

Does this seems like a reasonable solution for flutter applications?

examples/dart/pubspec.yaml
```diff
@@ -8,4 +8,4 @@ environment:
 dependencies:
   dio: ^3.0.0
   cookie_jar: ^1.0.0
```
eldadfux, line 10: are we making sure cookies persist?

src/SDK/Language/Dart.php
```diff
@@ -99,7 +99,7 @@ public function getTypeName($type)
                 return 'String';
             break;
             case self::TYPE_FILE:
-                return 'File';
+                return '';
```
eldadfux, line 102: We need to figure out a way to handle file uploads. This is usually a harder task in most language where we had to implement it.

templates/dart/lib/client.dart.twig
```diff
@@ -8,49 +8,42 @@ class Client {
...
+
+    Client({this.endPoint: 'https://appwrite.io/v1', this.selfSigned: false, Dio http}) : this.http = http ?? Dio() {
         this.headers = {
             'content-type': 'application/json',
             'x-sdk-version': '{{spec.title | caseDash}}:{{ language.name | caseLower }}:{{ sdk.version }}',
         };
-        this.selfSigned = false;
 
-        this.http = Dio();
+        assert(endPoint.startsWith(RegExp("http://|https://")), "endPoint $endPoint must start with 'http'");
```
eldadfux, line 19: Is this method related to a unit-tests library?

eldadfux:
Another thing I forgot about. Is there a way to enable self-signed certificates requests for people using Appwrite from localhost?

This is what we do in the Node SDK:
https://github.com/appwrite/sdk-for-node/blob/master/lib/client.js#L103

bartektartanus:
> Another thing I forgot about. Is there a way to enable self-signed certificates requests for people using Appwrite from localhost?
> 
> This is what we do in the Node SDK:
> https://github.com/appwrite/sdk-for-node/blob/master/lib/client.js#L103

Do you want to accept all certificates or only or just certificates from localhost?

eldadfux:
> > Another thing I forgot about. Is there a way to enable self-signed certificates requests for people using Appwrite from localhost?
> > This is what we do in the Node SDK:
> > https://github.com/appwrite/sdk-for-node/blob/master/lib/client.js#L103
> 
> Do you want to accept all certificates or only or just certificates from localhost?

Appwrite developers can use the Appwrite server from any hostname or even an IP address locally. I think it will be best to allow all certificates. This should only be allowed if the user has used the setSelfSigned method as we don't want to weaken security by default.

templates/dart/lib/client.dart.twig
bartektartanus: No, it's not. But I've decided to dive deeper into docs and find out that assertions are not used in production code. So if developer makes a mistake, he will see an error in logs during development. But it won't be evaluated on end users phones.

examples/dart/pubspec.yaml
bartektartanus: Hard to say, never used this lib and it was there earlier. But there is only one class `CookieManager` which is an interceptor and it is added to http client.

src/SDK/Language/Dart.php
bartektartanus: I removed this, because `File` class is for representing file on file system. It's just a reference, not binary content of the file - which I think is needed in this case. Currently there is no type, so it's "dynamic" and user can pass whatever he wants to. So until we'll figure it out - this is the best solution.

examples/dart/pubspec.yaml
eldadfux: I think I can add a test for this in the mock server, it will also allow us to enforce that all SDKs know how to handle this.

src/SDK/Language/Dart.php
eldadfux: 👌🏻

templates/dart/lib/client.dart.twig
eldadfux: 👌🏻

examples/dart/lib/services/account.dart
eldadfux: This seems to be a little more complex than I thought. Gonna merge the PR and test this locally.

bartektartanus: There are some packages for authenticating via oauth2:
https://pub.dev/packages/oauth2

PR #40, opened 2020-05-23
feat: add Deno language templates

Almoullim:
Missing:
 - ~~Missing testing (Deno doesn't have an official docker image yet, might have to use one from the community)~~ 
- Missing upload test case (If implemented it will always fail because `Blob` is making the file size bigger, which could be a bug on Deno's side)

Info:
- For publishing, the repo can be added here https://deno.land/x and optionally here https://module.land/


templates/deno/README.md.twig
```diff
@@ -0,0 +1,30 @@
...
+
+{{ sdk.description }}
+
+
+
+![{{ spec.title }}]({{ sdk.logo }})
+
+## Installation
+
+To install via [NPM](https://www.npmjs.com/):
```
eldadfux, line 18: Just read here that Deno isn't using any package manager:
https://github.com/denoland/deno/issues/47

From what I understand they are using import of remote git URLs, any idea how this should look like?

Almoullim: Yea, I copied it from the TS templates and forgot to update it. later on add the repo through here https://deno.land/x

eldadfux:
🥳Merged!

We'll keep an eye on next Deno releases to make sure file uploads also work properly....

PR #55, opened 2020-09-05
Adding support for oAuth using universal_html

lohanidamodar:
@eldadfux this PR adds support for oAuth on Flutter Web using universal_html, as suggested, I have just redirected the web to the oAuth url. In documentation we need to add that, if we are building for flutter web, we need to pass success and failure URL, that is how it gets redirected.

The success case works fine, we get logged in and redirected. I am  not sure how it works for failure case, we should get redirected, but not know how we can handle the error case, as i was not able to reproduce a error case.


eldadfux:
Thanks @lohanidamodar! I'll try double check the failure URL.

Issue #33, opened 2020-04-09
Replace deprecated Node.js request library


eldadfux:
The request library has been marked as deprecated. We need to replace it with a valid alternative.

More information and alternative libraries are listed here:
https://github.com/request/request/issues/3143

meysam81:
Hey there,

I have zero experience using PHP. I was looking for a way to test my changes against the current code and I couldn't find one, to see if it breaks anything or not. Can you guide me please?

eldadfux:
You will need composer installed (php equivalent to npm).

When composer is installed, run:

```bash
composer install
```

and then run:

```
vendor/bin/phpunit
```

to execute all the tests

eldadfux:
This issue is open for grab.

PineappleIOnic:
Can I try changing the library?

eldadfux:
@PineappleIOnic yes please 🙏🏻

PineappleIOnic:
#66 has been merged, I'm pretty sure this issue can be closed 😄

eldadfux:
@PineappleIOnic I'm leaving it open as a reminder to publish the new SDK version :)

PineappleIOnic:
Ah, alright 👍

eldadfux:
Didn't forget about this we'll probably release in 1-2 weeks max! Adding myself a reminder!

eldadfux:
Forgot to update here, but this has been released on the latest version together with the release of Appwrite 0.7. Thank for the help @PineappleIOnic 👏🏻

PR #66, opened 2020-10-06
Switch from requests to axios

PineappleIOnic:
+ Change request library from Requests to Axios

This is a PR in accordance to #33

### DIFF
--- a/templates/node/lib/client.js.twig
+++ b/templates/node/lib/client.js.twig
@@ -1,5 +1,6 @@
 const URL = require('url').URL;
-const request = require('request-promise-native');
+const axios = require('axios');
+const FormData = require('form-data');
 
 class Client {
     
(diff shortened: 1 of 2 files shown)


templates/node/lib/client.js.twig
```diff
@@ -70,23 +71,39 @@ class Client {
...
 
         let contentType = headers['content-type'].toLowerCase();
+
+        let formData = null;
+
+        // Compute FormData for axios and appwrite.
+        if (contentType.startsWith('multipart/form-data')) {
+            const form = new FormData();
+            
+            let test = this.flatten(params);
```
eldadfux, line 81: Probably better use a more meaningful name here

templates/node/lib/client.js.twig
```diff
@@ -70,23 +71,39 @@ class Client {
...
+            
+            for (const key in test) {
+                form.append(key, test[key])
```
eldadfux, line 84: same here obviously.

templates/node/lib/client.js.twig
PineappleIOnic: Oh, yeah. I forgot about that, that's when I was trying to figure it out by quickly prototyping new ideas

PineappleIOnic: I'll fix it now

PR #73, opened 2021-01-01
add tests to c# sdk

TorstenDittmann:
- adds build step for C#
- introduces tests for C# run with PowerShell
- fixes array interpretation for query strings
- adds self signing to SDK


eldadfux:
Looks awesome!

Issue #76, opened 2021-01-04
Incorrect parameter type Functions API


lohanidamodar:
In the functions API, in Create and Update method, the param `vars` is treated as `Array` instead of object, the default value assigned is `array` when generating the SDK.

PR #78, opened 2021-01-04
[Draft] Dart sdk

lohanidamodar:
Hey,
- I have created Dart SDK templates,
- cleaned up Dart language file a bit with some fixes
- Removed Flutter specific and unused dependencies

Running into incorrect parameter type issue in functions api as mentioned here #76

### DIFF
--- a/templates/dart/lib/client.dart.twig
+++ b/templates/dart/lib/client.dart.twig
@@ -59,8 +59,6 @@ class Client {
 
     Future init() async {
         if(!initialized) {
-          cookieJar = new PersistCookieJar(dir:'./cookies');
-          this.http.interceptors.add(CookieManager(cookieJar));
           this.http.options.baseUrl = this.endPoint;
           this.http.options.validateStatus = (status) => status < 400;
         }
--- a/templates/dart/lib/client.dart.twig
+++ b/templates/dart/lib/client.dart.twig
@@ -1,11 +1,8 @@
 import 'dart:io';
-
 import 'package:dio/dio.dart';
 import 'package:dio/adapter.dart';
 import 'package:dio_cookie_manager/dio_cookie_manager.dart';
 import 'package:cookie_jar/cookie_jar.dart';
-import 'package:path_provider/path_provider.dart';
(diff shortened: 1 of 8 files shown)


TorstenDittmann:
@lohanidamodar Uhh nice 👀 

Will this also be compatible with Flutter Web? I remember it being a problem in the past.

lohanidamodar:
@TorstenDittmann we have separate flutter SDK which is already compatible for flutter web as well

TorstenDittmann:
> @TorstenDittmann we have separate flutter SDK which is already compatible for flutter web as well

Ah, so Dart SDK != Flutter SDK? 😄 
That confused me for a moment 👍

templates/dart/lib/client.dart.twig
```diff
@@ -76,17 +59,10 @@ class Client {
 
     Future init() async {
         if(!initialized) {
-          final Directory cookieDir = await _getCookiePath();
-
-          cookieJar = new PersistCookieJar(dir:cookieDir.path);
-
+          cookieJar = new PersistCookieJar(dir:'./cookies');
```
eldadfux, line 62: I think we don't need the cookie jar for the server integration. No need for any cookie based auth when integrating with the server API, we'll only use API Key & JWT for authentication.

eldadfux:
> > @TorstenDittmann we have separate flutter SDK which is already compatible for flutter web as well
> 
> Ah, so Dart SDK != Flutter SDK? 😄
> That confused me for a moment 👍

Yeh, our naming conventions for SDKs are sdk-for-[platform]. On the server the platform name will be the language name (sdk-for-php, sdk-for-python). On the client it would be the platform name (sdk-for-web, sdk-for-flutter, sdk-for-ios, sdk-for-android, sdk-for-unity...). That's also the reason for changing sdk-for-js to sdk-for-web

templates/dart/lib/client.dart.twig
lohanidamodar: @eldadfux Ok

PR #82, opened 2021-01-05
Fix for object default value

eldadfux:
Fix for #76


lohanidamodar:
@eldadfux Great, looks like this resolves #76 . I was able to generate correct values using this branch.

eldadfux:
Awesome. Merging changes.

PR #85, opened 2021-01-07
fix(dotnet): update deps, templates and examples

TorstenDittmann:
- Added default values to parameters in order to utilize named/optional arguments
- Adjusted dev dependencies to support more platforms
- Added placeholders to the examples (the expected data type)
    - If I would add complete examples, things would get messy.


eldadfux:
🤘

PR #86, opened 2021-01-07
fix(dotnet): nullish optional datatype

TorstenDittmann:
Some default values data types needed a `?` appended to indicate or possible nullish value.


src/SDK/Language/DotNet.php
```diff
@@ -181,26 +182,20 @@ public function getParamDefault(array $param)
             switch ($type) {
                 case self::TYPE_NUMBER:
                 case self::TYPE_INTEGER:
+                case self::TYPE_ARRAY:
```
eldadfux, line 185: why not use something similar to [] for array?

TorstenDittmann: Because with the data type `List<object>` and a value of `null`, it becomes an empty array. That was confusing to me as well 👍

TorstenDittmann:
Also changed methods that don't do any HTTP requests to non async ones for friendlier use.

src/SDK/Language/DotNet.php
```diff
@@ -152,7 +152,8 @@ public function getTypeName($type)
                 return 'bool';
             break;
             case self::TYPE_ARRAY:
-            	return 'List<object>';
+                return 'List<object>';
+            break;
 			case self::TYPE_OBJECT:
```
eldadfux, line 157: Indentation got a bit off here

eldadfux (changes requested):

templates/dotnet/src/Appwrite/Services/ServiceTemplate.cs.twig
```diff
@@ -1,6 +1,6 @@
 {% macro parameter(parameter) %}
-{% if parameter.name == 'orderType' %}{{ 'OrderType orderType' }}{% else %}
-{{ parameter.type | typeName }} {{ parameter.name | caseCamel | escapeKeyword }}{{ parameter | escapeDollarSign | paramDefault }}{% endif %}
+{% if parameter.name == 'orderType' %}{{ 'OrderType orderType = OrderType.ASC' }}{% else %}
+{{ parameter.type | typeName }}{% if not parameter.required and (parameter.type == 'boolean' or parameter.type == 'integer') %}?{% endif %} {{ parameter.name | caseCamel | escapeKeyword }}{{ parameter | escapeDollarSign | paramDefault }}{% endif %}
```
eldadfux, line 3: Why do we have a special condition for booleans and ints?

TorstenDittmann: They only allow `null` values with a `?`.

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-value-types

src/SDK/Language/DotNet.php
TorstenDittmann: good catch 👍

eldadfux (approved):

eldadfux:
👏🏻

PR #88, opened 2021-01-08
fix(dotnet): prevent headers from being an array

TorstenDittmann:
In some occasions it is possible to transform headers like project or api key to a comma seperated list. This fix prevents it 👍 

> If the specified header is already present, values are added to the comma-separated list of values associated with the header.


eldadfux:
👍🏻

PR #90, opened 2021-01-10
Feature Ruby SDK upload support

lohanidamodar:
Hey,

This PR adds Upload support to Ruby SDK.

## Few things to watch for
1. The build phase required ruby image instead of docker because of gemspec git dependency
2. The file content-type 'multipart/form-data' always assumes that a parameter named 'file' is there and is a valid path to a file
3. The binary content of the file is loaded in memory at once might not be suitable for large files at the moment

### DIFF
--- a/templates/ruby/lib/container/client.rb.twig
+++ b/templates/ruby/lib/container/client.rb.twig
@@ -100,7 +100,7 @@ module {{spec.title | caseUcfirst}}
             when nil   then ''
             else
                 post_body = []
-                if(value.instance_of? AppwriteFile)
+                if(value.instance_of? {{spec.title | caseUcfirst}}File)
                     post_body << "Content-Disposition: form-data; name=\"#{key}\"; filename=\"#{value.name}\"\r\n"
                     post_body << "Content-Type: #{value.mimeType}\r\n\r\n"
                     post_body << value.content
--- a/templates/ruby/lib/container/client.rb.twig
+++ b/templates/ruby/lib/container/client.rb.twig
@@ -2,7 +2,6 @@ require 'net/http'
 require 'uri'
 require 'json'
 require 'cgi'
-require 'mime/types'
 
 module {{spec.title | caseUcfirst}}
     class Client
--- a/templates/ruby/lib/container/client.rb.twig
+++ b/templates/ruby/lib/container/client.rb.twig
@@ -100,10 +100,10 @@ module {{spec.title | caseUcfirst}}
             when nil   then ''
             else
                 post_body = []
-                if(key.to_s == "file")
-                    post_body << "Content-Disposition: form-data; name=\"#{key}\"; filename=\"#{File.basename(value)}\"\r\n"
-                    post_body << "Content-Type: #{MIME::Types.type_for(value)}\r\n\r\n"
-                    post_body << File.read(value)
+                if(value.instance_of? AppwriteFile)
+                    post_body << "Content-Disposition: form-data; name=\"#{key}\"; filename=\"#{value.name}\"\r\n"
+                    post_body << "Content-Type: #{value.mimeType}\r\n\r\n"
+                    post_body << value.content
                     post_body << "\r\n--#{@BOUNDARY}--\r\n"
                 else          
                     post_body << "Content-Disposition: form-data; name=\"#{key}\"\r\n\r\n"
--- a/templates/ruby/lib/container/client.rb.twig
+++ b/templates/ruby/lib/container/client.rb.twig
@@ -99,7 +99,7 @@ module {{spec.title | caseUcfirst}}
             when nil   then ''
             else
                 post_body = []
-                if(value.instance_of? {{spec.title | caseUcfirst}}File)
+                if(value.instance_of? File)
                     post_body << "Content-Disposition: form-data; name=\"#{key}\"; filename=\"#{value.name}\"\r\n"
                     post_body << "Content-Type: #{value.mimeType}\r\n\r\n"
                     post_body << value.content
--- a/templates/ruby/lib/container/client.rb.twig
+++ b/templates/ruby/lib/container/client.rb.twig
@@ -2,6 +2,7 @@ require 'net/http'
 require 'uri'
 require 'json'
 require 'cgi'
+require 'mime/types'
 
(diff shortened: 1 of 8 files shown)


tests/SDKTest.php
```diff
@@ -124,15 +124,16 @@ class SDKTest extends TestCase
         'ruby' => [
             'class' => 'Appwrite\SDK\Language\Ruby',
             'build' => [
+                'docker run --rm -v $(pwd):/app -w /app/tests/sdks/ruby --env GEM_HOME=/app/vendor ruby:latest bundle install'
             ],
             'envs' => [
-                'ruby-2.7' => 'docker run --rm -v $(pwd):/app -w /app ruby:2.7-alpine ruby tests/languages/ruby/tests.rb',
+                'ruby-2.7' => 'docker run --rm -v $(pwd):/app -w /app --env GEM_HOME=vendor ruby:2.7-alpine ruby tests/languages/ruby/tests.rb',
```
eldadfux, line 130: It seems like 2.7 tests are working and the rest are not. Maybe related to the fact you only added `--env GEM_HOME=vendor` to the 2.7 test.

lohanidamodar: Pushed the fix, should work now.

templates/ruby/lib/container/client.rb.twig
```diff
@@ -88,6 +92,27 @@ module {{spec.title | caseUcfirst}}
...
         end
+        
+        def encodeFormData(value, key=nil)
+            case value
+            when Hash  then value.map { |k,v| encodeFormData(v,k) }.join
+            when Array then value.map { |v| encodeFormData(v, "#{key}[]") }.join
+            when nil   then ''
+            else
+                post_body = []
+                if(key.to_s == "file")
```
eldadfux, line 103: Yep the hardcoded key might be a bit problematic, cloud functions will also need file upload under a different key name. In other languages we accept a file instance and that way we can detect its type. Maybe the createFile method should expect a File.read(value) instance and not a string.

Another thing we need to do, is to verify the code examples for file uploads is correct.

eldadfux (changes requested):

templates/ruby/lib/container/client.rb.twig
lohanidamodar: My bad, did i forget to mention the hard coded file key. I will investigate further.

lohanidamodar:
@eldadfux I have introduced new AppwriteFile class to handle File uploads instead of hard coding the file key. Please review.

templates/ruby/lib/container/file.rb.twig
```diff
@@ -0,0 +1,9 @@
+module {{spec.title | caseUcfirst}}
+    class AppwriteFile
```
eldadfux, line 4: Avoid the word `Appwrite` inside the templates. You can use a Twig param to produce in the generated code.

templates/ruby/lib/container/client.rb.twig
```diff
@@ -100,10 +100,10 @@ module {{spec.title | caseUcfirst}}
             when nil   then ''
             else
                 post_body = []
-                if(key.to_s == "file")
-                    post_body << "Content-Disposition: form-data; name=\"#{key}\"; filename=\"#{File.basename(value)}\"\r\n"
-                    post_body << "Content-Type: #{MIME::Types.type_for(value)}\r\n\r\n"
-                    post_body << File.read(value)
+                if(value.instance_of? AppwriteFile)
```
eldadfux, line 103: Avoid the word `Appwrite` inside the templates. You can use a Twig param to produce in the generated code.

eldadfux: Because we already use {{ spec.title | caseDash }} as a namespace we might be able to use just `File` for the class name.

templates/ruby/lib/container/file.rb.twig
eldadfux: Because we already use {{ spec.title | caseDash }} as a namespace we might be able to use just `File` for the class name.

eldadfux (changes requested):

templates/ruby/lib/container/file.rb.twig
lohanidamodar: @eldadfux File will clash with the File keyword that already exists in Ruby lang for file operations

lohanidamodar: @eldadfux If i named the class File then while accessing can I use Module(Appwrite)::File.new() instead of File.new() ?

eldadfux: I'm not sure, please test it. Worst case we can use the Appwrite keyword but auto-generate it using the twig param and not have it hardcoded in the template.

lohanidamodar: Okay!

tests/languages/ruby/tests.rb
```diff
@@ -45,3 +45,8 @@
 
 response = general.redirect();
 puts response['result']
+
+
+file = Appwrite::AppwriteFile.new('./tests/resources/file.png')
```
eldadfux, line 50: Are we sure we can't just use Appwrite::File.new? I also notice we still use `File.new()` for the code examples.

lohanidamodar: I am sure, as inside the Appwrite module itself we will be using File and that will clash with the File keyword.

lohanidamodar: I was unable to make it work locally I tried various options.

lohanidamodar: @eldadfux Found a way to make it as you requested, I have updated to make it (`Appwrite::File.new(...)`) to work.

eldadfux: Nice! Restarted the build to make sure tests pass.

eldadfux:
We have 2 places here that are still missing the new File class namespace. We use this for generating examples.
https://github.com/lohanidamodar/sdk-generator/blob/feat-ruby-sdk-upload-support/src/SDK/Language/Ruby.php#L276 (and line 255)

eldadfux (approved):

eldadfux:
🥳

PR #99, opened 2021-02-08
feat: add support for file previews

christyjacob4:
For endpoints with that send back a binary file, provide a URL to view it in the browser.


templates/cli/app/services/service.php.twig
```diff
@@ -94,6 +94,15 @@ $cli
         $params['{{ parameter.name }}'] = ${{ parameter.name | caseCamel }};
 {% endfor %}
 
+{% if method.type == 'location' %}
+
+        $params['project'] = $client->getPreference('X-Appwrite-Project');
+        $params['key'] = $client->getPreference('X-Appwrite-Key');
+        $path = $client->getPreference(Client::PREFERENCE_ENDPOINT).$path . "?" . http_build_query($params);
+        echo "Please follow this URL to view your file \n".$path;
```
eldadfux, line 102: I think we should skip this message in case someone need to use this command in a programatic way

christyjacob4: Done

PR #101, opened 2021-02-09
Fixed descriptions

eldadfux:


src/SDK/Language/PHP.php
```diff
@@ -10,8 +10,8 @@ class PHP extends Language {
      * @var array
      */
     protected $params = [
-        'composerVendor' => 'vendorName',
-        'composerPackage' => 'packageName',
+        'composerVendor' => 'vendor-name',
```
lohanidamodar, line 13: Not sure why this change was made?

lohanidamodar (approved):
LGTM.

src/SDK/Language/PHP.php
eldadfux: Composer seem to have issue with uppercase letters in these values.

PR #102, opened 2021-02-10
Fix flutter desktop with package info plus

lohanidamodar:
Updates and fixes to make work with latest Flutter and Dart and in desktop platforms.


eldadfux:
✅

PR #110, opened 2021-02-15
feat-flutter-dart-sdk-private-storage-files

lohanidamodar:


TorstenDittmann (approved):

eldadfux:
@lohanidamodar can we make the flutter code example for images show how to use the payload with the image widget as you showed me?

lohanidamodar:
> @lohanidamodar can we make the flutter code example for images show how to use the payload with the image widget as you showed me?

Sure. Where do you think will be the right place?

eldadfux:
> > @lohanidamodar can we make the flutter code example for images show how to use the payload with the image widget as you showed me?
> 
> Sure. Where do you think will be the right place?

As part of the generated examples we create to show on appwrite.io specifically for the image relate endpoints. Maybe we can use this attribute to detect the relevant endpoints: https://github.com/appwrite/appwrite/blob/master/app/controllers/api/storage.php#L237 I think it should be available on the Swagger format, but we might need to pass it to the twig params if we don't do that already.

lohanidamodar:
@eldadfux You want me to add this example to get file preview and get file view endpoint docs?  we can use the same check (.location) that we use in the client to update those endpoint.

eldadfux:
> @eldadfux You want me to add this example to get file preview and get file view endpoint docs? we can use the same check (.location) that we use in the client to update those endpoint.

👍🏻

lohanidamodar:
@eldadfux I have updated the docs, but the docs for these three endpoints look totally different as it will consists of not only main functions but also, widgets and flutter specific stuff

eldadfux:
> @eldadfux I have updated the docs, but the docs for these three endpoints look totally different as it will consists of not only main functions but also, widgets and flutter specific stuff

Please post an example of the new example output 🙏🏻

lohanidamodar:
Here is the output of getFileDownload() endpoint
```dart
import 'package:appwrite/appwrite.dart';

Client client = Client();
Storage storage = Storage(client);

client
  .setEndpoint('https://[HOSTNAME_OR_IP]/v1') // Your API Endpoint
  .setProject('5df5acd0d48c2') // Your project ID
  .setKey('919c2d18fb5d4...a2ae413da83346ad2') // Your secret API key
  ;

void main() { // Init SDK
  runApp(MyApp());
}

class MyApp extends StatelessWidget {
  @override
  Widget build(BuildContext context) {
    return Scaffold(
      appBer: AppBar(
        title: Text("Appwrite+Flutter Private Image"),
      ),
      body: Container(
        child: FutureBuilder(
          future: storage.getFileDownload(
            fileId: '[FILE_ID]',
          ), //works for both public file and private file, for private files you need to be logged in
          builder: (context, snapshot) {
            return snapshot.hasData && snapshot.data != null
                ? Image.memory(
                    snapshot.data.data,
                  )
                : CircularProgressIndicator();
          },
        ),
      )
    )
  }
}
```

eldadfux:
@lohanidamodar any other creative way to show this method used with a widget in a shorter example? Right now this seems to be x3 long than any other example we have.

lohanidamodar:
The shorter example looks like
```
import 'package:appwrite/appwrite.dart';

void main() { // Init SDK
  Client client = Client();
  Storage storage = Storage(client);

  client
    .setEndpoint('https://[HOSTNAME_OR_IP]/v1') // Your API Endpoint
    .setProject('5df5acd0d48c2') // Your project ID
    .setKey('919c2d18fb5d4...a2ae413da83346ad2') // Your secret API key
  ;
}

//displaying image
FutureBuilder(
  future: storage.getFileDownload(
    fileId: '[FILE_ID]',
  ), //works for both public file and private file, for private files you need to be logged in
  builder: (context, snapshot) {
    return snapshot.hasData && snapshot.data != null
      ? Image.memory(
          snapshot.data.data,
        )
      : CircularProgressIndicator();
  },
);

```

eldadfux:
👏🏻

PR #112, opened 2021-02-19
Fix cli issues

christyjacob4:
* Moved the install script to the main repo
* Added allowList to install script to prevent users from executing other commands in the appwrite/cli docker container
* Show {} for empty arrays in the table
* Replace echo calls with Console::log calls.
* fix indentation and spacing in the twig templates
* removed unused imports
* remove references to appwrite in the templates
* move Appwrite logo to params
* Use Wordwrap to pretty print appwrite help output to display longer lines elegantly
* Replace usages of prinf with Console::Log()

Fixes : https://github.com/appwrite/sdk-for-cli/issues/3


templates/cli/install.ps1.twig
```diff
@@ -0,0 +1,108 @@
+<?php
+$version = $this->getParam('version', '');
+?>
```
eldadfux, line 3: You can remove this, and just print the twig version param instead. The output of this file can't parse PHP code blocks.

templates/cli/install.sh.twig
```diff
@@ -0,0 +1,187 @@
+<?php
+$version = $this->getParam('version', '');
+?>
```
eldadfux, line 3: You can remove this, and just print the twig version param instead. The output of this file can't parse PHP code blocks.

templates/cli/install.sh.twig
```diff
@@ -0,0 +1,187 @@
...
+APPWRITE_EXECUTABLE_NAME=appwrite
+
+# Appwrite executable file path 
+APPWRITE_EXECUTABLE_FILEPATH="${APPWRITE_INSTALL_DIR}/${APPWRITE_EXECUTABLE_NAME}"
+
+# Appwrite CLI image name
+APPWRITE_CLI_IMAGE_NAME=appwrite/cli
+
+# Appwrite CLI image version 
+APPWRITE_CLI_IMAGE_VERSION=$version
```
eldadfux, line 33: Use the twig template param for SDK version here.

templates/cli/install.ps1.twig
```diff
@@ -0,0 +1,108 @@
...
+$APPWRITE_EXECUTABLE_NAME = "appwrite.ps1"
+
+# Appwrite executable file path 
+$APPWRITE_EXECUTABLE_FILEPATH = Join-Path -Path $APPWRITE_INSTALL_DIR -ChildPath $APPWRITE_EXECUTABLE_NAME
+
+# Appwrite CLI image name
+$APPWRITE_CLI_IMAGE_NAME = "appwrite/cli"
+
+# Appwrite CLI image version 
+$APPWRITE_CLI_IMAGE_VERSION = "$version"
```
eldadfux, line 32: Use the twig template param for SDK version here.

eldadfux (changes requested):
Please also try to avoid the name 'Appwrite' in the twig templates, use a variable or generic name.

christyjacob4:
Resolved all comments @eldadfux

src/SDK/Language/CLI.php
```diff
@@ -4,6 +4,39 @@
...
 class CLI extends PHP {
 
+    /**
+     * @var array
+     */
+    protected $params = [
+        'composerVendor' => 'vendor-name',
+        'composerPackage' => 'package-name',
+        'executableName' => 'executable',
+        'logo' => 'logo'
```
eldadfux, line 14: Am empty string would be a better default here.

templates/cli/app/services/client.php.twig
```diff
@@ -49,4 +72,22 @@ $cli
...
 
+
+$cli
+    ->task('help')
+    ->action(function() use ($parser) {
+        Console::log("\e[0;31;m {{ language.params.logo | raw }} \e[0m") ;
+        Console::log("\nUsage : {{ language.params.executableName }} client [COMMAND]\n");
+        Console::log("Commands :");
+        $commands = [
+            "setEndpoint" => "Set your server endpoint.",
```
eldadfux, line 83: This is the data you wanted me to add to the spec, right?

src/SDK/Language/CLI.php
christyjacob4: Okay I will do that 👍

templates/cli/app/services/client.php.twig
christyjacob4: Yeap. It would be great if we could add `client` and `init` as services or under some other key in the spec. 

so I can iterate over the methods in the twig templates like so
```
{% for method in service.methods %}
 I would want methods to contain [setEndpoint, setProject, setKey, setLocale] for client and each method should have the required parameter and their description just like how we have for the other services. 
{% endfor %}
```

eldadfux (approved):

PR #113, opened 2021-02-22
feat-custom-exception-model

lohanidamodar:


templates/flutter/lib/exception.dart.twig
```diff
@@ -0,0 +1,10 @@
+part of {{ language.params.packageName }};
+
+class {{spec.title | caseUcfirst}}Exception implements Exception {
+  final String message;
+  final int code;
+  final dynamic response;
```
JulianBissekkou, line 6: Can we maybe have a specific type here? 🤔 
`String` maybe? 
`Map<String, dynamic>` is fine as well, but as a mobile developer who is coming from strongly typed languages like kotlin, java or swift I like having types 👍

JulianBissekkou: If the response can be different in some cases it's fine to have `dynamic` but then I would add a little comment to it. In our case, we are casting to some type.

lohanidamodar: I believe the response can be Json or can be String as well. @eldadfux do you agree?

templates/flutter/lib/client.dart.twig
```diff
@@ -118,18 +108,27 @@ class Client {
...
+                params[key] = params[key].toString();
+                }});
+                
+                return await http.get(path, queryParameters: params, options: options);
+            } else {
+                return await http.request(path, data: params, options: options);
+            }
+        } on DioError catch(e) {
+          if(e.response == null) {
+            throw {{spec.title | caseUcfirst}}AppwriteException(e.message);
```
eldadfux, line 127: I think this line in the Exceptions PR is wrong: `throw {{spec.title | caseUcfirst}}AppwriteException(e.message);` as it resolve to `AppwriteAppwriteException`

lohanidamodar: Ah !

templates/flutter/lib/exception.dart.twig
eldadfux: Yeh It might be tricky because sometimes a proxy can throw the error and we can't be sure of the format.

eldadfux (approved):

PR #123, opened 2021-03-03
Fix added missing file to the list

lohanidamodar:
Hey, looks like here again I made small mistake.

### DIFF
--- a/src/SDK/Language/Flutter.php
+++ b/src/SDK/Language/Flutter.php
@@ -80,6 +80,12 @@ class Flutter extends Dart {
 				'minify'        => false,
 			],
             [
+				'scope'         => 'default',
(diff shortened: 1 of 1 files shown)


src/SDK/Language/Flutter.php
```diff
@@ -79,6 +79,12 @@ public function getFiles()
 				'template'      => '/flutter/lib/enums.dart.twig',
 				'minify'        => false,
 			],
+            [
+				'scope'         => 'default',
+				'destination'   => '/lib/exception.dart',
+				'template'      => '/flutter/lib/exception.dart.twig',
+				'minify'        => false,
+			],
```
eldadfux, line 87: Indentation is off

eldadfux (changes requested):

PR #128, opened 2021-03-04
fix-flutter-sdk-file-issues

lohanidamodar:


templates/dart/lib/client.dart.twig
```diff
@@ -96,6 +96,10 @@ class Client {
           if(e.response == null) {
             throw {{spec.title | caseUcfirst}}Exception(e.message);
           }
+          if(responseType == ResponseType.bytes) {
+            final res = json.decode(utf8.decode(e.response.data));
```
eldadfux, line 100: Why do we need to decode the image to UTF8 and not sure why json.decode is used

eldadfux: I think I got it, this is to return the JSON error properly?

eldadfux: Maybe we can have a nicer switch-case based on the *response* (not request header) content-type header?

Also note, your'e using `AppwriteException` instead of `{{spec.title | caseUcfirst}}Exception`

lohanidamodar: 1. AppwriteException hardcoded is Fixed
2. We don't need to check anything else for now, just when ResponseType is bytes, so I think its' fine? What do you suggest?

eldadfux: I think we should change it to avoid future surprises. Check the response content-type header is the correct way to figure out which parsing to apply for the payload.

lohanidamodar: got it

templates/flutter/lib/client.dart.twig
```diff
@@ -128,7 +128,7 @@ class Client {
             throw {{spec.title | caseUcfirst}}Exception(e.message);
           }
           if(responseType == ResponseType.bytes) {
-            fif(e.response.headers['content-type'].contains('application/json')) {
+            if(e.response.headers['content-type'].contains('application/json')) {
               final res = json.decode(utf8.decode(e.response.data));
```
eldadfux, line 132: Do we need to explicitly use `utf8.decode` on any JSON response?

lohanidamodar: Yes, not any JSON response, but this, when we set ResponseType as bytes for Dio, the return value is utf8 encoded bytes instead of JSON string, (which we require for binary files) so even JSON response is encoded as bytes so we need to decode here

eldadfux: OK

eldadfux (approved):

PR #129, opened 2021-03-04
Feat python exception

lohanidamodar:


templates/python/package/client.py.twig
```diff
@@ -63,26 +64,32 @@ class Client:
...
+
+            content_type = response.headers['Content-Type']
+
+            if content_type.startswith('application/json'):
+                return response.json()
+
+            return response._content
+        except Exception as e:
+            if response:
+                raise AppwriteException(response.json()['message'], response.status_code, response.json())
```
eldadfux, line 90: Don't use `Appwrite`

lohanidamodar:
@eldadfux Ready!

PR #130, opened 2021-03-04
feat-getting-started-section

lohanidamodar:


eldadfux:
Looks good! @lohanidamodar lets discuss how the Flutter getting started should look like, and then will do something similar for other SDKs

eldadfux (approved):

PR #131, opened 2021-03-04
Feat .NET Exceptions

TorstenDittmann:
- code clean up
- add travis config for auto deploy to NuGet
- adds AppwriteException


templates/dotnet/src/Appwrite/Models/Exception.cs.twig
```diff
@@ -0,0 +1,21 @@
+using System;
+
+namespace Appwrite
+{
+    public class {{spec.title | caseUcfirst}}Exception : Exception
+    {
+        public int? Code;
+        public string Response = null;
```
lohanidamodar, line 8: Is response String?

templates/dotnet/src/Appwrite/Client.cs.twig
```diff
@@ -172,10 +158,24 @@ namespace {{ spec.title | caseUcfirst }}
...
+                if (code >= 400) {
+                    var message = response["message"].ToString();
+                    throw new AppwriteException(message, code, response.ToString());
+                }
+
+                return httpResponseMessage;
+            }
+            catch (System.Exception e)
+            {
+                throw new AppwriteException(e.Message, e);
```
lohanidamodar, line 176: passing second param as exception itself, will it work?

templates/dotnet/src/Appwrite/Client.cs.twig
```diff
@@ -172,10 +158,24 @@ namespace {{ spec.title | caseUcfirst }}
...
-            HttpResponseMessage httpResponseMessage = await http.SendAsync(request);
+            try
+            {
+                var httpResponseMessage = await http.SendAsync(request);
+                var code = (int) httpResponseMessage.StatusCode;
+                var response = JObject.Parse(await httpResponseMessage.Content.ReadAsStringAsync());
+                
+                if (code >= 400) {
+                    var message = response["message"].ToString();
+                    throw new AppwriteException(message, code, response.ToString());
```
eldadfux, line 169: Avoid mentioning the name `Appwrite` use the twig param, you can use references from other @lohanidamodar PRs.

eldadfux (changes requested):

templates/dotnet/src/Appwrite/Models/Exception.cs.twig
TorstenDittmann: It must be the JSON string in .NET - since we don't have Response Models yet and C# doesn't like unknown data models 👍🏻

templates/dotnet/src/Appwrite/Client.cs.twig
TorstenDittmann: Yes, that way I can pass the exception as an inner exception 🙂 

See the 2 constructors here:
https://github.com/appwrite/sdk-generator/blob/feat-dotnet-exceptions/templates/dotnet/src/Appwrite/Models/Exception.cs.twig#L15

templates/dotnet/src/Appwrite/Client.cs.twig
TorstenDittmann: hehe, missed these one 😄

lohanidamodar:
Looks good!

PR #132, opened 2021-03-04
add realtime prototype

TorstenDittmann:
Before refactoring the whole init process, I just added a `setEndpointRealtime` method for this.

Adds a new `realtime` object which contains the following:
- a socket property that holds the WebSocket connection
- a channels property that is an object, with each key representing the channel name holding an array of callbacks looking like this:
```
channels: {
  documents: [
    (payload) => // callback
  ],
  files: [
    (payload) => // callback,
    (payload) => // callback
  ]
}
```

Exports a new `subscribe(channel, callback)` method that subscribes to events and returns the unsubscribe method.

Example:
```javascript
sdk.subscribe("files", payload => {
  console.log(payload);
});
const unsubscribe = sdk.subscribe("documents.603fa2e96ad3b", payload => {
  console.log(payload);
});
unsubscribe(); // removes the channel and the event listening
```


eldadfux:
I think this API might be a bit more intuitive:

```
sdk.subscribe("files", payload => {
  console.log(payload);
});

sdk.unsubscribe("files");  // removes the channel and the event listening

const channels = sdk.channels(); // array of strings with all the open channels

```

What do you think?

TorstenDittmann:
> I think this API might be a bit more intuitive:
> 
> ```
> sdk.subscribe("files", payload => {
>   console.log(payload);
> });
> 
> sdk.unsubscribe("files");  // removes the channel and the event listening
> 
> const channels = sdk.channels(); // array of strings with all the open channels
> 
> ```
> 
> What do you think?

I would disagree, because you might have 10+ callbacks for a single channel and you only want to unsubscribe a single on 🙂

I don't know if the channels method would give any value 🤔

eldadfux:
@TorstenDittmann agree 👍🏻, but still getting a callback as the return value feel a bit less intuitive for me, maybe we can return an identifier and use an interface that looks a bit like this: https://www.w3schools.com/jsref/met_win_cleartimeout.asp ? might make the API a bit more verbose

TorstenDittmann:
@eldadfux this is actually common practice with Observable Patterns used in  many modern libraries like RxJS, Vue, Svelte & co.

A dedicated method would also make scope/reference issues unnecessary, since the unsubscribe method is always with the subscription.

References:
https://vuex.vuejs.org/api/#subscribe
https://svelte.dev/docs#Store_contract
https://rxjs-dev.firebaseapp.com/guide/observable#disposing-observable-executions
https://angular.io/guide/observables-in-angular

eldadfux:
> @eldadfux this is actually common practice with Observable Patterns used in many modern libraries like RxJS, Vue, Svelte & co.
> 
> An included method would also make scope/reference issues unnecessary, since the unsubscribe method is always with the subscription.
> 
> References:
> https://vuex.vuejs.org/api/#subscribe
> https://svelte.dev/docs#Store_contract
> https://rxjs-dev.firebaseapp.com/guide/observable#disposing-observable-executions
> https://angular.io/guide/observables-in-angular

👌🏻

TorstenDittmann:
Still need to add stuff like:
- Keeping the WebSocket alive
- Changing the SDK initialization (for that I am gonna work on the RFC first)

TorstenDittmann:
Refactored the whole script to act more smart and connect depending on a debounce (1ms).

Also need to add event listeners everytime we reconnect the WebSocket - got fixed.

WebSocket will reconnect itself now after 1sec.

eldadfux:
@lohanidamodar do we have a Flutter equivalent PR in the pipeline?

PR #135, opened 2021-03-04
Update test spec

eldadfux:
Following this PR: https://github.com/appwrite/appwrite/pull/955, I have updated the Swagger test file we use to generate tests, the new spec will allow us to generate test SDKs that have new method that will allow us to call the new endpoints added to appwrite/appwrite. For the new endpoint to work we need to merge both PR, and deploy a new version of appwrite.io

### DIFF
--- a/tests/resources/spec.json
+++ b/tests/resources/spec.json
@@ -42,7 +42,7 @@
         "summary": "Mock a get request for SDK tests",
         "operationId": "barGet",
         "consumes": ["application/json"],
-        "produces": ["application/json"],
+        "produces": [],
         "tags": ["bar"],
         "description": "",
         "responses": {
@@ -50,16 +50,17 @@
         },
         "x-appwrite": {
           "method": "get",
-          "weight": 173,
+          "weight": 174,
           "cookies": false,
           "type": "",
-          "demo": "docs/examples/bar/get.md",
+          "demo": "bar/get.md",
           "edit": "https://github.com/appwrite/appwrite/edit/masterMock a get request for SDK tests",
           "rate-limit": 0,
           "rate-time": 3600,
           "rate-key": "url:{url},ip:{ip}",
           "scope": "public",
-          "platforms": ["client", "server"]
+          "platforms": ["client", "server"],
+          "packaging": false
         },
         "security": [{ "Project": [] }],
         "parameters": [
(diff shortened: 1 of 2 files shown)


tests/resources/spec.json
```diff
@@ -593,29 +603,86 @@
...
+        "security": [{ "Project": [] }]
+      }
+    },
     "/mock/tests/general/empty": {
       "get": {
         "summary": "Mock a post request for SDK tests",
         "operationId": "generalEmpty",
         "consumes": ["application/json"],
-        "produces": ["application/json"],
+        "produces": [],
```
lohanidamodar, line 667: Not sure why this was changed everywhere? What's the significance?

tests/resources/spec.json
```diff
@@ -593,29 +603,86 @@
...
+    "/mock/tests/general/400-error": {
+      "get": {
+        "summary": "Mock a an 400 failed request",
+        "operationId": "generalError400",
+        "consumes": ["application/json"],
+        "produces": [],
+        "tags": ["general"],
+        "description": "",
+        "responses": {
+          "500": { "description": "File", "schema": { "type": "file" } }
```
lohanidamodar, line 615: Response, Description and schema seems wrong, this should be 400 not 500 and not file

tests/resources/spec.json
```diff
@@ -593,29 +603,86 @@
...
+    "/mock/tests/general/500-error": {
+      "get": {
+        "summary": "Mock a an 500 failed request",
+        "operationId": "generalError500",
+        "consumes": ["application/json"],
+        "produces": [],
+        "tags": ["general"],
+        "description": "",
+        "responses": {
+          "500": { "description": "File", "schema": { "type": "file" } }
```
lohanidamodar, line 643: Description and schema seems wrong

lohanidamodar (changes requested):

tests/resources/spec.json
eldadfux: 👍🏻 I'll fix that

tests/resources/spec.json
eldadfux: I'll fix that

eldadfux:
Fixed the spec file, this will also help us to test the future response formats :)

lohanidamodar (approved):

tests/resources/spec.json
eldadfux: This is used mainly in the docs, because of a bug it was missing the previous time I generated the spec

PR #137, opened 2021-03-15
Rust SDK

PineappleIOnic:
This PR has Template files and Tests for a Appwrite Rust SDK,
I have also written the foo bar tests and ran them myself to confirm that this SDK works correctly, This SDK Conforms as closely as it can with the [Rust Language's Standards](https://rust-lang.github.io/rfcs/2436-style-guide.html) so it shouldn't compile with any warnings.

Currently the API is blocking I/O but I plan to add async support in the future behind a feature in the crate which I'm quite excited to do considering I'm not very good with async rust.

Thanks for reviewing the PR,
IOnicisere


templates/rust/Cargo.toml.twig
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "appwrite"
+version = "{{ sdk.version }}"
+authors = ["Bradley Schofield <ionicisere@gmail.com>"]
+edition = "2018"
+description = "{{ sdk.shortDescription }}"
+licence = "{{ sdk.license }}"
```
TorstenDittmann, line 7: Use `{{spec.licenseName}}` here 👍🏻

templates/rust/src/services/service.rs.twig
```diff
@@ -0,0 +1,46 @@
...
+  }
+  {% for method in service.methods %}
+
+{% if method.description %}
+{{ method.description|dartComment }}
+{% endif %}
+    pub fn {{ method.name | caseSnake }}(&self{% if method.parameters.all|length >= 1 %}, {% endif %}{% for parameter in method.parameters.all %}{{ parameter.name | caseSnake | escapeKeyword}}: {{ parameter.type | typeName }}{% if not loop.last %}, {% endif %}{% endfor %}) -> reqwest::blocking::Response {
+        let path = "{{ method.path|replace({'{': '', '}': ""}) }}"{% for parameter in method.parameters.path %}.replace("{{ parameter.name | caseSnake }}", &{{ parameter.name | caseSnake }}){% endfor %};
+
+    {% if method.headers %}
```
TorstenDittmann, line 23: Try avoiding indentation for every template syntax from twig. Meaning, the spaces left of `{% if method.headers %}` are still going to be in the generated file. Just remove the spaces so it is just `{% if method.headers %}` in the line 👍🏻 

This needs to be adapted to every line where only twig stuff is in 🙂

TorstenDittmann (changes requested):

TorstenDittmann (approved):

templates/dart/lib/services/service.dart.twig
```diff
@@ -16,11 +16,11 @@ class {{ service.name | caseUcfirst }} extends Service {
     {{ service.name | caseUcfirst }}(Client client): super(client);
 {% for method in service.methods %}
 
-     /// {{ method.title }}
+    /// {{ method.title }}
 {% if method.description %}
-     ///
+    ///
 {{ method.description|dartComment }}
```
lohanidamodar, line 22: dartComment is for dart style comment `///` where as I found that rust uses `//` may be we can introduce new filter?

lohanidamodar (changes requested):

templates/dart/lib/services/service.dart.twig
TorstenDittmann: https://doc.rust-lang.org/beta/rust-by-example/meta/doc.html#doc-comments

templates/rust/README.md.twig
```diff
@@ -0,0 +1,35 @@
...
+{% if sdk.logo %}
+![{{ spec.title }}]({{ sdk.logo }})
+{% endif %}
+
+## Installation
+
+To install via [Crates.io](https://www.crates.io/) add the following to your Cargo.toml under dependencies:
+
+```toml
+appwrite = "1"
```
lohanidamodar, line 21: Can we remove this hardcoded `appwrite` too?

lohanidamodar (changes requested):
Simple change, everything else looks in order.

Also, Is it possible to have example too? Example seems empty.

PineappleIOnic:
I have been working on a playground which will show a bit more of how it works, also rust's documentation tool generates a few examples which will be shown on the crate's doc.rs page once its been uploaded

TorstenDittmann:
> I have been working on a playground which will show a bit more of how it works, also rust's documentation tool generates a few examples which will be shown on the crate's doc.rs page once its been uploaded

The examples we generate are used on the official Appwrite documentation 👍🏻

PineappleIOnic:
I have added template generation aswell as fixing the README.md package name

templates/rust/src/client.rs.twig
```diff
@@ -0,0 +1,254 @@
...
+        match request.send() {
+            Ok(data) => {
+                if data.status().is_success() {
+                    Ok(data)    
+                } else {
+                    // Format error
+                    Err(match data.json() {
+                        Ok(data) => data,
+                        Err(err) => {
+                            AppwriteException::new(format!("A error occoured. ERR: {}, This could either be a connection error or an internal Appwrite error. Please check your Appwrite instance logs. ", err), 0, "".to_string())
```
TorstenDittmann, line 175: Appwrite is hardcoded. Regarding the error message I have to check with the team.

templates/rust/src/client.rs.twig
```diff
@@ -0,0 +1,254 @@
...
+                        }
+                    })
+                }
+            },
+            Err(err) => {
+                // Throw Appwrite Exception
+                Err(AppwriteException::new(format!("A error occoured. ERR: {}, This could either be a connection error or an internal Appwrite error. Please check your Appwrite instance logs. ", err), 0, "".to_string()))
```
TorstenDittmann, line 182: Same here, Appwrite is hardocded.

TorstenDittmann (changes requested):

templates/rust/src/client.rs.twig
```diff
@@ -0,0 +1,254 @@
...
+        request = request.headers(request_headers);
+
+        match request.send() {
+            Ok(data) => {
+                if data.status().is_success() {
+                    Ok(data)    
+                } else {
+                    // Format error
+                    Err(match data.json() {
+                        Ok(data) => data,
```
lohanidamodar, line 173: Once error has occurred should the `Ok(data)` be called? or just throwing exception?

templates/rust/src/client.rs.twig
```diff
@@ -0,0 +1,254 @@
...
+                        Err(err) => {
+                            {{spec.title | caseUcfirst}}Exception::new(format!("A error occoured. ERR: {}, This could either be a connection error or an internal Appwrite error. Please check your Appwrite instance logs. ", err), 0, "".to_string())
```
lohanidamodar, line 175: Here the error message should be the one received from the API `data` i presume.

templates/rust/src/client.rs.twig
```diff
@@ -0,0 +1,254 @@
...
+                        }
+                    })
+                }
+            },
+            Err(err) => {
+                // Throw {{spec.title | caseUcfirst}} Exception
+                Err({{spec.title | caseUcfirst}}Exception::new(format!("A error occoured. ERR: {}, This could either be a connection error or an internal Appwrite error. Please check your Appwrite instance logs. ", err), 0, "".to_string()))
```
lohanidamodar, line 182: May be only throw the underlying err, without extra message as we do with other SDKs

lohanidamodar (changes requested):
Great job @PineappleIOnic Great job so far fixing all the requests. One more thing we could dothe improve the exceptions, making it similar to rest of the SDKs.

Thank you 🙏🏻

templates/rust/src/client.rs.twig
PineappleIOnic: The Ok(data) is for the match phrase called on data.json() since it returns a Option. We have to handle if it's a Ok or Err in order to get to the data

templates/rust/src/client.rs.twig
PineappleIOnic: Yes .json() already turns it into a AppwriteException so we can just return it into the Err() block

eldadfux:
@PineappleIOnic waiting for last confirmation that all changes from the review have been applied, hope we can merge this early next week

PineappleIOnic:
Alright, No worries any more issues let me know and I'll get them fixed ASAP 😄

templates/rust/src/client.rs.twig
lohanidamodar: @PineappleIOnic yes instead of custom message, it should return the error message in the json Data

lohanidamodar (changes requested):
@PineappleIOnic everything looks great, just make when http exception occurs, we are returning the exception with message we get from Appwrite server not a custom message. You can check how we are doing for other SDKs - for example in PHP
https://github.com/appwrite/sdk-generator/blob/feb6d752d9ac4d76c8b20995ff30b18b4870f778/templates/php/src/Client.php.twig#L190

templates/rust/examples/foo_test.rs.twig
```diff
@@ -0,0 +1,74 @@
...
+fn main() {
+    let mut client = {{ language.params.packageName }}::client::Client::new();
+
+    client.set_endpoint("https://appwrite.io/");
+    client.set_project("");
+
+    let foo = {{ language.params.packageName }}::services::Foo::new(&client);
+    let bar = {{ language.params.packageName }}::services::Bar::new(&client);
+    let general = {{ language.params.packageName }}::services::General::new(&client);
+
```
lohanidamodar, line 43: Also add a test-started message first in a new line which is what we use to check the actual test messages as some sdks output some extra outputs. Here's how we are doing for PHP
https://github.com/appwrite/sdk-generator/blob/feb6d752d9ac4d76c8b20995ff30b18b4870f778/tests/languages/php/test.php#L23
```suggestion

```

lohanidamodar (changes requested):

lohanidamodar:
@PineappleIOnic please let us know if you need any help. Thanks!

PineappleIOnic:
> @PineappleIOnic everything looks great, just make when http exception occurs, we are returning the exception with message we get from Appwrite server not a custom message. You can check how we are doing for other SDKs - for example in PHP
> 
> https://github.com/appwrite/sdk-generator/blob/feb6d752d9ac4d76c8b20995ff30b18b4870f778/templates/php/src/Client.php.twig#L190

It does give the error message from the Appwrite server, the following code is only ever thrown when it fails to convert the Appwrite Error JSON into a AppwriteException this will only happen if invalid JSON or a lower level error occurs such as not being able to connect to the server in the first place.
https://github.com/PineappleIOnic/sdk-generator/blob/23e3f9e23896518b0675786952c402d3fc54eeec/templates/rust/src/client.rs.twig#L172-L177

If this isn't a good solution to handle a non parse-able error I'll be happy to figure out alternatives to handle the issue

lohanidamodar:
> > @PineappleIOnic everything looks great, just make when http exception occurs, we are returning the exception with message we get from Appwrite server not a custom message. You can check how we are doing for other SDKs - for example in PHP
> > https://github.com/appwrite/sdk-generator/blob/feb6d752d9ac4d76c8b20995ff30b18b4870f778/templates/php/src/Client.php.twig#L190
> 
> It does give the error message from the Appwrite server, the following code is only ever thrown when it fails to convert the Appwrite Error JSON into a AppwriteException this will only happen if invalid JSON or a lower level error occurs such as not being able to connect to the server in the first place.
> https://github.com/PineappleIOnic/sdk-generator/blob/23e3f9e23896518b0675786952c402d3fc54eeec/templates/rust/src/client.rs.twig#L172-L177
> 
> If this isn't a good solution to handle a non parse-able error I'll be happy to figure out alternatives to handle the issue

If that's the case, then it should be fine.
But even if it's a lower level error, we still get some error message right? and for other server errors, we still get some response from server (might be String?) so why not just show those messages instead of custom error message. It will be more easier for the users to understand?

@PineappleIOnic Check if this can be addressed. Let us know your thoughts. I am however Approving the PR. Should be good enough for our first version.

lohanidamodar (approved):

PineappleIOnic:
> > > @PineappleIOnic everything looks great, just make when http exception occurs, we are returning the exception with message we get from Appwrite server not a custom message. You can check how we are doing for other SDKs - for example in PHP
> > > https://github.com/appwrite/sdk-generator/blob/feb6d752d9ac4d76c8b20995ff30b18b4870f778/templates/php/src/Client.php.twig#L190
> > 
> > 
> > It does give the error message from the Appwrite server, the following code is only ever thrown when it fails to convert the Appwrite Error JSON into a AppwriteException this will only happen if invalid JSON or a lower level error occurs such as not being able to connect to the server in the first place.
> > https://github.com/PineappleIOnic/sdk-generator/blob/23e3f9e23896518b0675786952c402d3fc54eeec/templates/rust/src/client.rs.twig#L172-L177
> > If this isn't a good solution to handle a non parse-able error I'll be happy to figure out alternatives to handle the issue
> 
> If that's the case, then it should be fine.
> But even if it's a lower level error, we still get some error message right? and for other server errors, we still get some response from server (might be String?) so why not just show those messages instead of custom error message. It will be more easier for the users to understand?
> 
> @PineappleIOnic Check if this can be addressed. Let us know your thoughts. I am however Approving the PR. Should be good enough for our first version.

:partying_face: I'm happy to hear that, the lower level error is printed where the {} is if there is one. I can look into the other server errors soon when I get home

lohanidamodar:
@PineappleIOnic I see that in order to skip ssl cert verification you are using the `RUST_TLS_REJECT_UNAUTHORIZED` environment variable, instead of Appwrite's convention of providing a method to (setSelfSigned) to allow self signed certificate.
**Any particular region?**
@eldadfux What do you think?

PineappleIOnic:
Yeah, I ran into a issue with that since once the client has been built on initialisation with ::new() you can not change Reqwest's Client to skip SSL Cert validation. You can only appear to do it on initialisation. I'm open to figuring out a better way of setting that toggle since I myself don't think it's the best way

lohanidamodar (changes requested):

templates/rust/src/services/service.rs.twig
```diff
@@ -0,0 +1,47 @@
...
+            client: client.clone()
+        }
+    }
+{% for method in service.methods %}
+
+{% if method.description %}
+{{ method.description|dartComment }}
+{% endif %}
+    pub fn {{ method.name | caseSnake }}(&self{% if method.parameters.all|length >= 1 %}, {% endif %}{% for parameter in method.parameters.all %}{{ parameter.name | caseSnake | escapeKeyword}}: {{ parameter.type | typeName }}{% if not loop.last %}, {% endif %}{% endfor %}) -> Result<reqwest::blocking::Response, {{spec.title | caseUcfirst}}Exception> {
+        let path = "{{ method.path|replace({'{': '', '}': ""}) }}"{% for parameter in method.parameters.path %}.replace("{{ parameter.name | caseSnake }}", &{{ parameter.name | caseSnake }}){% endfor %};
```
lohanidamodar, line 22:
```suggestion
        let path = "{{ method.path|replace({'{': '', '}': ""}) }}"{% for parameter in method.parameters.path %}.replace("{{ parameter.name | caseCamel }}", &{{ parameter.name | caseSnake }}){% endfor %};
```
Because the Swagger spec, path variables are in camel case

lohanidamodar:
> Yeah, I ran into a issue with that since once the client has been built on initialisation with ::new() you can not change Reqwest's Client to skip SSL Cert validation. You can only appear to do it on initialisation. I'm open to figuring out a better way of setting that toggle since I myself don't think it's the best way

Hey, keeping things in align with other sdk, we could achieve this by reinitializing the reqwest client like below
```rust
pub fn set_self_signed( &mut self, value: bool) {
    self.client = reqwest::blocking::Client::builder().danger_accept_invalid_certs(value).build().unwrap();
}
```

PineappleIOnic:
> 
> 
> > Yeah, I ran into a issue with that since once the client has been built on initialisation with ::new() you can not change Reqwest's Client to skip SSL Cert validation. You can only appear to do it on initialisation. I'm open to figuring out a better way of setting that toggle since I myself don't think it's the best way
> 
> Hey, keeping things in align with other sdk, we could achieve this by reinitializing the reqwest client like below
> 
> In constructor don't override any SSL settings
> 
> ```rust
> Self {
>         endpoint: "https://appwrite.io/v1".parse().unwrap(),
>         headers: new_headers,
>         client: reqwest::blocking::Client::builder()
>         .build().unwrap(),
>        }
> ```
> 
> add, `set_self_signed` function..
> 
> ```rust
> pub fn set_self_signed( &mut self, value: bool) {
>     self.client = reqwest::blocking::Client::builder().danger_accept_invalid_certs(value).build().unwrap();
> }
> ```

I've implemented it and it seems to work well 👍 Good Idea on reinitialising the reqwest client, Thank you.

lohanidamodar (approved):

lohanidamodar:
@PineappleIOnic I also need your thought. Do you think the SDK is ready?

PineappleIOnic:
> @PineappleIOnic I also need your thought. Do you think the SDK is ready for first release?

Yes, I do think it's ready for a first release. I've been using it myself for my twitch project and it's been working well so far. If I find any issues I'll create a PR.

👍

lohanidamodar:
@PineappleIOnic please sync and resolve conflict.

TorstenDittmann (approved):

lohanidamodar (approved):

PineappleIOnic:
Closing this PR as the Response Model PR is closer to the final SDK.

PR #138, opened 2021-03-15
Feat dart test exception

lohanidamodar:


tests/SDKTest.php
```diff
@@ -287,9 +288,20 @@ public function testHTTPSuccess()
                 
                 $this->assertEquals('GET:/v1/mock/tests/general/redirect/done:passed', $output[10]);
                 
-                if($options['supportUpload']) {
+                $supportsUpload = $options['supportUpload'];
```
eldadfux, line 291: Unless we don't use this var in other places, I think we can keep it they way it was.

tests/SDKTest.php
```diff
@@ -287,9 +288,20 @@ public function testHTTPSuccess()
...
+                if($supportsUpload) {
                     $this->assertEquals($output[11], 'POST:/v1/mock/tests/general/upload:passed');
                 }
+                
+                if($options['supportException']) {
```
eldadfux, line 296: Maybe we can leave blank lines reserved for the upload test or even a comment in the code, otherwise every new feature here is going to get a really big block of if conditions.

tests/SDKTest.php
```diff
@@ -200,7 +201,7 @@ public function testHTTPSuccess()
         }
 
         $whitelist = ['php', 'cli', 'java', 'node', 'ruby', 'python', 'typescript', 'deno', 'dotnet', 'dart'];
-
+        $whitelist = ['dart'];
```
eldadfux, line 204: Reminder to remove this from the PR

eldadfux (changes requested):

tests/SDKTest.php
lohanidamodar: using below

tests/SDKTest.php
lohanidamodar: yes we need a way, may be simply print some value ('upload not supported yet') to make the outputs length and index consistent, so that we don't need multiple checks, let me know what you think?

lohanidamodar:
@eldadfux Looks like Java SDK doesn't support UPLOAD yet

TorstenDittmann (approved):
lgtm 👍🏻

lohanidamodar:
@eldadfux even typescript SDK did not have upload support, I just printed the actual output of upload success to pass the typescript test. Apart from that Dart exception test is ✅

eldadfux (approved):

eldadfux:
Looks good!

PR #140, opened 2021-03-23
Refactor Web SDK

TorstenDittmann:
This refactor has following goal:
- A single source code base
- Generates native ESM code
- Works with existing tooling
- Generates a hybrid package for ESM, CJS and IIFE (browser)
- Isomorphic now, meaning it can work in node and web

Users of the ESM distribution get the benefit of increased performance and easier debugging since the Bundler can speak natively to the Appwrite package. CJS is used as a fallback for Node.js and IIFE for Browsers.

The root package.json does not have a `type` property. Rather, we push that down to the package.json files under the ./dist/* sub-directories.

We define an exports map which defines the entry points for the package: one for ESM and one for CJS. 

The only breaking change introduced is, there is not default export anymore.
Reason for this is avoiding bad practice and better DX (autocomplete in IDE, etc).

Meaning, this needs to be changed as following:
`import * as Appwrite from "appwrite";` -> `import { Appwrite } from "appwrite";`

Example of this template generated for 0.7.x can be found here: https://github.com/TorstenDittmann/appwrite-refactor
As published package here: https://www.npmjs.com/package/aw-refactor


eldadfux:
LGTM

eldadfux:
@TorstenDittmann will this changes still work with the CDN setup: `<script src="https://cdn.jsdelivr.net/npm/appwrite@2.0.0"></script>`

TorstenDittmann:
> @TorstenDittmann will this changes still work with the CDN setup: `<script src="https://cdn.jsdelivr.net/npm/appwrite@2.0.0"></script>`

Potentially, have to look into implementing a UMD/AMD/Browser package.

But to be fair, implementing libraries like this today is very unlikely nowadays.

eldadfux:
I think it can help for a quick and dirty test. Anyway let me know what you can do about it and if we should update the docs following this change.

TorstenDittmann:
> I think it can help for a quick and dirty test. Anyway let me know what you can do about it and if we should update the docs following this change.

Okay, we actually don't have to change anything for that und just use https://unpkg.com/.

This is probably the most cleanest way 👍🏻

eldadfux:
> > I think it can help for a quick and dirty test. Anyway let me know what you can do about it and if we should update the docs following this change.
> 
> Okay, we actually don't have to change anything for that und just use https://unpkg.com/.
> 
> This is probably the most cleanest way 👍🏻

Sounds good enough, but we'll need to update both the Appwrite docs, playground, and the README twig template.

TorstenDittmann:
Added Exceptions and a testing environment via Playwright now (thanks @kodumbeats for mentioning that one).

The test runs in Chrome, Firefox and Webkit 👍🏻

TorstenDittmann:
Added native UMD support now for `script:src` usage 👍🏻

trydalch:
Still trying to wrap my mind around this whole process, so forgive me if this is a dumb question.

Will this change provide better return types for each of the operations rather than just Promise<object>? 

If we could add template types for functions, that would be awesome. Like listDocuments<T>(...): Promise<T>

TorstenDittmann:
> Still trying to wrap my mind around this whole process, so forgive me if this is a dumb question.
> 
> Will this change provide better return types for each of the operations rather than just Promise?
> 
> If we could add template types for functions, that would be awesome. Like `listDocuments<T>(...): Promise<T>`

Not yet, but among many improvements and that we are moving natively to Typescript we are building the foundation for it.

lohanidamodar:
> Still trying to wrap my mind around this whole process, so forgive me if this is a dumb question.
> 
> Will this change provide better return types for each of the operations rather than just Promise?
> 
> If we could add template types for functions, that would be awesome. Like `listDocuments<T>(...): Promise<T>`

@trydalch We already have a RFC in progress, you can follow along and provide feedback
https://github.com/appwrite/rfc/pull/19

TorstenDittmann:
> Still trying to wrap my mind around this whole process, so forgive me if this is a dumb question.
> 
> Will this change provide better return types for each of the operations rather than just Promise?
> 
> If we could add template types for functions, that would be awesome. Like `listDocuments<T>(...): Promise<T>`

Just introduced that generic type 👍🏻 Even if we don't have Response Models yet, this should make it friendlier for devs right now.

src/SDK/SDK.php
```diff
@@ -128,14 +128,14 @@ public function __construct(Language $language, Spec $spec)
         $this->twig->addFilter(new TwigFilter('comment2', function ($value) {
             $value = explode("\n", $value);
             foreach ($value as $key => $line) {
-                $value[$key] = "         * " . wordwrap($value[$key], 75, "\n         * ");
+                $value[$key] = "     * " . wordwrap($value[$key], 75, "\n     * ");
```
eldadfux, line 131: Are we sure this doesn't affect other langs?

src/SDK/SDK.php
```diff
@@ -128,14 +128,14 @@ public function __construct(Language $language, Spec $spec)
...
             }
             return implode("\n", $value);
         }, ['is_safe' => ['html']]));
         $this->twig->addFilter(new TwigFilter('comment3', function ($value) {
             $value = explode("\n", $value);
             foreach ($value as $key => $line) {
-                $value[$key] = "             * " . wordwrap($value[$key], 75, "\n             * ");
+                $value[$key] = "         * " . wordwrap($value[$key], 75, "\n         * ");
```
eldadfux, line 138: Are we sure this doesn't affect other langs?

src/SDK/SDK.php
TorstenDittmann: Yes, `comment2` and `comment3` is only used in the Web SDK 👍🏻

PR #144, opened 2021-03-28
feat-flutter-tests

lohanidamodar:


lohanidamodar:
@eldadfux  please restart the tests

eldadfux:
Done

lohanidamodar:
This also fixes https://github.com/appwrite/sdk-for-flutter/issues/16. So requires more manual testing.

templates/flutter/lib/client.dart.twig
```diff
@@ -73,22 +74,21 @@ class Client {
...
-          }else{
-            // if web set httpClientAdapter as BrowserHttpClientAdapter with withCredentials true to make cookies work
-            this.http.options.extra['withCredentials'] = true;
-          }
-
-          this.http.options.baseUrl = this.endPoint;
-          this.http.options.validateStatus = (status) => status < 400;
+        // if web skip cookie implementation and origin header as those are automatically handled by browsers
+        if(!kIsWeb) {
+        final Directory cookieDir = await _getCookiePath();
```
TorstenDittmann, line 79: indentation is off here.

templates/flutter/lib/client.dart.twig
```diff
@@ -73,22 +74,21 @@ class Client {
...
+        cookieJar = new PersistCookieJar(dir:cookieDir.path);
+        this.http.interceptors.add(CookieManager(cookieJar));
+        PackageInfo packageInfo = await PackageInfo.fromPlatform();
+        addHeader('Origin', 'appwrite-$type://${packageInfo.packageName ?? packageInfo.appName}');
+        }else{
```
TorstenDittmann, line 84: Missing whitespace

tests/SDKTest.php
```diff
@@ -261,6 +273,12 @@ public function testHTTPSuccess()
                 }
 
                 $this->assertIsArray($output);
+                
+                $removed = '';
+                do {
+                    $removed = array_shift($output);
+                }while($removed != 'Test Started' && sizeof($output) != 0);
```
TorstenDittmann, line 280: missing whitespace

ryanhlewis:
How is checking for !initialized every Call method different from the previous way of calling a method every Call to check for it? 

There's not much difference with this application of it, other than calling init() once within the constructor. 

I suggest the (!initialized) init() call be placed within the specific web request of Account.cs CreateSession(). The Call() should be made, and the cookies initialized by evaluating the results of the CreateSession() task temporarily.

This alleviates the !initialized boolean altogether, as it is never necessary. Should the cookie exist, it will be present in the cookie container already. Should it not exist, the user will have to login anyway. 

I've included some pseudo code below. I used this code for my custom Dotnet client library. 

(PlayerPrefs is a Serializable Class in Unity. It is just storing the value. You handled this already in Flutter by creating your init() method to encompass it.)


```
           var login = await _client.Call("POST", path, headers, parameters);

            //Unity method to wait the function until the web request is completed. 
            yield return new WaitUntil(() => login.IsCompleted);

            //SPECIAL- The user has logged in. Grab their cookies!
            string cookie = login.Result.Headers.GetValues("Set-Cookie").First();

            PlayerPrefs.SetString("loginCookie", cookie.Substring(0,cookie.IndexOf(';')));

            return login;

```
The above code is tested and works. The below code is mostly pseudo. 

With your Flutter code, I would think it would go like..

```
var login = await client.call(HttpMethod.post, path: path, params: params, headers: headers);

//Flutter await completed is a .then() 
//.then implementation

client.init();

return login; 

```


The only issue I find within my own implementation is that it does not account for OAuth, but similarly, I do not see why the OAuth call cannot be taken, split, and the cookie stolen in the exact same way. 

TLDR: You do not need an initalized boolean nor an init() call in your Call() method if you know exactly where the user logins.

lohanidamodar:
@ryanhlewis I did try your suggestion, here, https://github.com/lohanidamodar/sdk-for-flutter/tree/feat-null-safety-new

However, the issue is, the loading of persisted cookies is done asynchronously, so when the user first instantiates client, we do call init() from the constructor, however we don't know if it's completed and cannot wait for the async event in the constructor. And if the cookie is persisted, but if the user makes any request before the init method has completed user will get permission error. And the workaround will be for the user themselves to call and await the `init` method. I don't think this is good either. So I couldn't find a better alternative than checking whether or not the cookie is loaded from the cookie jar before making the call, because every call must be ensured to have the cookie if they are persisted.

If might have worked for you, from the looks of it, I think you are persisting the cookie and passing it in the following request. The flutter SDK handles that internally, so to ensure that the cookie is passed in every request, we do need to check if it's initialized i.e. the cookie is loaded.

If you have a better solution do suggest.

Thank you

PR #145, opened 2021-03-29
Changed auth security examples

eldadfux:
Changed the way we generate client authentication to support the new JWT


lohanidamodar (commented):
@eldadfux all changes looks good. However I have a question, the new spec file for 8.0, for some of the endpoints the `auth/security` only contains `project` and `JWT` but not `key`.
Isn't with `key` user should be allowed to perform all the operations? How does it work?

eldadfux:
Account service only operation are not allowed by an API key because they can only be performed on behalf of a user.

PR #146, opened 2021-03-29
Added discord and twitter badges

eldadfux:
Just a small update to the badges, really bothered me hehe....


lohanidamodar (approved):
LGTM.
However I am thinking, Isn't this same for all the SDKs?
If yes why are we repeating it for each readme file, why not add it like we added warning or getting started section?

eldadfux:
To create more flexibility, some SDKs are using lang specific badges - which is totally fine

PR #156, opened 2021-04-16
feat: added new array params

christyjacob4:
- Support for new CLI array params
- Updated README and examples to use the new array syntax

### DIFF
--- a/templates/cli/README.md.twig
+++ b/templates/cli/README.md.twig
@@ -82,8 +82,8 @@ $ {{ language.params.executableName }} database createDocument --collectionId="Y
 
 ### Some Gotchas
 - `data` expects the JSON string to be escaped.
-- If using the wildcard (`*`) read or write permissions , make sure that it is properly escaped using a `\` or by enclosing it in `"*"`.
-- Some arguments like `read` and `write` permissions are expected to be arrays. In the {{ spec.title }} CLI arrays are passed by simply repeating the argument as seen in the `createDocument` example above.
+- If using the wildcard (`*`) read or write permissions , make sure that it is properly escaped using a `\` or by enclosing it in `"*"` since bash interprets them differently.
+- Some arguments like `read` and `write` permissions are expected to be arrays. In the {{ spec.title }} CLI, arrays are passed by simply repeating the argument as seen in the `createDocument` example above.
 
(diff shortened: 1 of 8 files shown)


templates/cli/README.md.twig
```diff
@@ -75,6 +75,16 @@ $ {{ language.params.executableName }} users create --email="hello@{{ spec.title
...
 
+To create a Document you can use the following command 
+```sh
+$ {{ language.params.executableName }} database createDocument --collectionId="YOUR COLLECTION ID" --data='A VALID JSON STRING' --read=role:member --read="*" --write=role:guest
+```
+
+### Some Gotchas
+- `data` expects the JSON string to be escaped.
+- If using the wildcard (`*`) read or write permissions , make sure that it is properly escaped using a `\` or by enclosing it in `"*"`.
+- Some arguments like `read` and `write` permissions are expected to be arrays. In the {{ spec.title }} CLI arrays are passed by simply repeating the argument as seen in the `createDocument` example above.
```
eldadfux, line 86: In the {{ spec.title }} CLI -> `In the {{ spec.title }} CLI, `

tests/SDKTest.php
```diff
@@ -200,7 +200,8 @@ public function testHTTPSuccess()
             throw new \Exception('Failed to fetch spec from Appwrite server');
         }
 
-        $whitelist = ['php', 'cli', 'node', 'ruby', 'python', 'typescript', 'deno', 'dotnet', 'dart', 'flutter'];
+        // $whitelist = ['php', 'cli', 'node', 'ruby', 'python', 'typescript', 'deno', 'dotnet', 'dart', 'flutter'];
+        $whitelist = ['cli'];
```
eldadfux, line 204: Need to replace these lines :)

eldadfux (changes requested):

lohanidamodar (approved):
Apart from the changes @eldadfux requested. Everything else LGTM.

tests/SDKTest.php
christyjacob4: Done :D

templates/cli/README.md.twig
christyjacob4: Done :D

eldadfux (approved):

PR #163, opened 2021-05-05
fix(web): conditional json parsing

TorstenDittmann:
- prevent parsing servers response with json, if there is no content.
- add 502 error test


eldadfux:
We need to make sure to enable the 502 check on the e2e php test file

PR #176, opened 2021-05-18
feat(web): realtime prototype

TorstenDittmann:
Check #132 for previous discussion. This is based on the new refactored Web SDK.

Before refactoring the whole init process, I just added a `setEndpointRealtime` method for this.

Adds a new `realtime` object which contains the following:
- a socket property that holds the WebSocket connection
- a channels property that is an object, with each key representing the channel name holding an array of callbacks looking like this:
```
channels: {
  documents: [
    (payload) => // callback
  ],
  files: [
    (payload) => // callback,
    (payload) => // callback
  ]
}
```

Exports a new `subscribe(channel, callback)` method that subscribes to events and returns the unsubscribe method.

Example:
```javascript
sdk.subscribe("files", payload => {
  console.log(payload);
});
const unsubscribe = sdk.subscribe("documents.603fa2e96ad3b", payload => {
  console.log(payload);
});
unsubscribe(); // removes the channel and the event listening
```


templates/web/src/sdk.ts.twig
```diff
@@ -9,6 +9,55 @@ type Headers = {
...
+    to: string;
+    success: boolean;
+    user: object;
+}
+
+type RealtimeRequestAuthenticate = {
+    session: string;
+}
+
+type RealtimeObject = {
```
lohanidamodar, line 49: Why not just `Realtime` instead of `RealtimeObject`? will be more in sync with other SDKs as well?

lohanidamodar (approved):
Small comment, rest looks good.

templates/web/src/sdk.ts.twig
TorstenDittmann: Good idea 👍🏻 

Will do.

abnegate (approved):
Looks good 👍

templates/web/src/sdk.ts.twig
```diff
@@ -9,6 +9,55 @@ type Headers = {
...
+type RealtimeRequest = {
+    type: "authentication";
+    data: RealtimeRequestAuthenticate;
+}
+
+type RealtimeResponseEvent = {
+    event: string;
+    channels: string[];
+    timestamp: number;
+    payload: unknown;
```
Meldiron, line 26: Would it be possible to use generic types? Usage would be something like:

```typescript
type PostsCollection = { title: string; text: string; likes: number };

sdk.subscribe<PostsCollection>("files", payload => {
  console.log(payload.tilte); // ERROR: Type of PostsCollection does not have 'tite' property. Did you mean 'title'?
});
```

templates/web/src/sdk.ts.twig
```diff
@@ -9,6 +9,55 @@ type Headers = {
...
+}
+
+type RealtimeResponseError = {
+    code: number;
+    message: string;
+}
+
+type RealtimeResponseConnected = {
+    channels: string[];
+    user?: object;
```
Meldiron, line 36: Can we type `user` objects too?

templates/web/src/sdk.ts.twig
TorstenDittmann: Sadly not, since you are able to subscribe to multiple channels👍🏻 

```typescript
sdk.subscribe<PostsCollection>(["files", "account"], payload => {
  console.log(payload.tilte); // ERROR: Type of PostsCollection does not have 'tite' property. Did you mean 'title'?
});
```

templates/web/src/sdk.ts.twig
TorstenDittmann: This object will not be exposed to any public method yet and is only used internally. This will be added when we add complete support for response models.

templates/web/src/sdk.ts.twig
TorstenDittmann: Anyway, I think we could still add it. Will be in the users responsibilty to use it correctly.

Meldiron: I see. I still think it is important.. Hmm 🤔 Can we have 2 subscribe methods? One for `string`, one for `string[]`.
This way we would type only the `string` one with a generic type, so people can use it if they want. I would personally like to split events into 10 subscriptions than be without types for my collections.

Another idea, what about this one?

```typescript
type PostsCollection = { title: string; text: string; likes: number };
type LikesCollection = { title: string; text: string; likes: number };

sdk.subscribe<PostsCollection | LikesCollection>(["collection:posts", "collection:likes"], (payload) => {
  // stuff
});
```

If you need multiple subscriptions inside one subscribe method, put in both types. Then it is up to you to know which one you have.. But the good solution is to make one `.subscribe` for one type.. The array solution seems good only if you have same or similar models for multiple collections.

TorstenDittmann: Added generic type, if they add multiple channels with different types - its the devs task to handle it for now 👍🏻

PR #178, opened 2021-05-19
change dart images for test

lohanidamodar:


lohanidamodar:
@eldadfux just updated the test images, everything is passing in the latest versions. Will run manual tests on generated SDKs.

eldadfux:
Don't we want to add this on top of current fixed versions?

lohanidamodar:
> Don't we want to add this on top of current fixed versions?

Nope same version, but official image. dart has now docker official images instead of Google ones

eldadfux:
My question was should we use latest & beta versions vs, fixed number version or both together. It's not always a great practice to test only vs latest versions. Gives us not control over unexpected changes.

lohanidamodar:
@eldadfux at that point the stable was `2.12` and beta was the `2.13`, now that stable is `2.13`, I have updated accordingly.

eldadfux (approved):

PR #184, opened 2021-05-21
required keyword and type_number

lohanidamodar:


src/SDK/Language/Dart.php
```diff
@@ -126,6 +127,8 @@ public function getTypeName($type)
             	return 'List';
 			case self::TYPE_OBJECT:
 				return 'Map';
```
eldadfux, line 129:
```suggestion
	case self::TYPE_OBJECT:
		return 'Map';
```

lohanidamodar:
@eldadfux the indentation problem is like before only github shows it as messed up. applying your suggestion actually made it worse, so I had to push again. Doesn't look right on the Github changes, you should check locally

PR #186, opened 2021-05-21
fix-cli-code-examples

lohanidamodar:


TorstenDittmann (commented):
We might run into issues here, since `\` is not supported on windows.

lohanidamodar:
> We might run into issues here, since `\` is not supported on windows.

@eldadfux we might need separate example for windows and unix?

eldadfux:
Still not sure how to do it, maybe we can create 2 separate folders in the CLI repo, and adjust the docs to handle multiple variations per SDK?

eldadfux:
I would still keep the unix one as is for easier transition and just add another `examples-cmd` / `examples/powershell` examples

PR #189, opened 2021-05-23
fix(web): implement polyfills by default and not conditional

TorstenDittmann:


christyjacob4 (approved):
Awesome Torsten! Looks Good 👍

PR #190, opened 2021-05-27
fix(node): proper response type for location requests

TorstenDittmann:
- fixes response type in the request config when request is of type `location`
- prevents encoding on binary files returned from the SDK method


eldadfux:
Can't we use the `content-type` header for this? It might be a more "standardized" way to handle this bug.

PR #193, opened 2021-05-31
fix(node): overall improvements

TorstenDittmann:
- add typescript definitions
- add parameter checks

Typescript definition file example: https://gist.github.com/TorstenDittmann/61a1c1fc4fd7f82b823c092a473ebdae

before:
```js
    /**
     * Update Document
     *
     * Update a document by its unique ID. Using the patch method you can pass
     * only specific fields that will get updated.
     *
     * @param string collectionId
     * @param string documentId
     * @param object data
     * @param string[] read
     * @param string[] write
     * @throws Exception
     * @return {}
     */
    async updateDocument(collectionId, documentId, data, read = [], write = []) {
        let path = '/database/collections/{collectionId}/documents/{documentId}'.replace(new RegExp('{collectionId}', 'g'), collectionId).replace(new RegExp('{documentId}', 'g'), documentId);
        
        return await this.client.call('patch', path, {
               },
               {
                'data': data,
                'read': read,
                'write': write
            });
    }
```

after:
```js
    /**
     * Update Document
     *
     * Update a document by its unique ID. Using the patch method you can pass
     * only specific fields that will get updated.
     *
     * @param {string} collectionId
     * @param {string} documentId
     * @param {object} data
     * @param {string[]} read
     * @param {string[]} write
     * @throws {AppwriteException}
     * @returns {Promise}
     */
    async updateDocument(collectionId, documentId, data, read, write) {
        if (typeof collectionId === undefined) {
            throw new AppwriteException('Missing required parameter: "collectionId"');
        }

        if (typeof documentId === undefined) {
            throw new AppwriteException('Missing required parameter: "documentId"');
        }

        if (typeof data === undefined) {
            throw new AppwriteException('Missing required parameter: "data"');
        }

        let path = '/database/collections/{collectionId}/documents/{documentId}'.replace('{collectionId}', collectionId).replace('{documentId}', documentId);
        let payload = {};

        if (typeof data !== undefined) {
            payload['data'] = data;
        }

        if (typeof read !== undefined) {
            payload['read'] = read;
        }

        if (typeof write !== undefined) {
            payload['write'] = write;
        }

        return await this.client.call('patch', path, {
        }, payload);
    }
```


templates/node/lib/services/service.js.twig
```diff
@@ -11,32 +12,33 @@ class {{ service.name | caseUcfirst }} extends Service {
...
-        let path = '{{ method.path }}'{% for parameter in method.parameters.path %}.replace(new RegExp('{{ '{' }}{{ parameter.name | caseCamel }}{{ '}' }}', 'g'), {{ parameter.name | caseCamel }}){% endfor %};
-        
+    async {{ method.name | caseCamel }}({% for parameter in method.parameters.all %}{{ parameter.name | caseCamel }}{% if not loop.last %}, {% endif %}{% endfor %}) {
+        let path = '{{ method.path }}'{% for parameter in method.parameters.path %}.replace('{{ '{' }}{{ parameter.name | caseCamel }}{{ '}' }}', {{ parameter.name | caseCamel }}){% endfor %};
+        let payload = {};
+{% for parameter in method.parameters.all %}
+
+        if ({{ parameter.name | caseCamel }} !== undefined) {
+            payload['{{ parameter.name }}'] = {{ parameter.name | caseCamel }};
+        }{% if parameter.required %} else {
```
eldadfux, line 28: Won't we get a Node error if a required param is missing from the method call anyway?

TorstenDittmann: There is not such thing as a required parameter for Node.js or Javascript 🙂 

If a param is not used - it is automatically undefined.

Bildschirmfoto 2021-06-01 um 10 01 51

TorstenDittmann: There is not such thing as a required parameter for Node.js or Javascript 🙂

If a param is not used - it is automatically undefined.

Bildschirmfoto 2021-06-01 um 10 05 59

PR #198, opened 2021-06-02
fix(python): overall improvements

TorstenDittmann:
- add param checking

before:
```python
def update_document(self, collection_id, document_id, data, read=[], write=[]):
    """Update Document"""

    params = {}
    path = '/database/collections/{collectionId}/documents/{documentId}'
    path = path.replace('{collectionId}', collection_id)                
    path = path.replace('{documentId}', document_id)                
    params['data'] = data
    params['read'] = read
    params['write'] = write

    return self.client.call('patch', path, {
    }, params)
```

after:
```python
def update_document(self, collection_id, document_id, data, read = None, write = None):
    """Update Document"""

    if collection_id is None: 
        raise AppwriteException('Missing required parameter: "collection_id"')

    if document_id is None: 
        raise AppwriteException('Missing required parameter: "document_id"')

    if data is None: 
        raise AppwriteException('Missing required parameter: "data"')

    params = {}
    path = '/database/collections/{collectionId}/documents/{documentId}'
    path = path.replace('{collectionId}', collection_id)                
    path = path.replace('{documentId}', document_id)                

    if data is not None: 
        params['data'] = data

    if read is not None: 
        params['read'] = read

    if write is not None: 
        params['write'] = write

    return self.client.call('patch', path, {
    }, params)
```

### DIFF
--- a/templates/python/package/services/service.py.twig
+++ b/templates/python/package/services/service.py.twig
@@ -1,5 +1,5 @@
 from ..service import Service
-
+from ..exception import AppwriteException
 
 class {{ service.name | caseUcfirst }}(Service):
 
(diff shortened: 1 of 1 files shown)


christyjacob4 (approved):
Looks Good @TorstenDittmann

PR #205, opened 2021-06-03
Static flatten and Code Cleanup

Aliyss:
@TorstenDittmann Hope this is correct here.

[Reference  1](https://github.com/appwrite/sdk-for-node/pull/15)
[Reference 2](https://github.com/appwrite/sdk-for-node/pull/14)


TorstenDittmann:
@Aliyss Now that `flatten` is static - we can't call it with `this` anymore. You would need to adapt the calls as well 🙂

TorstenDittmann (approved):

eldadfux:
Thanks @Aliyss

PR #206, opened 2021-06-05
Updated coding conventions

eldadfux:
- `protected` are usually dev friendlier than `private` method
- Fixed some indentations and coding standards
- Arrays should be casted as `array` and not `Array`


TorstenDittmann (approved):

christyjacob4 (approved):
Something new to learn everyday! :100:

PR #213, opened 2021-06-08
gravity support in flutter sdk

lohanidamodar:
Added Gravity enum to support


lohanidamodar:
as we are removing enums this should no longer be required.

PR #214, opened 2021-06-09
remove enums support in Flutter/Dart

lohanidamodar:


PineappleIOnic (approved):

templates/flutter/lib/enums.dart.twig
```diff
@@ -7,11 +7,3 @@ extension HttpMethodString on HttpMethod {
     return this.toString().split('.').last.toUpperCase();
   }
 }
```
eldadfux, line 9: Don't we need to completely remove this template file?

templates/dart/lib/enums.dart.twig
```diff
@@ -7,11 +7,3 @@ extension HttpMethodString on HttpMethod {
     return this.toString().split('.').last.toUpperCase();
   }
 }
```
eldadfux, line 9: Don't we need to completely remove this template file?

templates/flutter/lib/enums.dart.twig
lohanidamodar: this HttpMethod is used internally, so required.

templates/dart/lib/enums.dart.twig
lohanidamodar: Same, HttpMethod emum is used internally and required

PR #217, opened 2021-06-16
feat: sdk for kotlin

christyjacob4:
* Added templates for kotlin
* Removed unused files from android templates
* Added Kotlin tests
* Setup Github actions for Kotlin SDK
* Fix default and example parameters in Kotlin.php


templates/kotlin/src/main/kotlin/io/appwrite/servicesTest/ServicesTest.kt.twig
```diff
@@ -0,0 +1,57 @@
...
+}
+{% endfor %}
+
+
+
+
+
+
+
+
```
eldadfux, line 57:
```suggestion
```

PR #220, opened 2021-06-18
fix oauth null param error

lohanidamodar:
- Fixing issues with optional param in createOAuth2Session reported here https://github.com/appwrite/sdk-for-flutter/issues/23 This occured as we removed default values from SDK
- Fixing issues in client when params is empty while using appwrite client externally as reported here https://github.com/appwrite/sdk-for-flutter/issues/24 - this is for those who are using Client for features that are not yet implemented in the SDK and is under development in Appwrite.


eldadfux:
@lohanidamodar can you share more about the problem and the fix?

templates/flutter/lib/client.dart.twig
```diff
@@ -134,7 +134,9 @@ class Client {
             await this.init();
         }
 
-        params.removeWhere((key,value) => value == null);
+        if(params.isNotEmpty()) {
```
JulianBissekkou, line 137: In dart this is a getter and not a function ☺️ 

```suggestion
        if(params.isNotEmpty) {
```

templates/dart/lib/client.dart.twig
```diff
@@ -68,7 +68,9 @@ class Client {
             };
         }
 
-        params.removeWhere((key,value) => value == null);
+        if(params.isNotEmpty()) {
```
JulianBissekkou, line 71:
```suggestion
        if(params.isNotEmpty) {
```

templates/flutter/lib/client.dart.twig
lohanidamodar: Thanks

PR #221, opened 2021-06-23
fixed: non-ASCII char in request header

kutear:
> Historically, HTTP has allowed field content with text in the
ISO-8859-1 charset [ISO-8859-1], supporting other charsets only
through use of [RFC2047] encoding. In practice, most HTTP header
field values use only a subset of the US-ASCII charset [USASCII].
Newly defined header fields SHOULD limit their field values to
US-ASCII octets. A recipient SHOULD treat other octets in field
content (obs-text) as opaque data.

But `appName` may be contain non-ASCII char. so replace with `package`

ref: https://github.com/appwrite/sdk-for-flutter/issues/25


lohanidamodar (approved):
Seems logical. Thank you for your PR

lohanidamodar:
@kutear can you please sync with latest master so that we can see the tests passing. Thank you 🙏🏻

kutear:
> @kutear can you please sync with latest master so that we can see the tests passing. Thank you 🙏🏻

I had merge `master` and ci was passed.

eldadfux:
Thanks @kutear !

PR #234, opened 2021-07-22
feat-new-db-query

lohanidamodar:
New class for updated database query parameters


templates/dart/lib/services/service.dart.twig
```diff
@@ -1,11 +1,16 @@
 part of {{ language.params.packageName }};
 
 {% macro parameter(parameter) %}
-{% if parameter.required %}required {{ parameter.type | typeName }}{% else %}{{ parameter.type | typeName }}?{% endif %} {{ parameter.name | caseCamel }}{% endmacro %}
+{% if parameter.name == 'queries' %}Query? queries{% else %}
```
TorstenDittmann, line 4: How safe is depending on the name itself for us? 🤔

TorstenDittmann (approved):
Looks good to me 👍🏻

templates/dart/lib/services/service.dart.twig
lohanidamodar: we did it before for orderType, but we removed that enum

eldadfux (changes requested):
I don't feel comfortable at all with relying on the param name for using the new Query filter. We can and need to change the swagger format so we add and x-something attribute to indicate this param has special capabilities.

lohanidamodar:
@eldadfux This is how the listDocuments method looks
```dart
Future<Response> listDocuments({required String collectionId, Query? queries, int? limit, int? offset, List? orderAttributes, List? orderTypes}) {
        final String path = '/database/collections/{collectionId}/documents'.replaceAll(RegExp('{collectionId}'), collectionId);

        final Map<String, dynamic> params = {
            'queries': queries?.queries,
            'limit': limit,
            'offset': offset,
            'orderAttributes': orderAttributes,
            'orderTypes': orderTypes,
        };

        final Map<String, String> headers = {
        };

        return client.call(HttpMethod.get, path: path, params: params, headers: headers);
    }
```

templates/dart/lib/query.dart.twig
```diff
@@ -0,0 +1,50 @@
+part of {{ language.params.packageName }};
+
+class Query {
```
lohanidamodar, line 3: @eldadfux this is how the query class looks

eldadfux:
> @eldadfux This is how the listDocuments method looks
> 
> ```dart
> Future<Response> listDocuments({required String collectionId, Query? queries, int? limit, int? offset, List? orderAttributes, List? orderTypes}) {
>         final String path = '/database/collections/{collectionId}/documents'.replaceAll(RegExp('{collectionId}'), collectionId);
> 
>         final Map<String, dynamic> params = {
>             'queries': queries?.queries,
>             'limit': limit,
>             'offset': offset,
>             'orderAttributes': orderAttributes,
>             'orderTypes': orderTypes,
>         };
> 
>         final Map<String, String> headers = {
>             'content-type': 'application/json',
>         };
> 
>         return client.call(HttpMethod.get, path: path, params: params, headers: headers);
>     }
> ```

Would be great if you could also share an example of how it will be called with multiple queries.

lohanidamodar:
> > @eldadfux This is how the listDocuments method looks
> > ```dart
> > Future<Response> listDocuments({required String collectionId, Query? queries, int? limit, int? offset, List? orderAttributes, List? orderTypes}) {
> >         final String path = '/database/collections/{collectionId}/documents'.replaceAll(RegExp('{collectionId}'), collectionId);
> > 
> >         final Map<String, dynamic> params = {
> >             'queries': queries?.queries,
> >             'limit': limit,
> >             'offset': offset,
> >             'orderAttributes': orderAttributes,
> >             'orderTypes': orderTypes,
> >         };
> > 
> >         final Map<String, String> headers = {
> >             'content-type': 'application/json',
> >         };
> > 
> >         return client.call(HttpMethod.get, path: path, params: params, headers: headers);
> >     }
> > ```
> 
> Would be great if you could also share an example of how it will be called with multiple queries.

```dart
db.listDocuments(collectionId: 'test', queries: Query().equal('director','Charles').contains('cast', ['Arnold', 'Vin']).lesser('release_year', 2005), limit: 5);
```

lohanidamodar:
@eldadfux I just remembered, we could add a build method in the query class that will return the queries as list of strings so above example can be like
```
db.listDocuments(
    collectionId: 'test',
    queries: Query()
        .equal('director','Charles')
        .contains('cast', ['Arnold', 'Vin'])
        .lesser('release_year', 2005)
        .build(),
    limit: 5,
);
```
Doing this, I could change the listDocuments to 

```dart
Future<Response> listDocuments({required String collectionId, List<String> queries, int? limit, int? offset, List? orderAttributes, List? orderTypes}) {
        final String path = '/database/collections/{collectionId}/documents'.replaceAll(RegExp('{collectionId}'), collectionId);

        final Map<String, dynamic> params = {
            'queries': queries,
            'limit': limit,
            'offset': offset,
            'orderAttributes': orderAttributes,
            'orderTypes': orderTypes,
        };

        final Map<String, String> headers = {
        };

        return client.call(HttpMethod.get, path: path, params: params, headers: headers);
    }
```

eldadfux:
@lohanidamodar @TorstenDittmann @kodumbeats what approach would you like better:

1. shorter syntax, easier to prepare internally and convert to an HTTP call: 
```dart
db.listDocuments(
    collectionId: 'test',
    queries: Query()
        .equal('director','Charles')
        .contains('cast', ['Arnold', 'Vin'])
        .lesser('release_year', 2005),
    limit: 5,
);
```

2. Longer, but more verbose, better relationship between `queries` and the `Query` class. More obvious this is an array.

```dart
db.listDocuments(
    collectionId: 'test',
    queries: [
        Query().equal('director','Charles'),
        Query().contains('cast', ['Arnold', 'Vin']),
        Query().lesser('release_year', 2005),
    ]
    limit: 5,
);
```

lohanidamodar:
With the new query class calling listDocuments looks like
```dart
db.listDocuments(
    collectionId: 'test',
    queries: [
        Query.equal('director','Charles'),
        Query.contains('cast', ['Arnold', 'Vin']),
        Query.lesser('release_year', 2005),
    ]
    limit: 5,
);
```

templates/web/src/query.ts.twig
```diff
@@ -0,0 +1,37 @@
+class Query {
+  static equal(attribute: string, value: any): string {
```
TorstenDittmann, line 2: I would replace `any` with the actual types we allow 🤔 

Lets add a type like this 
```ts
type QueryTypes = string | string[] | number | number[] | boolean | boolean[];
```

and we can use it like this
```ts
static equal(attribute: string, value: QueryTypes) {

}
```

templates/web/src/query.ts.twig
```diff
@@ -0,0 +1,37 @@
...
+
+  static greaterEqual(attribute: string, value: any): string {
+    return Query.addQuery(attribute, 'greaterEqual', value);
+  }
+
+  static contains(attribute: string, value: any): string {
+    return Query.addQuery(attribute, 'contains', value);
+  }
+
+  static addQuery(attribute: string, oper: string, value: any): string {
```
TorstenDittmann, line 30:
```suggestion
  private static addQuery(attribute: string, oper: string, value: any): string {
```

templates/web/src/query.ts.twig
```diff
@@ -0,0 +1,37 @@
...
+    if (typeof value === 'string') {
```
TorstenDittmann, line 31: How de we handle arrays here?

lohanidamodar: should work automatically.

TorstenDittmann: I don't think so.

```ts
Query.equal("attribute", ["test", "with,comma"])
```
will result in 
```ts
'attribute.equal(test,with,comma)'
```

templates/android/library/src/main/java/io/appwrite/Query.kt.twig
```diff
@@ -0,0 +1,41 @@
...
+    fun contains(attribute: String, value: List<Any>): String {
+      return Query.addQuery(attribute, "contains", value);
+    }
+
+    fun addQuery(attribute: String, oper: String, value: Any): String {
+      if (value is String) {
+        return "${attribute}.${oper}(\"${value}\")";
+      } else {
+        return "${attribute}.${oper}(${value})";
+      }
```
abnegate:
```suggestion
      return when (value) {
         is String -> "${attribute}.${oper}(\"${value}\")"
         else -> "${attribute}.${oper}(${value})"
       }
```

templates/android/library/src/main/java/io/appwrite/Query.kt.twig
```diff
@@ -0,0 +1,41 @@
+package {{ sdk.namespace | caseDot }}
+
+class Query {
+  companion object {
+    fun equal(attribute: String, value: Any): String {
+      return Query.addQuery(attribute, "equal", value);
+    }
```
abnegate, line 7: These functions could be expression bodied:

```suggestion
    fun equal(attribute: String, value: Any) -> addQuery(attribute, "equal", value)
```

TorstenDittmann (approved):
Web and Deno lgtm 👍🏻

PR #235, opened 2021-07-23
php SDK chunked upload

lohanidamodar:
- Breaking Change : File is now a String, path to file instead of CURLFile


templates/php/src/Client.php.twig
```diff
@@ -14,6 +14,8 @@ class Client
     const METHOD_CONNECT = 'CONNECT';
     const METHOD_TRACE = 'TRACE';
 
+    const CHUNK_SIZE=5000000;
```
TorstenDittmann, line 17: Can we have delimitier between 3 char or a comment.
```suggestion
    const CHUNK_SIZE=5_000_000;
```

templates/php/src/Services/Service.php.twig
```diff
@@ -55,6 +55,39 @@ class {{ service.name | caseUcfirst }} extends Service
...
+            $fileId = null;
+            $mimeType = mime_content_type($file);
+            $counter = 0;
+            while (!feof($handle)) {
+                $params['file'] = new \CURLFile('data://' . $mimeType . ';base64,' . base64_encode(@fread($handle, Client::CHUNK_SIZE)), $mimeType, pathinfo($file, PATHINFO_FILENAME));
+                $contentRanges = 'bytes ' . ($counter * Client::CHUNK_SIZE) . '-' . min(((($counter * Client::CHUNK_SIZE) + Client::CHUNK_SIZE) - 1), $size) . '/' . $size;
+                $response = $this->client->call(Client::METHOD_POST, $path, [
+                    'content-type' => 'multipart/form-data',
+                    'content-range' => $contentRanges,
+                    'x-appwrite-file-id' => $fileId,
```
TorstenDittmann, line 82: we can remove the `appwrite` from the template.

templates/php/src/Services/Service.php.twig
```diff
@@ -55,6 +55,39 @@ class {{ service.name | caseUcfirst }} extends Service
...
+                '{{ key }}' => '{{ header }}',
+{% endfor %}
+                ], $params);
+        } else {
+            $handle = @fopen($file, "rb");
+            $fileId = null;
+            $mimeType = mime_content_type($file);
+            $counter = 0;
+            while (!feof($handle)) {
+                $params['file'] = new \CURLFile('data://' . $mimeType . ';base64,' . base64_encode(@fread($handle, Client::CHUNK_SIZE)), $mimeType, pathinfo($file, PATHINFO_FILENAME));
```
TorstenDittmann, line 77: Do we really need to encode it to base64? 🤔

templates/php/src/Services/Service.php.twig
lohanidamodar: This will not even be required once custom ID support is merged.

templates/php/src/Services/Service.php.twig
lohanidamodar: @TorstenDittmann I am unable to create `CURLFile` with a chunk, it directly creates from File, loading whole file, only way to load partial content is either to use this basc64 encoded string or create a chunk file for each chunk.

PHP 8.1 might solve this problem, there seems to have introduced new function `CURLFileFromString` or something

templates/php/src/Client.php.twig
lohanidamodar: @TorstenDittmann seems to cause problem in php 7
image

templates/php/src/Services/Service.php.twig
```diff
@@ -55,6 +55,44 @@ class {{ service.name | caseUcfirst }} extends Service
...
+                $params['{{ parameter.name }}'] = new \CURLFile('data://' . $mimeType . ';base64,' . base64_encode(@fread($handle, Client::CHUNK_SIZE)), $mimeType, basename(${{ parameter.name | caseCamel }}));
+                $contentRanges = 'bytes ' . ($counter * Client::CHUNK_SIZE) . '-' . min(((($counter * Client::CHUNK_SIZE) + Client::CHUNK_SIZE) - 1), $size) . '/' . $size;
+                $response = $this->client->call(Client::METHOD_POST, $path, [
+                    'content-type' => 'multipart/form-data',
+                    'content-range' => $contentRanges,
+                    'x-appwrite-id' => $id,
+                ], $params);
+                $counter++;
+                $id = $response['$id'];
+                $params['{{ method.parameters.all[1].name }}'] = $id; {# Asuming that with custom ID support, second parameter (fileID and executionID both are second param), but we don't have custom ID for tags yet so will fail for createTag #}
```
lohanidamodar, line 88: @eldadfux Here, while making a chunked upload request, for subsequent request, we need the update the ID parameter (if we don't want to use extra header `x-appwrite-id`), but how will we know which parameter is the actual ID parameter?

Also, createTag, doesn't yet support custom ID so I think, using the header approach is best at the moment.

src/SDK/Language/PHP.php
```diff
@@ -208,7 +208,7 @@ public function getTypeName($type)
                 $type = 'array';
                 break;
             case self::TYPE_FILE:
-                $type = '\CurlFile';
+                $type = 'string';
```
TorstenDittmann, line 211: Reminder for us that this will be a breaking change 👍🏻

templates/php/src/Services/Service.php.twig
```diff
@@ -55,6 +55,44 @@ class {{ service.name | caseUcfirst }} extends Service
...
+            $headers = ['content-type' => 'multipart/form-data'];
+            while (!feof($handle)) {
+                $params['{{ parameter.name }}'] = new \CURLFile('data://' . $mimeType . ';base64,' . base64_encode(@fread($handle, Client::CHUNK_SIZE)), $mimeType, $postedName);
+                $headers['content-range'] = 'bytes ' . ($counter * Client::CHUNK_SIZE) . '-' . min(((($counter * Client::CHUNK_SIZE) + Client::CHUNK_SIZE) - 1), $size) . '/' . $size;
+                if(!empty($id)) {
+                    $headers['x-appwrite-id'] = $id;
+                }
+                $response = $this->client->call(Client::METHOD_POST, $path, $headers, $params);
+                $counter++;
+                $id = $response['$id'];
```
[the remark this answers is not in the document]
lohanidamodar: Ok!

tests/languages/php/test.php
```diff
@@ -19,6 +19,7 @@
 $general = new General($client);
 
 $client->addHeader('Origin', 'http://localhost');
+$client->setEndpoint("http://chunked.appwrite.org/v1"); //TODO remember to remove this
```
Meldiron, line 22: Let's not forget

Meldiron (commented):
Looks good to me. Spotted one todo, can that be removed now?

tests/languages/php/test.php
lohanidamodar: Not yet, we don't have the mock deployed to Appwrite.io yet

Meldiron: Makes sense.

Meldiron (approved):
Looks good to me 😎

templates/php/src/Services/Service.php.twig
```diff
@@ -55,6 +55,49 @@ class {{ service.name | caseUcfirst }} extends Service
...
+                if(!empty($id)) {
+                    $headers['x-appwrite-id'] = $id;
+                }
+                $response = $this->client->call(Client::METHOD_POST, $path, $headers, $params);
+                $counter++;
+                if(empty($id)) {
+                    $id = $response['$id'];
+                }
+                if($onProgress !== null) {
+                    $onProgress(min($offset, $size) / $size * 100);
```
Meldiron, line 92: Does this work? Where is `$offset` defined?

Meldiron, line 92: Are we sure we want to only return percentage? Let's say someone wanted to display "12 chunks left", he cannot do that.. Because he never have access to Client::CHUNK_SIZE, so even if he wanted, he can't. Should we return data object instead? For example:

```
{
  "progress": 58.777192290628086,
  "chunksTotal": 4,
  "chunksUploaded": 2,
  "sizeTotal": 17839845,
  "sizeUploaded": 10485760
}
```

Meldiron: If we decide to use a more informative object, then we might want to execute onProgress also if the file is <5MB, so the frontend has this information.

lohanidamodar: Ah good catch.

lohanidamodar: This can be a good Idea. However, for less than 5MB not sure how we can handle progress for each framework, haven't been able to figure anything similar for Dart/Flutter

PR #243, opened 2021-08-19
Feat android/kotlin realtime

abnegate:


templates/kotlin/src/main/kotlin/io/appwrite/Client.kt.twig
```diff
@@ -127,14 +128,31 @@ class Client @JvmOverloads constructor(
...
+    * Set endpoint and realtime endpoint.
+    *
+    * @param endpoint
+    *
+    * @return this
+    */
     fun setEndpoint(endPoint: String): Client {
         this.endPoint = endPoint
+
+        if (this.endPointRealtime == null && endPoint.startsWith("http")) {
```
TorstenDittmann, line 140: What if the endpoint starts with `https`? 🙂

tests/languages/android/ServiceTest.kt
```diff
@@ -21,28 +30,46 @@ import java.nio.file.Paths
...
         val general = General(client)
-        client.addHeader("Origin", "http://localhost")
-        client.setSelfSigned(true)
+        val realtime = Realtime(client)
+        var realtimeResponse = "Realtime failed!"
+
+        realtime.subscribe("tests") {
+            realtimeResponse = it
+                .toJson()
+                .fromJson(Map::class.java)["response"]!! as String
```
TorstenDittmann, line 67: Is it possible to skip the `toJson` and `fromJson` here and make it part of the internal callback?

TorstenDittmann (commented):
Can't really say much about the kotlin part - except that you've done a great job as it looks like 😄

templates/android/library/src/main/java/io/appwrite/Client.kt.twig
```diff
@@ -151,14 +152,31 @@ class Client @JvmOverloads constructor(
...
+     * Set endpoint and realtime endpoint.
      * 
      * @param endpoint
      *
      * @return this     
      */
     fun setEndpoint(endPoint: String): Client {
         this.endPoint = endPoint
+
+        if (this.endPointRealtime == null && endPoint.startsWith("http")) {
```
christyjacob4, line 164: There's also a case where the endpoint starts with `https` where we need to use `wss`

abnegate: In that case `startsWith("http")` is true and this block only replaces `http` with `ws`, so `https` will become `wss`

templates/android/library/src/main/java/io/appwrite/services/Realtime.kt.twig
```diff
@@ -0,0 +1,165 @@
...
+        socket = RealWebSocket(
+            taskRunner = TaskRunner.INSTANCE,
+            originalRequest = request,
+            listener = {{ spec.title | caseUcfirst }}WebSocketListener(),
+            random = Random(),
+            pingIntervalMillis = client.http.pingIntervalMillis.toLong(),
+            extensions = null,
+            minimumDeflateSize = client.http.minWebSocketMessageToCompress
+        )
+
```
christyjacob4, line 63: We could throw an error here if the socket connection was not created

christyjacob4: Or does the `RealWebSocket()` constructor already throw an exception

templates/android/library/src/main/java/io/appwrite/services/Realtime.kt.twig
```diff
@@ -0,0 +1,165 @@
...
+    fun doOnError(callback: (RealtimeError) -> Unit) {
+        errorCallbacks.add(callback)
+    }
+
+    private inner class {{ spec.title | caseUcfirst }}WebSocketListener : WebSocketListener() {
+
+        override fun onOpen(webSocket: WebSocket, response: Response) {
+            super.onOpen(webSocket, response)
+            print("WebSocket connected.")
```
christyjacob4, line 109: Do we continue to leave these log messages here ? If yes, I think it's better to replace print with `Log.i()`

templates/kotlin/src/main/kotlin/io/appwrite/Client.kt.twig
TorstenDittmann: nvm, will work with https as well :-D

templates/android/library/src/main/java/io/appwrite/services/Realtime.kt.twig
```diff
@@ -0,0 +1,165 @@
...
+            t.printStackTrace()
+        }
+
+        private inline fun <reified T> parse(text: String): T? {
+            return try {
+                Gson().fromJson(text, T::class.java)
+            } catch (ex: JsonSyntaxException) {
+                ex.printStackTrace()
+                null
+            } catch (ex: JsonParseException) {
```
christyjacob4, line 159: Are these the only two possible exceptions that we want to catch ? Shouldn't we just catch the general `Exception` and print stack trace and return `null` ?

christyjacob4 (changes requested):
Added a few comments. Rest looks great :rocket:  Awesome work @abnegate

templates/android/library/src/main/java/io/appwrite/Client.kt.twig
christyjacob4: Ah right! Perfect :+1:

templates/android/library/src/main/java/io/appwrite/services/Realtime.kt.twig
abnegate: The constructor throws if the request method is not `GET`; anything thrown after that will be routed to `onFailure` of the `AppwriteWebsocketListener`. I've changed to a non-null assertion on the `socket` for the `connect` call so any case where instantiation fails but execution continues will throw as well

christyjacob4: Okay sounds good :+1:

christyjacob4 (approved):
Looks good @abnegate  :+1:

templates/android/library/src/main/java/io/appwrite/models/RealtimeEvent.kt.twig
```diff
@@ -0,0 +1,7 @@
+package {{ sdk.namespace | caseDot }}.models
+
+data class RealtimeEvent(
```
lohanidamodar, line 3: I don't think we are using this anywhere?

lohanidamodar (approved):
Looks Good to me now

templates/android/library/src/main/java/io/appwrite/services/Realtime.kt.twig
```diff
@@ -0,0 +1,157 @@
...
+        launch {
+            subCallDepth++
+            delay(DEBOUNCE_MILLIS)
+            if (subCallDepth == 1) {
+                createSocket()
+            }
+            subCallDepth--
+        }
+
+        return RealtimeSubscription { unsubscribe(*channels) }
```
TorstenDittmann, line 92: Do we return the unsubscribe method right away or an object containing the method?

abnegate: This returns a `RealtimeSubscription` object that has a `close` function, there's a couple of reasons I did it this way:
- It implements [`java.io.Closeable`](https://docs.oracle.com/javase/8/docs/api/java/io/Closeable.html), which is the Kotlin/Android way to represent closable resources (and allows usage with try-with-resoruces or .use() constructs)
- We could add other info to the subscription model later without it being a breaking SDK change

templates/android/library/src/main/java/io/appwrite/Client.kt.twig
```diff
@@ -151,14 +152,31 @@ class Client @JvmOverloads constructor(
...
+            this.endPointRealtime = endPoint.replace("http", "ws")
```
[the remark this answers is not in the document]
abnegate: @TorstenDittmann This would have failed - I've updated to replace only the first occurrence

TorstenDittmann (approved):

PR #249, opened 2021-09-04
fix(web): exception code

TorstenDittmann:


templates/web/src/sdk.ts.twig
```diff
@@ -319,12 +319,15 @@ class {{ spec.title | caseUcfirst }} {
...
-                window.console.warn('Appwrite is using localStorage for session management. Increase your security by adding a custom domain as your API endpoint.');
+                window.console.warn('{{spec.title | caseUcfirst}} is using localStorage for session management. Increase your security by adding a custom domain as your API endpoint.');
                 window.localStorage.setItem('cookieFallback', cookieFallback);
             }
 
             return data;
         } catch (e) {
+            if (e instanceof {{spec.title | caseUcfirst}}Exception) {
+                throw e;
+            }
```
eldadfux, line 330: What exactly caused the missing error code?

TorstenDittmann: It was a nested exception caught and then thrown by the parent try/catch.

lohanidamodar (approved):

Issue #244, opened 2021-08-24
Email Verification Returns Empty if False


fidst:
https://github.com/appwrite/sdk-for-php/blob/981134a47e3bbcb1d1e17b1b36a22254a14530d4/src/Appwrite/Services/Users.php#L338

https://discord.com/channels/564160730845151244/564161373148414012/879169819814019112

When attempting to use the PHP server sdk, and sending:
 $result = $users->updateVerification([USERID], [FALSE]);

It returns "missing required parameter for emailverification"

eldadfux:
@TorstenDittmann can we close this issue?

PR #256, opened 2021-09-07
fix(php): null/false check

TorstenDittmann:
- replaces `empty()` with `isset()` to support passing `false`

#244


lohanidamodar (approved):

lohanidamodar (approved):
Good catch!

PR #262, opened 2021-09-24
Fix-realtime-and-doc-comments

lohanidamodar:
## Changelog
- Fix realtime close and reconnect working only 1 out of two times due to future returning too early
- Add dart doc comments to newly added response models

### DIFF
--- a/templates/flutter/lib/package.dart.twig
+++ b/templates/flutter/lib/package.dart.twig
@@ -8,7 +8,6 @@ import 'src/redirect_stub.dart'
     if (dart.library.html) 'src/redirect_browser.dart';
 import 'src/enums.dart';
 import 'src/client.dart';
-import 'src/response.dart';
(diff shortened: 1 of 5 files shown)


templates/flutter/lib/package.dart.twig
```diff
@@ -8,7 +8,6 @@ import 'src/redirect_stub.dart'
     if (dart.library.html) 'src/redirect_browser.dart';
 import 'src/enums.dart';
 import 'src/client.dart';
-import 'src/response.dart';
```
TorstenDittmann, line 11: Is this intended?

TorstenDittmann (approved):

templates/flutter/lib/package.dart.twig
lohanidamodar: Yes!

PR #270, opened 2021-10-01
Fix CLI encoding of JSON objects

stnguyen90:
When setting a JSON object string at the CLI, it is considered a string
and passed to the API as a string, even though it's expected as an
object. This change ensures that JSON object strings are decoded into
PHP arrays so that they are properly encoded when sent to the Appwrite
API.


stnguyen90:
Hmm I believe the CI build is failing due to the recent CA certificate expiring, and not due to my code. Should I look into fixing that as part of this PR?

stnguyen90:
> Hmm I believe the CI build is failing due to the recent CA certificate expiring, and not due to my code. Should I look into fixing that as part of this PR?

I fixed the problem by updating the image to run the php sdk test for php version 7 to php:7.4-cli-alpine.

christyjacob4:
@stnguyen90  Can you share some screenshots before and after the fix ?

stnguyen90:
Here's a screenshot of the error before the fix:

Update User Preferences - Error

Here's a screenshot of the successful command after the fix:

Update User Preferences - Successful

Here's a screenshot of the updated user preferences:

Updated User Preferences

Let me know if there's anything else you need.

PR #283, opened 2021-10-13
Fix SSL Verification in Python SDK

stnguyen90:
All the SDKs verify SSL certificates by default except for the Python SDK due to a but in the logic for SSL verification. This change fixes the logic to ensure SSL verification is done by default.

Closes: https://github.com/appwrite/sdk-for-python/issues/40

Before the update, SSL verification isn't done and there is a warning that SSL verification isn't done. The following screenshot shows the warning as well as the request continuing to https://dne.appwrite.io/v1 (which results in HTTP Status 400 as expected).

Python SDK SSL Warning

After the update, SSL verification is done and the request is not sent due to the SSL handshake failure:

Python SDK Error

### DIFF
--- a/templates/python/package/client.py.twig
+++ b/templates/python/package/client.py.twig
@@ -74,7 +74,7 @@ class Client:
                 json=json,
                 files=files,
                 headers=headers,
-                verify=self._self_signed,
+                verify=(not self._self_signed),
             )
 
             response.raise_for_status()


TorstenDittmann:
> The Travis CI build error looks like the same error I experienced over in #270. Should I apply the same fix in this PR?

We fixed the CI on `master`, please get in sync with it and the tests should run again 🙏🏻

stnguyen90:
> > The Travis CI build error looks like the same error I experienced over in #270. Should I apply the same fix in this PR?
> 
> We fixed the CI on `master`, please get in sync with it and the tests should run again 🙏🏻

I rebased and the build is now passing.

PR #302, opened 2021-11-02
Large File Chunked Upload support on Flutter

lohanidamodar:
- Chunked upload support on Flutter and Dart
- Limitation: Flutter Web and Dart JS platforms will not work with files above 5 MB


TorstenDittmann (approved):

templates/dart/lib/src/chunked_upload_io.dart.twig
```diff
@@ -0,0 +1,49 @@
...
+  }
+  io.File iofile = io.File(file.path!);
+  final size = await iofile.length();
+
+  late Response res;
+  if (size <= Client.CHUNK_SIZE) {
+    params['file'] = await http.MultipartFile.fromPath(paramName, file.path!, filename: file.fileName);
+    res = await client.call(HttpMethod.post,
+        path: path, params: params, headers: headers);
+  } else {
```
wess, line 29: Why "else" statement? Perhaps just return "await client.call..." exit quicker to shorten cycles.

templates/dart/lib/src/chunked_upload_io.dart.twig
```diff
@@ -0,0 +1,49 @@
...
+    throw AppwriteException("File path must be provided for dart:io");
+  }
+  io.File iofile = io.File(file.path!);
+  final size = await iofile.length();
+
+  late Response res;
+  if (size <= Client.CHUNK_SIZE) {
+    params['file'] = await http.MultipartFile.fromPath(paramName, file.path!, filename: file.fileName);
+    res = await client.call(HttpMethod.post,
+        path: path, params: params, headers: headers);
```
wess, line 28: In service.dart.twig (line 70 - 76) you place each argument on a different line. I like this makes readability super clear. Here, perhaps we could follow the same pattern?

wess (commented):
Overall looks really good, just a couple of things, mainly inconsistency in patterns. Dart's VM/AST is weird, it transverses ELSE to additional if evaluations with look a-head, so if we can return in the if without fall through to an else statement, it might save us a few cycles

templates/dart/lib/src/chunked_upload_io.dart.twig
lohanidamodar: Hey, good catch. I've made the changes, let me know what you think.

PR #306, opened 2021-11-02
feat(deno): response models

TorstenDittmann:
Generated `models.d.ts` and the account service for example => https://gist.github.com/TorstenDittmann/2ed14650762f9e621138d4ad21038eb5


lohanidamodar (commented):
Looks good, except for the fact about nested generics that you pointed out.

christyjacob4 (approved):

PR #321, opened 2021-11-30
refactor: Add linting rules to flutter sdk, and change code to comply with the rules

KhaleelSH:
Also, it updates packages

### DIFF
--- a/src/SDK/Language/Dart.php
+++ b/src/SDK/Language/Dart.php
@@ -101,7 +101,8 @@ class Dart extends Language {
             "set",
             "yield",
             "required",
-            "default"
+            "extension",
+            "late"
         ];
     }
 
--- a/templates/flutter/lib/src/client_browser.dart.twig
+++ b/templates/flutter/lib/src/client_browser.dart.twig
@@ -17,10 +17,12 @@ ClientBase createClient({
 class ClientBrowser extends ClientBase with ClientMixin {
   String _endPoint;
   Map<String, String>? _headers;
+  @override
   late Map<String, String> config;
   late BrowserClient _httpClient;
   String? _endPointRealtime;
 
+  @override
   String? get endPointRealtime => _endPointRealtime;
 
   ClientBrowser({
@@ -46,12 +48,14 @@ class ClientBrowser extends ClientBase with ClientMixin {
     init();
   }
 
+  @override
   String get endPoint => _endPoint;
 
 {% for header in spec.global.headers %}
 {% if header.description %}
      /// {{header.description}}
 {% endif %}
+    @override
     ClientBrowser set{{header.key | caseUcfirst}}(value) {
         config['{{ header.key | caseCamel }}'] = value;
         addHeader('{{header.name}}', value);
@@ -59,10 +63,12 @@ class ClientBrowser extends ClientBase with ClientMixin {
     }
 {% endfor %}
 
+  @override
   ClientBrowser setSelfSigned({bool status = true}) {
     return this;
   }
 
+  @override
   ClientBrowser setEndpoint(String endPoint) {
     this._endPoint = endPoint;
     _endPointRealtime = endPoint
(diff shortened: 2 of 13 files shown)


src/SDK/Language/Dart.php
```diff
@@ -101,7 +101,8 @@ public function getKeywords()
             "set",
             "yield",
             "required",
-            "default"
+            "extension",
```
lohanidamodar, line 104: `default` should also be added here, it's still a keyword?

lohanidamodar (changes requested):
@KhaleelSH great work, just few comments and suggestions
1. Tests are failing, may be because of the latest version of device_info_plugin, I was having issues too on linux, if so revert back to old version
2. I would love if you could make same changes to the Dart SDK as well (as it's somewhat exact copy implementation of this SDK)
Feel free to ping us on discord if any issues.

src/SDK/Language/Dart.php
KhaleelSH: @lohanidamodar It's already added at line `91`, I removed the duplicated one.

KhaleelSH:
@lohanidamodar Thanks. Regarding point 1, I added some exception handling around the `device_info_plugin`, So I'm pretty sure it's not the updated plugin that causes it to fail.
I see the master fails for the same reason:
```
1) Tests\SDKTest::testHTTPSuccess
Failed asserting that 0 is equal to 16 or is greater than 16.
/home/travis/build/appwrite/sdk-generator/tests/SDKTest.php:414
```

Regarding point 2, I'll do it in a separate PR.

lohanidamodar:
> @lohanidamodar Thanks. Regarding point 1, I added some exception handling around the `device_info_plugin`, So I'm pretty sure it's not the updated plugin that causes it to fail. I see the master fails for the same reason:
> 
> ```
> 1) Tests\SDKTest::testHTTPSuccess
> Failed asserting that 0 is equal to 16 or is greater than 16.
> /home/travis/build/appwrite/sdk-generator/tests/SDKTest.php:414
> ```
> 
> Regarding point 2, I'll do it in a separate PR.

Seems like playwright issues again. @TorstenDittmann would love you to look here

lohanidamodar:
@KhaleelSH please sync with master to pass the test.

lohanidamodar:
@KhaleelSH Thanks for the update, We have merged slight update, the updated packages and the thanks to your suggestion to adding error handling against device info plugin. Please sync and resolve conflict and this should be good to go. And would love similar PR in Dart, want to keep both on same standards 🙏🏻

lohanidamodar (approved):

templates/flutter/lib/src/client_browser.dart.twig
```diff
@@ -71,17 +77,20 @@ class ClientBrowser extends ClientBase with ClientMixin {
     return this;
   }
 
+  @override
   ClientBrowser setEndPointRealtime(String endPoint) {
     _endPointRealtime = endPoint;
     return this;
   }
```
wess, line 84: on line 68 you set `_endPointRealtime` when `Endpoint` is set, should we do vice-versa here?

wess (commented):
Everything looks good, only one suggestion from me.

templates/flutter/lib/src/client_browser.dart.twig
KhaleelSH: @wess I didn't change that line. All I changed here is adding the `@override` annotation.

lohanidamodar: @wess we don't want to do that, not doing that in other SDKs as well (https://github.com/aw-tests/sdk-for-web/blob/c71bb37fcf33094752885397fb7fe3427696a1af/src/sdk.ts#L822) Just because, realtime can have separate endpoint than the main endpoint, right now we are using traefik to make realtime available via same endpoint /realtime but that might change, and one can directly access via different port as well.

lohanidamodar (approved):

wess (approved):

PR #326, opened 2021-12-14
fix(web): typed arguments and service classes

TorstenDittmann:
This PR add's the following:
 - properly typed arguments
 - filters in PHP to reduce template clutter
 - split up the Web SDK in Services to allow tree-shaking & co

Output => https://github.com/appwrite/sdk-for-web/tree/feat-service-refactor


lohanidamodar:
@TorstenDittmann will this be ready for 0.13? Otherwise we have some build issues with Console SDK? Might have similar issue in Server and client, haven't tested building yet?

TorstenDittmann:
> 

This only affects server side of Web SDK and so on the console. So this would only affect the Console SDK.

Let me bring the PR up2date during this week so we have no more manual fixing on that 👍🏻

abnegate (approved):
Looks good 👌

PR #328, opened 2021-12-15
Removes RegEx from path building as it's not needed.

wess:
- Removes regex from path string replace
- Updates .gitignore for local environment


lohanidamodar:
@wess Tests are failing, please check 🙏🏻

lohanidamodar (approved):
Looks good!

PR #340, opened 2022-01-18
Fix Kotlin exceptions for Java 16+

abnegate:
In Java 16+, GSON reflection fails because it can not access private properties inside a module it does not own. To work around this we can use an explicit `TypeToken` which will avoid reflection.

Closes https://github.com/appwrite/sdk-for-kotlin/issues/16


christyjacob4 (approved):
Looks good, @abnegate

christyjacob4:
Reviewed the generated SDK. We're good to go 🚀  🤩

PR #358, opened 2022-02-03
Feat android large file

abnegate:


templates/android/library/src/main/java/io/appwrite/Client.kt.twig
```diff
@@ -297,6 +300,83 @@ class Client @JvmOverloads constructor(
...
+    /**
+     * Upload a file in chunks
+     *
+     * @param path
+     * @param headers
+     * @param params
+     *
+     * @return [T]
+     */
+    @Throws(AppwriteException::class)
```
lohanidamodar, line 312: Let's not hardcode Appwrite here

lohanidamodar (changes requested):
Tiny change, rest looks good.

templates/kotlin/src/main/kotlin/io/appwrite/Client.kt.twig
```diff
@@ -256,6 +259,83 @@ class Client @JvmOverloads constructor(
...
+            result = call(
+                "POST",
+                path,
+                headers,
+                params,
+                Map::class.java
+            )
+
+            offset += CHUNK_SIZE
+            headers["x-appwrite-id"] = result!!["\$id"].toString()
```
lohanidamodar, line 332: let's use this from spec as well

templates/kotlin/src/main/kotlin/io/appwrite/Client.kt.twig
```diff
@@ -256,6 +259,83 @@ class Client @JvmOverloads constructor(
...
+    /**
+     * Upload a file in chunks
+     *
+     * @param path
+     * @param headers
+     * @param params
+     *
+     * @return [T]
+     */
+    @Throws(AppwriteException::class)
```
lohanidamodar, line 271: Let's remove the hardcoded value here too

templates/android/library/src/main/java/io/appwrite/Client.kt.twig
```diff
@@ -297,6 +300,83 @@ class Client @JvmOverloads constructor(
...
+            result = call(
+                "POST",
+                path,
+                headers,
+                params,
+                Map::class.java
+            )
+
+            offset += CHUNK_SIZE
+            headers["x-appwrite-id"] = result!!["\$id"].toString()
```
lohanidamodar, line 373: let's remove the hardcoded value here as well

lohanidamodar (changes requested):
some more small issues.

templates/android/library/src/main/java/io/appwrite/models/RealtimeModels.kt.twig
```diff
@@ -1,5 +1,6 @@
 package {{ sdk.namespace | caseDot }}.models
 
+import kotlin.collections.Collection
```
christyjacob4, line 3: Is this required?

christyjacob4 (approved):
Looks good 👍

templates/android/library/src/main/java/io/appwrite/models/RealtimeModels.kt.twig
abnegate: Yep it is now - not sure how it wasn't before, but now without this there is a compilation error stating `Collection` does not require type arguments when used [here](https://github.com/appwrite/sdk-generator/blob/9fdb7bf3c4615f6132da47d4beb7f77dbcb8c8e6/templates/android/library/src/main/java/io/appwrite/models/RealtimeModels.kt.twig#L13), as it was referring to `io.appwrite.models.Collection` instead

PR #360, opened 2022-02-08
Feat python large file

abnegate:


templates/python/package/client.py.twig
```diff
@@ -95,6 +98,54 @@ class Client:
...
+                'post',
+                path,
+                headers,
+                params,
+            )
+            
+            offset = offset + self._chunk_size
+            
+            if "$id" in result: 
+                headers["x-appwrite-id"] = result["$id"]
```
christyjacob4, line 142: Can we use the value from the spec instead of hard coding it ?

christyjacob4 (approved):
Looks great 👍 
Just some minor suggestions

PR #365, opened 2022-02-13
Resumable upload - Dart

lohanidamodar:
- Adds resumable upload support to chunked files
	- When fileId is provided during createFile, checks if the file already exists
	- if the file already exists, sets the offset based on the chunks already uploaded
	- resumes upload from the remaining chunk


templates/dart/lib/src/chunked_upload_io.dart.twig
```diff
@@ -32,14 +34,28 @@ Future<Response> chunkedUpload({
       headers: headers,
     );
   }
-  // read chunk and upload each chunk
+
   var offset = 0;
+  if(idParamName.isNotEmpty && params[idParamName] != 'unique()') {
```
Meldiron, line 39: What if they provide `idParamName`, and also provide `id=unique()`? Should we throw error that this combination is not allowed?

templates/dart/lib/src/chunked_upload_io.dart.twig
```diff
@@ -32,14 +34,28 @@ Future<Response> chunkedUpload({
...
+    //make a request to check if a file already exists
+    try {
+      res = await client.call(
+        HttpMethod.get,
+        path: path + '/' + params[idParamName],
+        headers: headers,
+      );
+      final int chunksUploaded = res.data['chunksUploaded'] as int;
+      offset = min(size, chunksUploaded * Client.CHUNK_SIZE);
```
Meldiron, line 48: What happens if file is 100% uploaded and they try to re-upload something? We should probably instantly return the file.

templates/dart/lib/src/chunked_upload_io.dart.twig
```diff
@@ -48,7 +64,15 @@ Future<Response> chunkedUpload({
     if(offset < size) {
       headers['x-appwrite-id'] = res.data['\$id'];
     }
-    onProgress?.call(min(offset,size)/size * 100);
+    final progress = UploadProgress(
+      $id: res.data['\$id'] ?? '',
```
Meldiron, line 68: Why the `?? ''` part? We should have ID always, right?

templates/dart/lib/src/chunked_upload_io.dart.twig
```diff
@@ -48,7 +64,15 @@ Future<Response> chunkedUpload({
...
+      progress: min(offset,size)/size * 100,
+      sizeUploaded: min(offset,size),
+      chunksTotal: res.data['chunksTotal'] ?? 0,
+      chunksUploaded: res.data['chunksUploaded'] ?? 0,
```
Meldiron, line 72: Can it occur we dont get these?

templates/deno/src/models.d.ts.twig
```diff
@@ -1,5 +1,5 @@
 {% macro sub_schema(property, definition, spec) %}
-{% spaceless %}
+{% apply spaceless %}
```
Meldiron, line 2: What does these change do? I should probably know so I can work on other sdks 😅

templates/dart/lib/src/chunked_upload_io.dart.twig
lohanidamodar: Nope, idParamName is autogenerated from our spec, not provided by user.

templates/dart/lib/src/chunked_upload_io.dart.twig
lohanidamodar: yes, it returns the file. Here I'm not sure what we should do, throw duplicate error like it happens for files < 5MB or just return the file. For now it returns the file.

templates/dart/lib/src/chunked_upload_io.dart.twig
lohanidamodar: Because we don't have that for mock endpoint and tests fail.

templates/dart/lib/src/chunked_upload_io.dart.twig
lohanidamodar: In test/mock endpoint

templates/deno/src/models.d.ts.twig
lohanidamodar: the spaceless removes all the white space, previously it was `{% spaceless %}` but I updated the twig to latest version and its now `{% apply spaceless %}`.

Meldiron (approved):
LGTM

wess (approved):

PR #372, opened 2022-02-17
Improve Flutter and Dart SDK using official lint rules

lohanidamodar:


templates/dart/lib/services/service.dart.twig
```diff
@@ -38,9 +38,9 @@ class {{ service.name | caseUcfirst }} extends Service {
         };
 
 {% if method.type == 'location' %}
-        params.keys.forEach((key) {if (params[key] is int || params[key] is double) {
+        for(var key in params.keys) {if (params[key] is int || params[key] is double) {
```
wess, line 41: Why a for loop when you could use filter/map/reduce? (Map is faster and resource efficient).

templates/dart/lib/src/client_mixin.dart.twig
```diff
@@ -40,15 +40,15 @@ class ClientMixin {
     } else if (method == HttpMethod.get) {
       final encoded = <String, dynamic>{};
       if (params.isNotEmpty) {
-        params.keys.forEach((key) {
+        for(var key in params.keys) {
```
wess, line 43: Why a for loop when you could use a map? (Map is faster and resource efficient).

templates/flutter/lib/services/service.dart.twig
```diff
@@ -61,9 +61,11 @@ class {{ service.name | caseUcfirst }} extends Service {
       return client.webAuth(url);
 
 {% elseif method.type == 'location' %}
-        params.keys.forEach((key) {if (params[key] is int || params[key] is double) {
-          params[key] = params[key].toString();
-        }});
+        for (var key in params.keys) {
```
wess, line 64: Suggest using a map/filter here

templates/flutter/lib/src/realtime_mixin.dart.twig
```diff
@@ -65,13 +66,13 @@ mixin RealtimeMixin {
             break;
           case 'event':
             final message = RealtimeMessage.fromMap(data.data);
-            message.channels.forEach((channel) {
-              if (this._channels[channel] != null) {
-                this._channels[channel]!.forEach((stream) {
+            for(var channel in message.channels) {
```
wess, line 69: since dealing with sinks, for loop looks good here.

wess (commented):
Just a couple of things to consider. This looks good and will work fine, I just wanted to drop notes on some ideas going forward.

templates/flutter/lib/services/service.dart.twig
lohanidamodar: removed, as I think this was a leftover

templates/dart/lib/src/client_mixin.dart.twig
lohanidamodar: updated

templates/dart/lib/services/service.dart.twig
lohanidamodar: removed, as I think this was a leftover

Meldiron (approved):
LGTM

PR #376, opened 2022-02-22
Update swagger-appwrite-0.12.0.json

loickal:
Fix version in Swagger Documentation.
"version":"0.11.0", -> "version":"0.12.0",


lohanidamodar:
@loickal thank you for your contribution. However this swagger specs are only used for generating sdk for manual testing, these are not the specs for our production SDKs so we are fine with these issues in the spec. And as it seems this spec though file is named 0.12, this seems to be 0.11 spec.

loickal:
> @loickal thank you for your contribution. However this swagger specs are only used for generating sdk for manual testing, these are not the specs for our production SDKs so we are fine with these issues in the spec. And as it seems this spec though file is named 0.12, this seems to be 0.11 spec.

Ahh okay, thank you then :)

PR #382, opened 2022-02-28
Fix: Tarring process

Meldiron:


christyjacob4 (approved):
Lokks good 👍  Just want confirmation this works :)

Meldiron:
Tests pass, and I was able to manual QA that also went well.

PR #385, opened 2022-02-28
Fix websocket client not respecting client self signed status

abnegate:


Meldiron (approved):
LGTM

christyjacob4 (approved):
Do we need these fixes in the other SDKs?

PR #387, opened 2022-03-01
Handle properties' additional properties as maps

abnegate:
Allows upstream JSON attributes to be handled as maps instead of strings


Meldiron (approved):
LGTM

christyjacob4 (approved):

PR #388, opened 2022-03-01
Refactor on progress

abnegate:
Use typed model for Android, Kotlin, Swift (Server + Client) and Ruby `upload` calls `onProgress` functional parameter


Meldiron (approved):
LGTM

PR #390, opened 2022-03-01
Fix: chunk upload

Meldiron:


src/SDK/Language/JS.php
```diff
@@ -130,7 +130,7 @@ public function getTypeName($type)
             case self::TYPE_ARRAY:
                 return 'string[]';
             case self::TYPE_FILE:
-                return 'File';
+                return 'string';
```
TorstenDittmann, line 133: This will break Web SDK.

PR #408, opened 2022-03-18
Add fixed cookie dependency for Android 12+

abnegate:
Add fixed cookie store dependency to resolve cookies issues on Android 12+


lohanidamodar (approved):

src/SDK/Language/Android.php
```diff
@@ -47,207 +44,193 @@ public function getFiles()
...
+            ],
+            [
+                'scope'         => 'default',
+                'destination'   => '/library/src/main/java/{{ sdk.namespace | caseSlash }}/cookies/InMemoryCookieStore.kt',
+                'template'      => '/android/library/src/main/java/io/appwrite/cookies/InMemoryCookieStore.kt.twig',
+            ],
+            [
+                'scope'         => 'default',
+                'destination'   => '/library/src/main/java/{{ sdk.namespace | caseSlash }}/cookies/SharedPreferencesCookieStore.kt',
+                'template'      => '/android/library/src/main/java/io/appwrite/cookies/SharedPreferencesCookieStore.kt.twig',
```
christyjacob4, line 182: Do you think a better naming convention is to have

io/appwrite/cookie
  - store
      - Extensions.kt
      - InMemory.kt
      - SharedPreference.kt 
  - Internal.kt 


WDYT @abnegate  ?

christyjacob4 (changes requested):
Looks good 👍 
Just a minor comment about the folder and file structure ...

src/SDK/Language/Android.php
abnegate: @christyjacob4 I think the folder structure provides a better separation of concerns but the naming makes it hard to tell what the classes are doing without reading some code.

I have updated with just the directory changes, does that work for you?

christyjacob4: Understood and your change makes sense 👍  Thank you

christyjacob4 (approved):

PR #409, opened 2022-03-18
Feat swift async await

abnegate:
- Add async/await support for client/service methods
- Defer existing methods to async versions and deprecate completion handler functions


lohanidamodar (changes requested):
@abnegate let's update the doc examples as well to use async/await

abnegate:
@lohanidamodar Update docs and integrated examples

christyjacob4 (approved):

lohanidamodar (approved):

PR #410, opened 2022-03-18
Fix Archive not being deleted when the deploy request fails

PineappleIOnic:
This PR adds a catch to the HTTP calls made in the CLI when we are deploying functions. The catch will make sure we delete the archive then will throw the error again so our handler can deal with it.


PineappleIOnic:
Fixes https://github.com/appwrite/sdk-for-cli/issues/28

christyjacob4 (approved):

PR #412, opened 2022-03-21
Feat resumable upload android

abnegate:


templates/android/library/src/main/java/io/appwrite/Client.kt.twig
```diff
@@ -330,74 +332,84 @@ class Client @JvmOverloads constructor(
...
-                id = result!!["\$id"].toString(),
-                progress = offset.coerceAtMost(size).toDouble()/size * 100,
-                sizeUploaded = offset.coerceAtMost(size),
-                chunksTotal = result!!["chunksTotal"].toString().toInt(),
-                chunksUploaded = result!!["chunksUploaded"].toString().toInt(),
-            ))
+            headers["x-appwrite-id"] = result!!["\$id"].toString()
+            onProgress?.invoke(
+                UploadProgress(
+                    id = result!!["\$id"].toString(),
```
christyjacob4, line 387: isn't using `!!` prone to null pointer exceptions ? Can we make it null safe ?

christyjacob4 (approved):

templates/android/library/src/main/java/io/appwrite/Client.kt.twig
abnegate: It is, force-unwrapping an optional - but conveys what we want here, if there is no `result`, it is an exceptional case and we should stop uploading

PR #415, opened 2022-03-21
Feat resumable upload swift

abnegate:


templates/swift/Sources/Client.swift.twig
```diff
@@ -228,7 +229,7 @@ open class Client {
         headers: [String: String] = [:],
         params: [String: Any?] = [:],
         sink: ((ByteBuffer) -> Void)? = nil,
-        convert: (([String: Any]) -> T)? = nil,
+        converter: (([String: Any]) -> T)? = nil,
```
lohanidamodar, line 232: Why this change?

lohanidamodar (commented):
Small query, rest looks good!

templates/swift/Sources/Client.swift.twig
abnegate: Just a clearer name for the parameter, it is only used internally

christyjacob4 (approved):

PR #416, opened 2022-03-21
Parallel env CI tests

abnegate:
- Make `tests/Base.php` test take a single `command` instead of an array of `envs` to force all tests to run in parallel


christyjacob4:
Wow @abnegate  This looks amazing!!! 🤩  Brilliant work!

christyjacob4 (approved):

PR #419, opened 2022-03-31
Fix: JSON object parsing

Meldiron:
Currently, prefs are passed as string into CLI, but are expected as object by server, resulting in invalid 4XX error:
CleanShot 2022-03-31 at 13 44 25

With these changes, if expected type is object, input from CLI is automatically parsed with `JSON.parse()`, resulting in successful execution:

CleanShot 2022-03-31 at 13 45 15

CleanShot 2022-03-31 at 13 45 26


Meldiron:
This PR most likely also addresses:

- Create document
- Update document
- Update function (env vars)
- Create function (env vas)

christyjacob4 (approved):

PR #423, opened 2022-04-14
Fix: Powershell install

Meldiron:
- File must end with `.exe` for it to work. Instead of making `Appwrite` file, we now make `appwrite.exe` inside `Appwrite` folder
- After installation, if we change PATH env, we also refresh it, so you can instantly use without restarting powershell. that's something like `source` on linux

## Tests

- [x] Manual QA

### DIFF
--- a/example.php
+++ b/example.php
@@ -169,7 +169,7 @@ try {
 
     $sdk
         ->setName('NAME')
-        ->setVersion('0.0.19')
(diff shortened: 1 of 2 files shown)


adityaoberai (approved):
Manually tested the script
Updates look good 👍

christyjacob4 (approved):

PR #425, opened 2022-04-17
Feat: Appwrite CLI validate endpoint on login and client --endpoint commands

lohanidamodar:


lohanidamodar:
Sample:
image

lohanidamodar:
Example login:
image

## valid
image

lohanidamodar:
@christyjacob4  I think we shouldn't depend on the mock endpoint, may be we can introduce something on the Health service to validate it's a valid Appwrite server? WDYT?

Meldiron (commented):
LGTM. A bit scared about using  existing mock endpoint.. It might change in future. Shouldn't we add new mock endpoint for CLI specifically?

templates/cli/lib/commands/generic.js.twig
```diff
@@ -68,6 +69,13 @@ const client = new Command("client")
             try {
                 let url = new URL(endpoint);
                 if (url.protocol === "http:" || url.protocol === "https:") {
+                    let apClient = new Client().setEndpoint(endpoint);
+                    let response = await apClient.call('GET', '/health/version', {}, {'x': 'a', 'y': 1, 'z': ['a']});
```
TorstenDittmann, line 73: Just a question, is `apClient` on purpose or a type and it should be `apiClient`?

TorstenDittmann: And we also don't need to send data with it 👍🏻

templates/cli/lib/sdks.js.twig
```diff
@@ -7,7 +7,24 @@ const questionGetEndpoint = [
...
     message: "Enter the endpoint of your {{ spec.title|caseUcfirst }} server",
-    default: "http://localhost/v1"
+    default: "http://localhost/v1",
+    validate: async (value) => {
+      if(!value) {
+        return 'Please provide valid endpoint';
+      }
+      let client = new Client().setEndpoint(value);
+      try {
+        let response = await client.call('GET', '/health/version', {}, {'x': 'a', 'y': 1, 'z': ['a']});
```
TorstenDittmann, line 17: We don't need to send data here right?

templates/cli/lib/commands/generic.js.twig
lohanidamodar: `ap` meaning `appwrite`

templates/cli/lib/sdks.js.twig
```diff
@@ -7,7 +7,24 @@ const questionGetEndpoint = [
...
+        if(response.version) {
+          return true;
+        } else {
+          return 'Not a valid endpoint or Appwrite server is not running as expected';
+        }
+      } catch (error) {
+        return 'Not a valid endpoint or Appwrite server is not running as expected';
+      }
+      return true;
+    }
```
christyjacob4, line 27:
```suggestion
    validate(value) {
      if (!value) {
        return "Please enter a valid endpoint.";
      }

      let client = new Client().setEndpoint(value);
      
      try {
        let response = await client.call('get', '/health/version');
        if (response.version) {
          return true;
        } else {
          throw new Error();
        }
      } catch (error) {
        return "Invalid endpoint or your Appwrite server is not running as expected.";
      }
    }
```

templates/cli/lib/commands/generic.js.twig
```diff
@@ -68,6 +69,13 @@ const client = new Command("client")
             try {
                 let url = new URL(endpoint);
                 if (url.protocol === "http:" || url.protocol === "https:") {
+                    let apClient = new Client().setEndpoint(endpoint);
```
christyjacob4, line 72:
```suggestion
                    let client = new Client().setEndpoint(endpoint);
```
Let's be consistent with the other places we use the `Client()` class

templates/cli/lib/commands/generic.js.twig
```diff
@@ -68,6 +69,13 @@ const client = new Command("client")
...
+                    let response = await apClient.call('GET', '/health/version');
```
christyjacob4, line 73:
```suggestion
                    let response = await client.call('GET', '/health/version');
```

templates/cli/lib/commands/generic.js.twig
```diff
@@ -68,6 +69,13 @@ const client = new Command("client")
...
+                    if(response.version) {
+                        globalConfig.setEndpoint(endpoint);
+                    } else {
+                        throw new Error();
```
christyjacob4, line 77:
```suggestion
                        throw new Error("Invalid endpoint or your Appwrite server is not running as expected.");
```

christyjacob4 (changes requested):
Looks awesome! 🚀  Please address the comments and validate if everything is working as expected 🙂

templates/cli/lib/sdks.js.twig
lohanidamodar: Why this? Arrow function works, Also, we need this to be async?

templates/cli/lib/commands/generic.js.twig
lohanidamodar: There is already a `client` defined in this scope.

templates/cli/lib/commands/generic.js.twig
lohanidamodar: There is already a `client` defined in this scope.

templates/cli/lib/sdks.js.twig
lohanidamodar: Updated this, removed arrow function for consistency

templates/cli/lib/sdks.js.twig
```diff
@@ -7,7 +7,24 @@ const questionGetEndpoint = [
     type: "input",
     name: "endpoint",
     message: "Enter the endpoint of your {{ spec.title|caseUcfirst }} server",
-    default: "http://localhost/v1"
+    default: "http://localhost/v1",
+    async validate(value) {
+      if(!value) {
+        return 'Please provide valid endpoint';
+      }
```
christyjacob4, line 14:
```suggestion
      if (!value) {
        return "Please enter a valid endpoint.";
      }
```
We're using `"` in all the places AFAIK.

templates/cli/lib/sdks.js.twig
```diff
@@ -7,7 +7,24 @@ const questionGetEndpoint = [
...
+      let client = new Client().setEndpoint(value);
+      try {
+        let response = await client.call('GET', '/health/version');
```
christyjacob4, line 17:
```suggestion
        let response = await client.call('get', '/health/version');
```
we're using lower case for method names

templates/cli/lib/sdks.js.twig
```diff
@@ -7,7 +7,24 @@ const questionGetEndpoint = [
...
+        if(response.version) {
+          return true;
+        } else {
+          return 'Not a valid endpoint or Appwrite server is not running as expected';
+        }
```
christyjacob4, line 22:
```suggestion
        if (response.version) {
          return true;
        } else {
          throw new Error();
        }
```

templates/cli/lib/sdks.js.twig
```diff
@@ -7,7 +7,24 @@ const questionGetEndpoint = [
...
+      } catch (error) {
+        return 'Not a valid endpoint or Appwrite server is not running as expected';
```
christyjacob4, line 24:
```suggestion
        return "Invalid endpoint or your Appwrite server is not running as expected.";
```

templates/cli/lib/sdks.js.twig
```diff
@@ -7,7 +7,24 @@ const questionGetEndpoint = [
...
+      }
+      return true;
```
christyjacob4, line 26:
```suggestion
```
This code is not reachable

templates/cli/lib/commands/generic.js.twig
christyjacob4: This will override the other `client` object.

templates/cli/lib/commands/generic.js.twig
christyjacob4: This will override the other `client` object.

templates/cli/lib/commands/generic.js.twig
```diff
@@ -68,6 +69,13 @@ const client = new Command("client")
             try {
                 let url = new URL(endpoint);
                 if (url.protocol === "http:" || url.protocol === "https:") {
+                    let apClient = new Client().setEndpoint(endpoint);
+                    let response = await apClient.call('GET', '/health/version');
+                    if(response.version) {
+                        globalConfig.setEndpoint(endpoint);
+                    } else {
+                        throw new Error("Invalid endpoint or your Appwrite server is not running as expected.");
```
christyjacob4, line 77:
```suggestion
                        throw new Error();
```
Just throw the error and add the message `Invalid endpoint or your Appwrite server is not running as expected.` 
here 
https://github.com/appwrite/sdk-generator/blob/fbd6effddaaba3f9ef6a6725b3df648436fe8212/templates/cli/lib/commands/generic.js.twig#L76

templates/cli/lib/commands/generic.js.twig
```diff
@@ -68,6 +69,13 @@ const client = new Command("client")
...
+                    }
                     globalConfig.setEndpoint(endpoint);
```
christyjacob4, line 79:
```suggestion
                    let apClient = new Client().setEndpoint(endpoint);
                    let response = await apClient.call('GET', '/health/version');
                    if(response.version) {
                        globalConfig.setEndpoint(endpoint);
                    } else {
                        throw new Error("Invalid endpoint or your Appwrite server is not running as expected.");
                    }
```

christyjacob4: Line 79 is redundant

templates/cli/lib/commands/generic.js.twig
```diff
@@ -67,13 +68,19 @@ const client = new Command("client")
...
                 let url = new URL(endpoint);
-                if (url.protocol === "http:" || url.protocol === "https:") {
-                    globalConfig.setEndpoint(endpoint);
-                } else {
+                if (url.protocol !== "http:" && url.protocol !== "https:") {
                     throw new Error();
                 }
+                
+                let apClient = new Client().setEndpoint(endpoint);
+                let response = await apClient.call('GET', '/health/version');
```
TorstenDittmann, line 76:
```suggestion
                let client = new Client().setEndpoint(endpoint);
                let response = await client.call('GET', '/health/version');
```

PR #427, opened 2022-04-20
Fix: Function deploy vars bug

Meldiron:
Currently, if you deploy function a second time (when an update is in place instead of creating), you get the following error:
CleanShot 2022-04-20 at 10 56 51

This PR addresses this problem. In the previous change, we consider all CLI JSON input as a string, because when you run CLI command like `appwrite database createDocument --data '{"age": 5}'`, it is a string. This PR makes sure the non-generated deploy command also passes JSON objects as string.

QA with UPDATE:

CleanShot 2022-04-20 at 10 58 32

QA with CREATE:

CleanShot 2022-04-20 at 10 59 13


templates/cli/lib/commands/deploy.js.twig
```diff
@@ -181,7 +181,7 @@ const deployFunction = async () => {
                     name: func.name,
                     runtime: func.runtime,
                     execute: func.execute,
-                    vars: func.vars,
+                    vars: JSON.stringify(func.vars),
```
stnguyen90, line 184: I'm not sure if this is really needed because `func.vars` comes from the config file which should already be a JSON object.

stnguyen90:
Related issue: https://github.com/appwrite/appwrite/issues/3097
Duplicate PR: https://github.com/appwrite/sdk-generator/pull/424

abnegate (approved):

TorstenDittmann (approved):

PR #428, opened 2022-04-20
Fix: Deno Chunk Upload

Meldiron:
Currently, if you upload a 7MB file, the whole 10MB is uploaded, the rest of the data being zeros:

CleanShot 2022-04-20 at 13 18 27

This PR fixes the issue and shrinks the buffer to remove zeros, making the file exact size.

### DIFF
--- a/templates/deno/src/services/service.ts.twig
+++ b/templates/deno/src/services/service.ts.twig
@@ -150,6 +150,9 @@ export class {{ service.name | caseUcfirst }} extends Service {
                         totalBuffer[(blockIndex * Client.DENO_READ_CHUNK_SIZE) + byteIndex] = buf[byteIndex];
                     }
                 }
+
+                // Shrink empty bytes
+                totalBuffer = new Uint8Array(totalBuffer.buffer, 0);
                 
                 payload['{{ parameter.name }}'] = new File([totalBuffer], basename({{ parameter.name | caseCamel | escapeKeyword }}));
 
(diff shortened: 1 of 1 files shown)


templates/deno/src/services/service.ts.twig
```diff
@@ -148,7 +149,20 @@ export class {{ service.name | caseUcfirst }} extends Service {
...
+                            lastBufferIndex = (blockIndex * Client.DENO_READ_CHUNK_SIZE) + byteIndex;
+                        }
+                    }
+                }
+
+                // Shrink empty bytes
+                if(lastBufferIndex !== -1) {
+                    const newTotalBuffer = new Uint8Array(lastBufferIndex + 1);
+                    for(let index = 0; index <= lastBufferIndex; index++) {
+                        newTotalBuffer[index] = totalBuffer[index];
```
abnegate, line 163: Could we use an `ArrayBuffer` and create the final `Uint8Array` using the `new Uint8Array(buffer, byteOffset, length)` constructor to avoid copying the bytes here?

abnegate (changes requested):

templates/deno/src/services/service.ts.twig
Meldiron: It took quite a lot of try&error to get the code working and I am honestly scared to touch it. I am making a note to myself to revisit it later, but to make sure this makes it into the next release, I would prefer to skip this improvement.

TorstenDittmann (approved):

templates/deno/src/services/service.ts.twig
TorstenDittmann: Let's keep it for now - but improve it in the future. Considering the release date 👍🏻

PR #429, opened 2022-04-20
Fix File Create Docs

gewenyu99:
Currently, the documentation for creating files passes in a file binary. Our current implementation only accepts a file path. This must be corrected in the documentation for all SDKs.

Node SDK ✅
Deno SDK ✅
PHP SDK ✅
Python SDK ✅
Ruby SDK ✅
Dart SDK ✅ (this hasn't changed)
Kotlin (Java) ✅ (this hasn't changed)
Kotlin (Kotlin) ✅ (this hasn't changed)
Swift ✅ (this hasn't changed)

Related to
https://github.com/appwrite/sdk-for-node/issues/32

### DIFF
--- a/src/SDK/Language/Node.php
+++ b/src/SDK/Language/Node.php
@@ -127,7 +127,7 @@ class Node extends JS
                     $output .= '{}';
                     break;
                 case self::TYPE_FILE:
-                    $output .= "fs.createReadStream(__dirname + '/file.png')";
+                    $output .= "__dirname + '/file.png'";
                     break;
             }
         }
@@ -146,7 +146,7 @@ class Node extends JS
                     $output .= "'{$example}'";
                     break;
                 case self::TYPE_FILE:
-                    $output .= "fs.createReadStream(__dirname + '/file.png')";
+                    $output .= "__dirname + '/file.png'";
                     break;
             }
         }
--- a/src/SDK/Language/Node.php
+++ b/src/SDK/Language/Node.php
@@ -127,7 +127,7 @@ class Node extends JS
                     $output .= '{}';
                     break;
                 case self::TYPE_FILE:
-                    $output .= "__dirname + '/file.png'";
+                    $output .= "'file.png'";
                     break;
             }
         }
@@ -146,7 +146,7 @@ class Node extends JS
                     $output .= "'{$example}'";
                     break;
                 case self::TYPE_FILE:
-                    $output .= "__dirname + '/file.png'";
+                    $output .= "'file.png'";
                     break;
             }
         }
--- a/src/SDK/Language/PHP.php
+++ b/src/SDK/Language/PHP.php
@@ -309,7 +309,7 @@ class PHP extends Language {
                     $output .= '[]';
                     break;
                 case self::TYPE_FILE:
-                    $output .= "new \CURLFile('/path/to/file.png', 'image/png', 'file.png')";
+                    $output .= "'/path/to/file.png'";
                     break;
             }
         }
@@ -330,7 +330,7 @@ class PHP extends Language {
                     $output .= "'{$example}'";
                     break;
                 case self::TYPE_FILE:
-                    $output .= "new \CURLFile('/path/to/file.png', 'image/png', 'file.png')";
+                    $output .= "'/path/to/file.png'";
                     break;
             }
         }
--- a/src/SDK/Language/PHP.php
+++ b/src/SDK/Language/PHP.php
@@ -309,7 +309,7 @@ class PHP extends Language {
                     $output .= '[]';
                     break;
                 case self::TYPE_FILE:
-                    $output .= "'/path/to/file.png'";
+                    $output .= "'file.png'";
                     break;
             }
         }
@@ -330,7 +330,7 @@ class PHP extends Language {
                     $output .= "'{$example}'";
                     break;
                 case self::TYPE_FILE:
-                    $output .= "'/path/to/file.png'";
+                    $output .= "'file.png'";
                     break;
             }
         }
--- a/src/SDK/Language/Deno.php
+++ b/src/SDK/Language/Deno.php
@@ -145,7 +145,7 @@ class Deno extends JS
                     $output .= '{}';
                     break;
                 case self::TYPE_FILE:
-                    $output .= "new File([fileBlob], 'file.png')";
+                    $output .= "'file.png'";
                     break;
             }
         }
@@ -164,7 +164,7 @@ class Deno extends JS
                     $output .= "'{$example}'";
                     break;
                 case self::TYPE_FILE:
-                    $output .= "new File([fileBlob], 'file.png')";
+                    $output .= "'file.png'";
                     break;
             }
         }
--- a/src/SDK/Language/Kotlin.php
+++ b/src/SDK/Language/Kotlin.php
@@ -197,7 +197,7 @@ class Kotlin extends Language {
         if(empty($example) && $example !== 0 && $example !== false) {
             switch ($type) {
                 case self::TYPE_FILE:
-                    $output .= 'File("./path-to-files/image.jpg")';
+                    $output .= 'File("file.png")';
                     break;
                 case self::TYPE_NUMBER:
                 case self::TYPE_INTEGER:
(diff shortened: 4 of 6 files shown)


gewenyu99:
@lohanidamodar I feel like the behavior of the Dart SDK isn't consistent with the other SDKs. It still takes an object like this:
```dart 
  Future result = storage.createFile(
    bucketId: 'default',
    fileId: 'unique()',
    file: InputFile(path: './test.file', filename: 'test.file'),
  );
```
When I update the docs, I will change the parameter description from `Binary file.` to `File path string.`
Screen Shot 2022-04-20 at 2 33 50 PM

We either have to make the description different for Dart (which I don't think is possible (?)) or we should change the current Dart SDK to also take a string.

gewenyu99:
@TorstenDittmann I'm requesting a PR review, but this PR really isn't done yet :P

lohanidamodar:
> @lohanidamodar I feel like the behavior of the Dart SDK isn't consistent with the other SDKs. It still takes an object like this:
> 
> ```dart
>   Future result = storage.createFile(
>     bucketId: 'default',
>     fileId: 'unique()',
>     file: InputFile(path: './test.file', filename: 'test.file'),
>   );
> ```
> 
> When I update the docs, I will change the parameter description from `Binary file.` to `File path string.` Screen Shot 2022-04-20 at 2 33 50 PM
> 
> We either have to make the description different for Dart (which I don't think is possible (?)) or we should change the current Dart SDK to also take a string.

We don't take path and instead use our own object, as we also accept MultiPart file, perks of Flutter web and Dart JS. So we cannot change this to accept file path. For now it's ok to leave the param description as it is, the `file` in the parameter list table should still be `Binary file` because we are not sending the Path in the HTTP request, and parameter table is for HTTP request, not SDK params. For SDK updating the examples should suffice, for dart/Flutter examples should already be updated.

src/SDK/Language/Node.php
```diff
@@ -127,7 +127,7 @@ public function getParamExample(array $param)
                     $output .= '{}';
                     break;
                 case self::TYPE_FILE:
-                    $output .= "fs.createReadStream(__dirname + '/file.png')";
+                    $output .= "__dirname + '/file.png'";
```
Meldiron, line 130: This can be as simple as Deno, IIRC `'file.png'`

src/SDK/Language/Node.php
```diff
@@ -146,11 +146,11 @@ public function getParamExample(array $param)
                     $output .= "'{$example}'";
                     break;
                 case self::TYPE_FILE:
-                    $output .= "fs.createReadStream(__dirname + '/file.png')";
+                    $output .= "__dirname + '/file.png'";
```
Meldiron, line 149: Same as above

src/SDK/Language/PHP.php
```diff
@@ -330,7 +330,7 @@ public function getParamExample(array $param)
                     $output .= "'{$example}'";
                     break;
                 case self::TYPE_FILE:
-                    $output .= "new \CURLFile('/path/to/file.png', 'image/png', 'file.png')";
+                    $output .= "'/path/to/file.png'";
```
Meldiron, line 333: Same as above

Meldiron (commented):
Left a few comments. Can we also test all of these to make sure they are correct? We could test and document all regarding relative and absolute path.

gewenyu99:
@Meldiron Sounds good. Let's standardize all the paths in examples. There's no reason to use a million different things

gewenyu99:
> Left a few comments. Can we also test all of these to make sure they are correct? We could test and document all regarding relative and absolute path.

Yep I tested these examples manually. I will do them again with the the same type of path params (ideally use the same path parameters)

I'm going to be out until May 1st, so I'll get on it when I get back.

gewenyu99:
Kotlin Test: https://gist.github.com/gewenyu99/fc660609e1bd20978077d0ff753b1741
PHP Test: https://gist.github.com/gewenyu99/f4ede144987b9893ac3c13df7535ea4e
Ruby Test: https://gist.github.com/gewenyu99/cf46bb2392a7f4d1916cfe29b9976743
Node Test: https://gist.github.com/gewenyu99/54e4ab45ed8ea0ff802937eb2ad18903
Python Test: https://gist.github.com/gewenyu99/6161adec9d831ad155126d0ce4bb6ef0
Deno Test: https://gist.github.com/gewenyu99/323ce11eb8d49df1f8dd541d1fbdb502

gewenyu99:
@lohanidamodar I believe I've covered every SDK that now uses a path instead of a buffer.

There are several that did not change.

christyjacob4 (approved):

ElectraPort:
Team, I hate to say it but your node.js documentation is still wrong. Your example on [The AppWrite documentation](https://appwrite.io/docs/server/storage?sdk=nodejs-default#storageCreateFile)  uses the object you discuss as a Dart object above:

`const promise = storage.createFile('[BUCKET_ID]', '[FILE_ID]', InputFile.fromPath('/path/to/file.png', 'file.png'));`

But there is no class InputFile in Node. Thus, using the example code results in an undefined error wherever this line exists in your code.
```
ep-oh-bu-41517a6905264eb4af4273741a904102-14Sep2022-0300.zip
/root/ep-util/unity.js:29
        const promise = storage.createFile( '63212f5575a8936b1d48', id, InputFile.fromPath( path + file, file ) );
                                                                        ^

ReferenceError: InputFile is not defined
    at /root/ep-util/unity.js:29:73
    at Array.forEach (<anonymous>)
    at /root/ep-util/unity.js:25:11
    at FSReqCallback.oncomplete (fs.js:180:23)
```

In #32 the OP says it takes a string on line 332 of storage.js. I find that in the current version it is back to taking an InputFile:
     * @param {InputFile} file
  
[Linked Code, not sure how to make it preview like OP did.](https://github.com/appwrite/sdk-for-node/blob/50c070c1e3eb4022c9f6fa587a87039922d0057a/lib/services/storage.js#L294)

There is nothing complex about my test code and in looking at the SDK it appears you define the model for InputFile and export it apparently properly but it's not working? 

```
const sdk = require( 'node-appwrite' );
const fs = require( 'fs' );

// Init SDK
const client = new sdk.Client();

const storage = new sdk.Storage( client );

client
    .setEndpoint( 'https://my url/v1' ) // Your API Endpoint
    .setProject( 'my project ID' ) // Your project ID
    .setKey( '<my API Key' );

var path = '/opt/our-path/d01/';

//passsing directoryPath and callback function
fs.readdir( path, function ( err, files ) {
    //handling error
    if ( err ) {
        return console.log('Unable to scan directory: ' + err);
    }
    //listing all files using forEach
    files.forEach( function ( file ) {
        var id = 'unique()';
        // Do whatever you want to do with the file
        console.log( file );
        const promise = storage.createFile( '63212f5575a8936b1d48', id, InputFile.fromPath( path + file, file ) );

        promise.then( function ( response ) {
            console.log( response );
        }, function ( error ) {
            console.log( error );
        });
    });
});
```
fs.readdir returns a string array by default, so file should be a string and this should work if the compiler could find InputFile's definition. 
But you should also change the documentation because it says it takes file which is type "file" which is "Binary file" which isn't true of type InputFile.
image

Any thoughts?

Thanks!

ElectraPort:
Team, 

Further to the above it occurred to me I could do named imports from the package. So, a working snippet that should be along the lines of what's in documentation:

`const { Client, InputFile, Storage }  = require( 'node-appwrite' );
const fs = require( 'fs' );

// Init SDK
const client = new Client();

const storage = new Storage( client );

client
    .setEndpoint( 'https://<Your_Project_URL>/v1' ) // Your API Endpoint
    .setProject( '<Your_Project_ID>' ) // Your project ID
    .setKey( '<Your API Key>' );

var path = '/your-path/';

//passsing directoryPath and callback function
fs.readdir( path, function ( err, files ) {
    //handling error
    if ( err ) {
        return console.log('Unable to scan directory: ' + err);
    }
    //listing all files using forEach
    files.forEach( function ( file ) {
        var id = 'unique()';
        // Do whatever you want to do with the file
        console.log( file );
        const promise = storage.createFile( '<Project_ID>', id, InputFile.fromPath( path + file, file ) );

        promise.then( function ( response ) {
            console.log( response );
        }, function ( error ) {
            console.log( error );
        });
    });
});
`

Best regards,

EP

PR #406, opened 2022-03-16
Fix init function on Windows

PineappleIOnic:
Convert the unix commands for initialising a function into windows ones if needed.

PR #430, opened 2022-04-26
Fix: Init function on Windows

Meldiron:
Currently, the Windows command is not working, because we rely too much on bash functionality. With this PR, we only relay in `git` commands - everything else is taken care of by the `fs` library in Node.

## Tests

- [x] QA on Linux (Gitpod):
CleanShot 2022-04-26 at 16 23 24@2x

- [x] QA in Windows (CMD):

With changes in sdk-for-cli:
CleanShot 2022-04-26 at 16 25 11@2x

With final changes using sdk-generator example:
image

### DIFF
--- a/templates/cli/lib/commands/init.js.twig
+++ b/templates/cli/lib/commands/init.js.twig
@@ -54,9 +54,17 @@ const initProject = async () => {
 }
 
 const initFunction = async () => {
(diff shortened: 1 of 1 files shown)


Meldiron:
Closing. I made clone of: https://github.com/appwrite/sdk-generator/pull/406

PR #436, opened 2022-04-29
Rust SDK

PineappleIOnic:
This PR has Template files and Tests for an Appwrite Rust SDK,
I have also written the foobar tests and ran them myself to confirm that this SDK works correctly, This SDK Conforms as closely as it can with the [Rust Language's Standards](https://rust-lang.github.io/rfcs/2436-style-guide.html) so it shouldn't compile with any warnings apart from unused imports which can't be avoided due to how the templates work.

Currently, the API is blocking I/O but I plan to add async support in the future behind a feature in the crate which I'm quite excited to do considering I'm not very good with async rust.

This PR also has full compatibility with Appwrite's response models making it as easy as possible for users to use the SDK. A Playground has already been written and tested and a PR is being prepared.


CEbbinghaus:
Love to see it. I was thinking of creating a Rust SDK however I have no experience with it and it's a lot of work. Thank you for all that you have done

PineappleIOnic:
Note: All Rust tests are passing in above commit, non-rust tests are failing.

ZanzyTHEbar:
How is this going? Would love to see a Rust SDK for server-side ;)

templates/rust/README.md.twig
```diff
@@ -0,0 +1,35 @@
+# {{ spec.title }} {{sdk.name}} SDK
+
+![License](https://img.shields.io/github/license/{{ sdk.gitUserName|url_encode }}/{{ sdk.gitRepoName|url_encode }}.svg?v=1)
+![Version](https://img.shields.io/badge/api%20version-{{ spec.version|url_encode }}-blue.svg?v=1)
+{% if sdk.warning %}
```
CEbbinghaus, line 5: would be nice if these if/endif blocks could be consistent with whitespace. I prefer:
```
{if}
thing..
{endif}```
without the whitespace between if & thing.

hugofpsilva:
@PineappleIOnic any news on this? Can I help out in any way?

LcsH0s:
@PineappleIOnic is this branch dead? Seems like PR was never reviewed

PineappleIOnic:
@LcsH0s Yeah unfortunately this branch is currently dead, it is very old and Rust, Appwrite and the ways we generate SDK's have changed quite drastically, so we'll need to possibly rewrite this from the ground up

LcsH0s:
@PineappleIOnic thank you for all the work you did at least
I guess this is going on the long list of OSS contributions I need to make lol

PR #437, opened 2022-05-02
Add extensions to allow propagating swift error message

abnegate:
Related https://github.com/appwrite/appwrite/issues/3163

### DIFF
--- a/templates/swift/Sources/Models/Error.swift.twig
+++ b/templates/swift/Sources/Models/Error.swift.twig
@@ -12,3 +12,19 @@ open class {{ spec.title | caseUcfirst}}Error : Swift.Error, Decodable {
         self.type = type
     }
 }
+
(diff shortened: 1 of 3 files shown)


templates/swift/Sources/Models/Error.swift.twig
```diff
@@ -12,3 +12,19 @@ open class {{ spec.title | caseUcfirst}}Error : Swift.Error, Decodable {
...
+extension {{ spec.title | caseUcfirst}}Error: CustomStringConvertible {
+    public var description: String {
+        get {
+            return self.message
+        }
+    }
+}
+
+extension {{ spec.title | caseUcfirst}}Error: LocalizedError {
+    public var errorDescription: String? {
```
Meldiron, line 25: One is `String`, whiel other `String?`. Is that on purpose? If so, you can ignore my comment

TorstenDittmann (approved):

PR #439, opened 2022-05-05
Add download tests

abnegate:
- Add file download tests


Meldiron (approved):

tests/Base.php
```diff
@@ -35,7 +35,7 @@ abstract class Base extends TestCase
         'POST:/v1/mock/tests/general/upload:passed',
     ];
 
-    const EXTENDED_GENERAL_RESPONSES = [
+    const DOWNLOAD_RESPONSES = [
         'Download test passed.',
```
lohanidamodar, line 39: @abnegate can we make this response same as others something like `GET:/v1/mock... passed`

tests/WebChromiumTest.php
```diff
@@ -21,6 +21,7 @@ class WebChromiumTest extends Base
         ...Base::BAR_RESPONSES,
         ...Base::GENERAL_RESPONSES,
         ...Base::LARGE_FILE_RESPONSES,
+        "https://appwrite.io/v1/mock/tests/general/download?project=console",
```
lohanidamodar, line 24: Why?

lohanidamodar (changes requested):
Looks good, few comments.

tests/WebChromiumTest.php
abnegate: @lohanidamodar The Web SDK returns a URL instead of the contents of the URL

abnegate:
@christyjacob4 This one will require an update of the test instance to include this change https://github.com/appwrite/appwrite/pull/3638

lohanidamodar:
@abnegate can we update this with latest changes?

lohanidamodar:
@abnegate Node is failing and also Master has updated Flutter tests that are passing please sync and fix node as well.

abnegate:
@Meldiron Any thoughts on how to fix the CLI here? The endpoint returns a string but it looks like CLI hardcodes the type as `arraybuffer` for `location` routes

lohanidamodar:
@abnegate seems node and deno are still failing.

tests/Base.php
```diff
@@ -41,6 +41,14 @@ abstract class Base extends TestCase
         'GET:/v1/mock/tests/general/download:passed',
     ];
```
lohanidamodar, line 43: Let's rename `EXTENDED_GENERAL_RESPONSES` to may be `DOWNLOAD_RESPONSES` ?

lohanidamodar (changes requested):

PR #443, opened 2022-05-13
Fix: Web file error message

Meldiron:
As reported here: https://github.com/appwrite/appwrite/issues/3209

You create file input. you get `event.target.value`. You get an array of files. I consider it a common mistake to try to pass the value directly into Appwrite.

Currently, Appwrite doesn't tell anything. During chunk upload logic the size will be `undefined`, resulting in 0 chunks being uploaded, but also probably no error thrown.

With this new logic, we check if a variable type is `File`, and throw a proper error if you pass an array, buffer, file path, base64, or anything that is unsupported.

Example of generated code:

CleanShot 2022-05-13 at 08 40 00@2x


Dakicka:
Hi, that looks like a helpful solution! Thank you.

TorstenDittmann (approved):

PR #447, opened 2022-05-19
Fix: CLI function deploy

Meldiron:
Currently if you:
```
appwrite login
appwrite init project
appwrite init function
appwrite deploy function
```

It fails in deploy function with message:
CleanShot 2022-05-19 at 09 11 11@2x

Problem solved by applying the same strategy as there already is regarding deploy collections - instead of passing BigIntJSON objects, we pass strings and then parse them. This seems to be a limitation of `inquirer` library.

- [x] Manual QA

CleanShot 2022-05-19 at 09 11 18@2x


lohanidamodar (approved):
LGTM

christyjacob4 (approved):

PR #448, opened 2022-05-19
Feat: Add function runtimes to CLI

Meldiron:
New runtimes were released, so we are adding their entry points into CLI so it is automatically picked up.


lohanidamodar (approved):
LGTM

christyjacob4 (approved):

PR #450, opened 2022-05-22
Adding Scoop support

CEbbinghaus:
This PR is to add scoop support to the CLI install.
With first a github hosted uri to install it manually:
```
> scoop install https://raw.githubusercontent.com/appwrite/sdk-for-cli/master/appwrite.json
```
and hopefully the ability for scoop to verify hashes as well as install it via the [main bucket](https://github.com/ScoopInstaller/Main)
```
> scoop install appwrite
```
which will require a working autoupdater and working version check

This PR implements Feature appwrite/sdk-for-cli/issues/39


src/SDK/Language/CLI.php
```diff
@@ -74,6 +74,12 @@ public function getFiles()
                 'template'      => 'cli/package.json.twig',
                 'minify'        => false,
             ],
+            [
+                'scope'         => 'default',
+                'destination'   => 'appwrite.json',
```
Meldiron, line 79: Can we name it something else? We use the file `appwrite.json` for project configuration used by Appwrite commands (`init, deploy`). It cold create some confusions when looking at CLI SDK

Meldiron: My idea: `scoop.json`

Meldiron (commented):
Left a comment, PR looks good! 🥳 (didn't test)

templates/cli/appwrite.json.twig
```diff
@@ -0,0 +1,23 @@
...
+    "license": "BSD-3-Clause",
+    "url": "https://github.com/{{ sdk.gitUserName }}/{{ sdk.gitRepoName | caseDash }}/releases/download/{{ sdk.version }}/{{ language.params.executableName }}-cli-win-x64.exe",
+    "bin": [
+        [
+            "{{ language.params.executableName }}-cli-win-x64.exe",
+            "{{ language.params.executableName|caseLower }}"
+        ]
+    ],
+    "homepage": "https://github.com/{{ sdk.gitUserName|url_encode }}/{{ sdk.gitRepoName|url_encode }}",
+    "hash": "https://github.com/appwrite/sdk-for-cli/releases/download/$version/hashes",
```
stnguyen90, line 14: I don't think this exists. Where's the update to the job that will generate this.

templates/cli/appwrite.json.twig
```diff
@@ -0,0 +1,23 @@
...
+    "checkver": {
+        "url": "https://api.github.com/repos/appwrite/sdk-for-cli/releases",
+        "jsonpath": "$[0].tag_name"
+    },
+    "autoupdate": {
+        "url": "https://github.com/appwrite/sdk-for-cli/releases/download/$version/appwrite-cli-win-x64.exe",
+        "hash": "$url.sha256"
```
stnguyen90, line 21: I don't think this exists. Where's the update to the job that will generate this.

templates/cli/appwrite.json.twig
```diff
@@ -0,0 +1,23 @@
+{
+    "$schema": "https://raw.githubusercontent.com/ScoopInstaller/Scoop/master/schema.json",
+    "version": "{{ sdk.version }}",
+    "description": "Appwrite cli project manager",
+    "license": "BSD-3-Clause",
+    "url": "https://github.com/{{ sdk.gitUserName }}/{{ sdk.gitRepoName | caseDash }}/releases/download/{{ sdk.version }}/{{ language.params.executableName }}-cli-win-x64.exe",
```
stnguyen90, line 6: Scoop let's you specify architecture as well. It would probably be good to differentiate between arm and x86.

stnguyen90 (changes requested):

src/SDK/Language/CLI.php
CEbbinghaus: I thought the same thing but since the manifest doesn't contain a name the name of the file is the name that it installs the program under. I tried using scoop.json but when I tried it actually uninstalled scoop because it thought this was a new version of it. So to have the program called appwrite it has to be appwrite.json. A little annoying but I think if we keep it in a folder called scoop it could work

templates/cli/appwrite.json.twig
CEbbinghaus: Nope. The architecture difference is x32 vs x64. This is addressed in issue ScoopInstaller/Scoop#3146, however, that is not fixed yet. I am subscribed to it though so as soon as they add support for ARM I will update the manifest to reflect those changes.

templates/cli/appwrite.json.twig
CEbbinghaus: You are right. This doesn't exist right now and I have no idea why it's here since I had taken it out to commit the file. My bad that it got pushed but it will be removed until a proper system has been figured out.

src/SDK/Language/CLI.php
stnguyen90: Scoop folder would be good.

templates/cli/appwrite.json.twig
CEbbinghaus: Wasn't meant to be committed. Will be removed until that update has been figured out

templates/cli/appwrite.json.twig
CEbbinghaus: yes, however, if we want those changes reflected in the JSON file in the commit then the workflow will have to create another commit on the master branch updating the hash. (which personally I don't find very appealing). The two other solutions would be providing the JSON as a release artifact (which I would think is a bit strange) or leaving away the hash for the base file and only having the auto-updater.

The package will work without the hash it just misses the client-side verification but that is only relevant when expecting an MTM attack which is going to be unlikely. Instead, I propose we create the file with the hash embedded and create an automated PR to the main bucket of the scoop installer to update the app there. We don't even have to do this all the time since the auto updater can get the hashes from the URL.

CEbbinghaus:
Install can be tested with this command:
```powershell
$ scoop install https://gist.githubusercontent.com/CEbbinghaus/3219a010325e9684839f7ce5ecea4afd/raw/2f19d33dc4581658d5e73edda4d763644129621a/appwrite.json
```

Generated with the SDK

CEbbinghaus:
The build task seems to be failing because of Docker rate limits not the changes I added

Meldiron (approved):
Looks good to me 😇

Meldiron:
> The build task seems to be failing because of Docker rate limits not the changes I added

Restarted failed jobs. All green now 💚

CEbbinghaus:
Sweet. This PR is ready to merge then as it shouldn't contain any extra changes. It's going to be the prerequisite for the other changes so i am very keen to get it merged asap

CEbbinghaus:
Requesting another review as its still not merged and the merging is blocked. It might be smart to leave it until ScoopInstaller/Scoop#3146 is merged as it would allow for the ARM release to be installed too

lohanidamodar:
@Meldiron looks like the blocking issue was merged can you help move forward with this?

CEbbinghaus:
Thanks @lohanidamodar, since opening this PR I have learned a lot about scoop & how it works. There are some small modifications I want to do to it as it stands right now since it would require a PullRequest to the scroop repository for every version which is not desired. Instead if we can generate the Appwrite.json file a little differently, it should auto update within the scoop repository. That means that the changes will be lagging behind until scoop's main bucket does another update cycle but that is probably faster than waiting for a human review of every version.

templates/cli/scoop/appwrite.json.twig
```diff
@@ -0,0 +1,18 @@
...
+        [
+            "{{ language.params.executableName }}-cli-win-x64.exe",
+            "{{ language.params.executableName|caseLower }}"
+        ]
+    ],
+    "homepage": "https://github.com/{{ sdk.gitUserName|url_encode }}/{{ sdk.gitRepoName|url_encode }}",
+    "checkver": {
+        "url": "https://api.github.com/repos/appwrite/sdk-for-cli/releases",
+        "jsonpath": "$[0].tag_name"
+    }
```
CEbbinghaus, line 17: Since the sdk-for-cli repository now does its releases through actual releases rather than Pre-Releases this can be simplified to just point at the git repository. this removes the burden of maintaining this configuration & keeping it up to date with the Github API.

templates/cli/scoop/appwrite.json.twig
```diff
@@ -0,0 +1,18 @@
...
+    "version": "{{ sdk.version }}",
+    "description": "Appwrite cli project manager",
+    "license": "BSD-3-Clause",
+    "url": "https://github.com/{{ sdk.gitUserName }}/{{ sdk.gitRepoName | caseDash }}/releases/download/{{ sdk.version }}/{{ language.params.executableName }}-cli-win-x64.exe",
+    "bin": [
+        [
+            "{{ language.params.executableName }}-cli-win-x64.exe",
+            "{{ language.params.executableName|caseLower }}"
+        ]
+    ],
```
CEbbinghaus, line 12: This should be changed to include ARM now that Scoop supports it ScoopInstaller/Scoop#3146.

templates/cli/README.md.twig
```diff
@@ -59,9 +59,14 @@ $ brew install --HEAD {{ language.params.executableName }}
 > Please note that `--HEAD` will be removed with official release.
 
 ### Windows
+Via Powershell
 ```powershell
 $ iwr -useb {{ sdk.url }}/cli/install.ps1 | iex
 ```
+Via [Scoop](https://scoop.sh/#/)
```
CEbbinghaus, line 66: Gotta remove the `/#/`since its not necessary

CEbbinghaus (commented):
Not quite ready to be merged. Will try and get these changes in today to bring it up to date. Also of note is that this will generate the json file every single version but only 1 will ever be committed to the main bucket scoop maintains as it self updates the json through the `"checkver"` property.

lohanidamodar:
@CEbbinghaus thank you for your immediate response. Please make the necessary changes and request another review.

CEbbinghaus:
@lohanidamodar who should I request the review from?

stnguyen90 (commented):
One suggestion.

stnguyen90 (approved):

CEbbinghaus:
So this is ready to merge? The failed Travis build seems to be more of an amnesty rather than due to my changes. Is anything hold it up?

stnguyen90 (changes requested):
> So this is ready to merge? The failed Travis build seems to be more of an amnesty rather than due to my changes. Is anything hold it up?

@CEbbinghaus would you please update this branch with the latest from master? There may be something on master that will fix the CI.

CEbbinghaus:
Updated with the latest from master. Lets hope it fixes the CI task :3

stnguyen90 (approved):

Issue #452, opened 2022-05-24
🚀 Feature: Async Client for Python


ms7m:
### 🔖 Feature description

An async client for the python-sdk.

### 🎤 Pitch

I'm planning to move over a couple of applications I maintain to Appwrite, and while in the process of moving I noticed the python-sdk client is non-asynchronous. Mainly due to ``requests`` being used as the http client.

A little bit of background: The applications I maintain are mostly FastAPI-based and they [recommend any I/O based libraries to be asynchronous for better performance.](https://fastapi.tiangolo.com/async/). 
***

I've already got a p.o.c [fork](https://github.com/ms7m/sdk-generator) and opened a PR #453 which passes all tests. This is a very initial version and I definitely make some time to optimize it further, but I think it's a good starting point.

There are a number of changes.

- Breaking: Moving from ``requests`` to [``httpx``](https://www.python-httpx.org/). Httpx is mostly requests-compatible as well contains both a sync and async client. 
    - I removed requests to keep the dependency count low, but I can understand if certain people are unable to install.. Maybe we can publish a spealized version of the sdk with async support? e.g: ```pip install appwrite[async]```?

- Async Client and functions is available under ```appwrite.aio``` and any I/O related functions must be called with ``await``.

There is a lot of room to further optimize, such as maybe using [``AnyIO``](https://anyio.readthedocs.io/en/stable/) for users who use Trio instead of asyncio.

### 👀 Have you spent some time to check if this issue has been raised before?

- [x] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

stnguyen90:
@ms7m, can your description be added to https://github.com/appwrite/sdk-for-python/issues/26 and then we close out this issue as a duplicate since there's a longer history/conversation in that issue?

PR #453, opened 2022-05-24
Async Support in Python SDK

### LINKED ISSUE #26: Idea: Make SDK  asynchronous.
We can optimize the performance of the SDK by making it compatible to run in an asynchronous manner.  
I have a few ideas about where to start. If this aligns with your current goal then I would be happy to contribute.

ms7m:
More information @ #452


templates/python/package/aio/client.py.twig
```diff
@@ -0,0 +1,191 @@
+import io
+import httpx
+import os
+from .input_file import InputFile
+from .exception import {{spec.title | caseUcfirst}}Exception
+
+class Client:
```
abnegate, line 7: Could we use the existing client as a base class and extend here as a child `AsyncClient`? Then we would only need to maintain the async methods in the `aio` package

templates/python/package/aio/services/service.py.twig
```diff
@@ -0,0 +1,83 @@
+from ..service import Service
+from ..exception import AppwriteException
+
+class {{ service.name | caseUcfirst }}(Service):
```
abnegate, line 4: Could we use the existing service as a base class and extend here as a child `AsyncService`? Then we would only need to maintain the async methods in the `aio` package

abnegate (changes requested):
Nice work so far! Just a few comments

ms7m:
> Nice work so far! Just a few comments

Thanks, I've been swamped at work and battling a cold -- but I'll get these changes sorted!

.travis.yml
```diff
@@ -37,8 +37,11 @@ env:
   - SDK=PHP74
   - SDK=PHP80
   - SDK=Python38
+  - SDK=Python39Async
```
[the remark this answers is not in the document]
ms7m: Fixed!

templates/python/package/aio/client.py.twig
ms7m: Updated to create a AsyncClient (inherits Client). Replaces ``call`` and ``chunked_upload`` methods.

stnguyen90:
@ms7m, sorry, but would you please fix the merge conflict?

ms7m:
@stnguyen90 sure, my branch is quite behind since this kinda fell off my radar.

stnguyen90 (changes requested):
@ms7m it seems like the tests are failing.

lohanidamodar:
@abnegate is this relevant?

stnguyen90:
@ms7m seems like the tests are still failing. Any ideas why?

ms7m:
@stnguyen90 -- running locally, only a single test is failing. I'll take a crack at this weekend and get this freed up!

abnegate:
@ms7m Is it possible to have the `async` methods without adding the `AsyncClient`? So the async versions of the methods would be together with the sync versions in the existing client?

iron3oxide:
It's been a while since the last update here and I'd love to see this feature implemented. @ms7m are you still working on it? If not, I might take a crack at it myself

burnthoney:
Are there any updates on this?

PR #457, opened 2022-05-27
web - add index type to Models that have additionalProperties set to …

nartc:
…true

## Current behavior

In the web SDK typings, `Preferences` is a `type Preferences = {}` which makes it tricky to work with in projects with TypeScript strict mode on. 

## Expected behavior

`type Preferences` should be an indexable type or a Record. 
```ts
type Preferences = Record<string, any>;
// or
type Preferences = {
   [key: string]?: any;
}
```

This PR adds an indexable record to a model under `namespace Models` if its definition has `additionalProperties: true`

## Additional information

I'm not sure how to verify the change with the current tests setup of this repo. Any suggestions would be much appreciated

### DIFF
--- a/templates/web/src/sdk.ts.twig
+++ b/templates/web/src/sdk.ts.twig
@@ -13,6 +13,7 @@ namespace Models {
          * {{ property.description }}
          */
         {{ property.name | escapeKeyword }}{% if not property.required %}?{% endif %}: {{_self.sub_schema(property)}};
+        {% if definition.additionalProperties %}[key: string]?: any;{% endif %}
 {% endfor %}
     }
 {% endfor %}


brandonroberts:
@TorstenDittmann @christyjacob4 we had a conversation about this PR already to allow more flexibility with the types. It will help Angular users better extend the base types.

TorstenDittmann:
> @TorstenDittmann @christyjacob4 we had a conversation about this PR already to allow more flexibility with the types. It will help Angular users better extend the base types.

I don't quite remember how we handle types in the preferences, it's possible that the values only allow strings.

TorstenDittmann (approved):

templates/web/src/sdk.ts.twig
```diff
@@ -13,6 +13,7 @@ namespace Models {
          * {{ property.description }}
          */
         {{ property.name | escapeKeyword }}{% if not property.required %}?{% endif %}: {{_self.sub_schema(property)}};
+        {% if definition.additionalProperties %}[key: string]?: any;{% endif %}
```
TorstenDittmann, line 16: actually, this needs to be outside of the loop - otherwise it will create multiple index types 😉

TorstenDittmann (changes requested):

templates/web/src/sdk.ts.twig
TorstenDittmann: also I think this is the correct syntax

```suggestion
        {% if definition.additionalProperties %}[key: string]: any;{% endif %}
```

nartc: Done

PR #460, opened 2022-06-10
PHP File Upload improvements

lohanidamodar:


templates/php/src/Services/Service.php.twig
```diff
@@ -58,43 +59,74 @@ class {{ service.name | caseUcfirst }} extends Service
...
-            fseek($handle, $start);
-            $params['{{ parameter.name }}'] = new \CURLFile('data://' . $mimeType . ';base64,' . base64_encode(@fread($handle, Client::CHUNK_SIZE)), $mimeType, $postedName);
+            $chunk = '';
+            if(!empty($handle)) {
+                fseek($handle, $start);
+                $chunk = @fread($handle, Client::CHUNK_SIZE);
+            } else {
+                $chunk = substr($file->getData(), $start, Client::CHUNK_SIZE);
+            }
+            $params['{{ parameter.name }}'] = new \CURLFile('data://' . $mimeType . ';base64,' . base64_encode($chunk), $mimeType, $postedName);
```
abnegate, line 129: Is there any way we can avoid base64 encoding each chunk?

tests/languages/php/test.php
```diff
@@ -59,10 +61,10 @@
 $response = $general->redirect();
 echo "{$response['result']}\n";
 
-$response = $general->upload('string', 123, ['string in array'], __DIR__ . '/../../resources/file.png');
+$response = $general->upload('string', 123, ['string in array'], InputFile::withPath(__DIR__ .'/../../resources/file.png'));
 echo "{$response['result']}\n";
 
-$response = $general->upload('string', 123, ['string in array'], __DIR__ . '/../../resources/large_file.mp4');
+$response = $general->upload('string', 123, ['string in array'], InputFile::withPath(__DIR__ .'/../../resources/large_file.mp4'));
 echo "{$response['result']}\n";
```
Meldiron, line 69: Can we add `withData` test?

templates/php/src/Services/Service.php.twig
```diff
@@ -58,43 +59,74 @@ class {{ service.name | caseUcfirst }} extends Service
...
 {% if parameter.isUploadID %}
-            if(${{ parameter.name | caseCamel | escapeKeyword }} != 'unique()') {
-                try {
-                    $response = $this->client->call(Client::METHOD_GET, new URL($path . '/' . {{ parameter.name }}), headers);
-                    $counter = $response['chunksUploaded'] ?? 0;
-                } catch(\Exception $e) {
-                }
+        if(${{ parameter.name | caseCamel | escapeKeyword }} != 'unique()') {
+            try {
+                $response = $this->client->call(Client::METHOD_GET, $path . '/' . ${{ parameter.name }},);
```
Meldiron, line 105: You removed headers here, and kept comma. Looks like syntax error, also unwanted change probably.

templates/php/src/InputFile.php.twig
```diff
@@ -0,0 +1,49 @@
...
+    public static function withPath(string $path, ?string $mimeType = null, ?string $filename = null)
+    {
+        $instance = new InputFile();
+        $instance->path = $path;
+        $instance->mimeType = $mimeType;
+        $instance->filename = $filename;
+        return $instance;
+    } 
+
+    public static function withData(string $data, ?string $mimeType = null, ?string $filename = null)
```
Meldiron, line 41: Can we come up with a better name? Data sounds too genetic. For example, in Node I called it `plaintext`, not sure if it's the same concept here.

templates/php/src/Services/Service.php.twig
lohanidamodar: We can drop later when we only support PHP 8.1 and above

templates/php/src/InputFile.php.twig
lohanidamodar: I've seen data been common for file data as string in PHP?

Meldiron: Let's keep it if it's common in PHP.

Meldiron (approved):
LGTM

PR #461, opened 2022-06-12
Flutter & Dart template refactoring

lohanidamodar:


abnegate (approved):
Looks good, much easier to manage, my only suggestion would be moving the api, file and oauth templates to a separate directory like `requests` or similar to keep concerns isolated

lohanidamodar:
> Looks good, much easier to manage, my only suggestion would be moving the api, file and oauth templates to a separate directory like `requests` or similar to keep concerns isolated

Well I added those to a separate directory called `base` as these are base templates or parts of templates. May be components/requests might be a better name for the directory. Let me know what you think sounds better.

abnegate:
Yep sounds good

templates/flutter/base/macros.twig
```diff
@@ -0,0 +1,5 @@
+{% macro parameter(parameter) %}{% if parameter.required %}required {{ parameter.type | typeName }}{% else %}{{ parameter.type | typeName }}?{% endif %} {{ parameter.name | caseCamel | escapeKeyword }}{% endmacro %}
```
eldadfux, line 1: Do we really need this as an individual template? Can it be just part of the service template? If not, I'm not sure macros is the most intuitive name.

lohanidamodar: It I keep these in service, I can't call the function in `api.twig` or other included template files, keeping this in separate file allows us to import and use it in any template files. so we can define all macros in a separate file and import and use from any file. What might be a better name? `functions` or `utils` ?

templates/flutter/base/api.twig
```diff
@@ -0,0 +1,34 @@
...
+{% for node in method.security %}
+{% for key,header in node|keys %}
+            '{{header|caseLower}}': client.config['{{header|caseLower}}'],
+{% endfor %}
+{% endfor %}
+{% endif %}
+{% endif %}
+        };
+
+{% if method.type == 'location' %}
```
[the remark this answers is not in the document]
lohanidamodar: Done!

Meldiron (approved):

abnegate (approved):

Meldiron (approved):
I love it! LGTM

PR #463, opened 2022-06-20
php template refactor

lohanidamodar:
- PHP templates refactored


Meldiron (approved):

templates/php/base/requests/api.twig
```diff
@@ -0,0 +1,8 @@
+        return $this->client->call(Client::METHOD_{{ method.method | caseUpper }}, $path, [
```
Meldiron, line 1: I noticed some files does not start on the left, is that on purpose?

lohanidamodar: yes to maintain the formatting on the generated SDK

PR #471, opened 2022-06-28
Update swift input file

lohanidamodar:


tests/languages/swift-server/Tests.swift
```diff
@@ -68,36 +68,57 @@ class Tests: XCTestCase {
...
-        }
-
-        do {
-            try await general.error502()
-        } catch {
-            print(String(describing: error))
-        }
-
-        try! await general.empty()
+//         do {
```
lohanidamodar, line 78: @abnegate All these tests are commented out?

abnegate: This was WIP at the time, sorted now 👍

PR #490, opened 2022-07-05
feat: refactor go templates

everly-gif:
refactor go templates


lohanidamodar (approved):
Looks good, we should figure out a way to extract the upload script from client to file.twig may be after chunked upload implementation

Meldiron (approved):

PR #496, opened 2022-07-25
Fix twig templates to add InputFile imports

gewenyu99:
`InputFile` is not imported in examples. This adds the import to relevant SDKs.


lohanidamodar (approved):
LGTM

PR #500, opened 2022-08-01
feat: refactor kotlin templates

everly-gif:
refactor kotlin templates

### DIFF
--- a/templates/kotlin/base/macros/methodNeedsSecurityParameters.twig
+++ b/templates/kotlin/base/macros/methodNeedsSecurityParameters.twig
@@ -1,0 +1,3 @@
+{% macro methodNeedsSecurityParameters(method) %}
+{% if (method.type == "webAuth" or method.type == "location") and method.security|length > 0 %}{{ true }}{% else %}{{false}}{% endif %}
+{% endmacro %}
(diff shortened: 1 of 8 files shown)


templates/kotlin/base/macros/methodNeedsSecurityParameters.twig
```diff
@@ -0,0 +1,3 @@
+{% macro methodNeedsSecurityParameters(method) %}
```
lohanidamodar, line 1: if macros are used in only one place place it in that file only. Let's keep all macros that are used more than once together as utils.twig

lohanidamodar (changes requested):

templates/kotlin/base/requests/location.twig
```diff
@@ -0,0 +1,6 @@
+        return client.call(
```
abnegate, line 1: This template needs to import the `resultType` macro

templates/kotlin/base/macros/methodNeedsSecurityParameters.twig
everly-gif: okay sure .

templates/kotlin/base/requests/location.twig
everly-gif: thank you !

abnegate (approved):

lohanidamodar (approved):

PR #504, opened 2022-08-03
Support nested array types

abnegate:
Generates array types with their item type as a generic type parameter

E.g. `List<String>` instead of `List` or `List<Any>` for applicable languages, when a parameters 'array' 'type' property contains 'string'.


lohanidamodar (approved):
Looks good. Make sure by generating SDKs with latest spec as well.

PR #523, opened 2022-08-22
WIP using example params

lohanidamodar:
## What does this PR do?

- [x] Dart
- [x] Flutter
- [x] Deno
- [x] Web
- [x] Node
- [x] Php

Swagger spec
- parameters can have `x-example-filter` with the name of the custom filter
- methods can have `x-example-params` with the name of params apart from required ones that we want to display on the example

- As collection level and document level support different permissions can we have separate filters for `collection`, `bucket` params and `document`, `files` permissions params @TorstenDittmann ?

## Test Plan

## Related PRs and Issues

(Write your answer here.)

### DIFF
--- a/src/SDK/SDK.php
+++ b/src/SDK/SDK.php
@@ -560,7 +560,6 @@ class SDK
                     foreach ($this->spec->getServices() as $key => $service) {
                         $methods = $this->spec->getMethods($key);
                         $params['service'] = [
-                            'globalParams' => $service['globalParams'] ?? [],
                             'description' => $service['description'] ?? '',
                             'name' => $key,
                             'features' => [
@@ -588,7 +587,6 @@ class SDK
                         $params['service'] = [
                             'name' => $key,
                             'methods' => $methods,
-                            'globalParams' => $service['globalParams'] ?? [],
                             'features' => [
                                 'upload' => $this->hasUploads($methods),
                                 'location' => $this->hasLocation($methods),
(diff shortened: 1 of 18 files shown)


src/SDK/SDK.php
```diff
@@ -82,9 +82,13 @@ public function __construct(Language $language, Spec $spec)
...
             $this->twig->addFilter($filter);
         }
-
+        
         $this->twig->addExtension(new \Twig\Extension\DebugExtension());
 
+        $this->twig->addFilter(new TwigFilter('customLanguageFilter', function($value, $filter) use ($language) {
+            return call_user_func([$language, $filter], $value);
+        }));
+
```
TorstenDittmann, line 91: not sure about that one, it might be better to be more explicit in the languages itself.

lohanidamodar: as this will be need in all language, doing it this way gives a nice error, if any language not implement any of the functions that are in the specs as well.

Also, even if we do it in each language, we do have to implement same customFilter in all the languages that will just call the functions. it's same for all the languages.

Or are you thinking another way?

TorstenDittmann: I think that this is like magic. If we want all languages to adapt these methods, we can have them as abstract in the base class or as an interface.

Also what is `customLanguageFilter`? The naming is not 100% clear imo 👍🏻

lohanidamodar: @TorstenDittmann What I want to achieve is,
The name of the filter we need to apply to the parameter is available from the spec, so I need to call the filter dynamically for each parameter if it exists. The only way I found was by calling a another filter that would accept the name of the filter, so `customLanguageFilter` is just a filter that calls specific filter method inside the language so that we can get proper example for that parameter based on the name of the function. If you have better way to achieve this I'd be more than happy.

PR #535, opened 2022-09-03
Fix PHP param syntax

Meldiron:
## What does this PR do?

Before:

```php
public function list(string$search = null, int $limit = null, int $offset = null, string$cursor = null, string$cursorDirection = null, string$orderType = null): array
```

> Syntax errors

After:

```php
public function list(string $search = null, int $limit = null, int $offset = null, string $cursor = null, string $cursorDirection = null, string $orderType = null): array
```

## Test Plan

- [x] Run example.php, visual check

## Related PRs and Issues

x

Yes

### DIFF
--- a/src/SDK/Language/PHP.php
+++ b/src/SDK/Language/PHP.php
@@ -238,7 +238,8 @@ class PHP extends Language {
     {
         switch ($parameter['type']) {
             case self::TYPE_STRING:
-                return 'string';
+                $type = 'string';
+                break;
             case self::TYPE_BOOLEAN:
                 $type = 'bool';
                 break;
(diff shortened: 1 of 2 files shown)


Meldiron:
Also tested all the other SDKs properly to ensure no syntax errors are in them.

CI/CD tests all passed except CLI because of the wrong endpoint:

CleanShot 2022-09-03 at 19 51 27

CLI tests then passed on Gitpod:

CleanShot 2022-09-03 at 19 52 09

CleanShot 2022-09-03 at 19 52 42

stnguyen90 (approved):

christyjacob4 (approved):

PR #560, opened 2022-09-28
Add type for node main function

daurensky:
## What does this PR do?

Adds type for main function

## Example
Code


abnegate:
Thank you for the PR 🙏  We appreciate the contribution, however, we are looking at other approaches to solve this problem. This would work when the `node-appwrite` package is used as a dependency but functions are made to work with or without Appwrite packages

PR #564, opened 2022-10-01
Feat: CLI settings init&deploy

Meldiron:
## What does this PR do?

As implemented by init&deploy collection, this PR adds the same logic for project settings. It' common practice to have a predefined structure of a project (enabled services, auth methods, users limit), just like it's common for databases. Since appwrite.json is supposed to be a source of truth for initial setup of Appwrite project, it should project settings.

This PR gets us one small step closer to "Deploy to Appwrite" 1-click setup.

Settings include:
- Max users limit
- Status (enabled/disabled) of auth methods (not OAuth)
- Status (enabled/disabled) of project services

## Test Plan

- [x] Manual QA

https://user-images.githubusercontent.com/19310830/193423232-575a8d9a-4522-4340-85ea-46a788c4d0fc.mp4

## Related PRs and Issues

x

Yes

### DIFF
--- a/templates/cli/lib/commands/deploy.js.twig
+++ b/templates/cli/lib/commands/deploy.js.twig
@@ -2,7 +2,7 @@ const inquirer = require("inquirer");
 const JSONbig = require("json-bigint")({ storeAsString: false });
 const { Command } = require("commander");
 const { localConfig } = require("../config");
-const { questionsDeployFunctions, questionsGetEntrypoint, questionsDeployCollections, questionsConfirmDeployCollections } = require("../questions");
+const { questionsDeploySettings, questionsDeployFunctions, questionsGetEntrypoint, questionsDeployCollections, questionsConfirmDeployCollections } = require("../questions");
 const { actionRunner, success, log, error, commandDescriptions } = require("../parser");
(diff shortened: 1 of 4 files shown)


templates/cli/lib/commands/deploy.js.twig
```diff
@@ -593,6 +599,80 @@ const deployCollection = async ({ all, yes } = {}) => {
     }
 }
 
+const deploySettings = async ({ yes } = {}) => {
```
lohanidamodar, line 602: Should we seprate these settings into individual deploys? Might be confusing as these are independent settings?

lohanidamodar (commented):
Looks good, small query

templates/cli/lib/commands/deploy.js.twig
Meldiron: We currently apply:
- OAuth enable/disable
- Services enable/disable
- Max users limit

I can see how it would be useful to only specify one of them to deploy. I can implement a similar passion to what we do in `deploy functions` where you either select ones to deploy manually, or do `--all` to auto-pick all of them. Sounds good?

lohanidamodar: yes that sounds good, otherwise people might be confused as to what settings will be deployed, initialized

lohanidamodar:
Skipping it for now, closing, we can reopen if we want to go ahead with this later.

PR #566, opened 2022-10-03
Display README as long description on PyPI (Python)

leha-code:
## What does this PR do?
This PR fixes the lack of README contents on PyPI by adding the `long_description` option into the setup.py template.
There was a previous invalid setup.cfg file that was supposed to add the README contents as long description but since that isn't needed anymore it's deleted.

## Test Plan

Almost no code has been changed.
When generating an SDK the template takes out the sdk.readme option for the long description

## Related PRs and Issues

Yes


templates/python/setup.py.twig
```diff
@@ -6,6 +6,7 @@ setuptools.setup(
   version = '{{sdk.version}}',
   license='{{spec.licenseName}}',
   description = '{{sdk.shortDescription}}',
+  long_description = '{{sdk.readme}}',
```
abnegate, line 9: The `readme` property is not populated from the Appwrite side, let's use the `gettingStarted` content instead

```suggestion
  long_description = '{{sdk.gettingStarted}}',
```

abnegate (changes requested):
Thanks for the PR! Looks good, just one small change 🙏

abnegate (approved):
Awesome thanks for the PR :grin:

PR #575, opened 2022-10-17
Doc updates

abnegate:
## What does this PR do?

Updates docs:
- Fix types for Java examples
- Clean up Kotlin examples
- Clean up Swift/Apple examples

## Test Plan

Manual generation

## Related PRs and Issues

N/A

### Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?

Yes

PR #579, opened 2022-10-21
Feat improve java interop

abnegate:
## What does this PR do?

Add coroutine continuation wrapper for easier use from Java

Before:
```java
        account.get(new Continuation<io.appwrite.models.Account>() {
            @NonNull
            @Override
            public CoroutineContext getContext() {
                return EmptyCoroutineContext.INSTANCE;
            }

            @Override
            public void resumeWith(@NonNull Object o) {
                if (o instanceof Result.Failure) {
                    Result.Failure failure = (Result.Failure) o;
                    failure.exception.printStackTrace();
                    return;
                }
                
                System.out.println((io.appwrite.models.Account) o);
            }
        });
```

After:
```java
        account.get(new CoroutineCallback<>((account, error) -> {
            if (error != null) {
                error.printStackTrace();
                return;
            }

            System.out.println(account)
        }));
```

## Test Plan

Manual test

## Related PRs and Issues

Supercedes https://github.com/appwrite/sdk-generator/pull/575

Yes


templates/swift/docs/example.md.twig
```diff
@@ -1,25 +1,24 @@
...
 {% if loop.last %}
 
-    )
+)
 {% endif %}
 
-{% endfor %}
-    print(String(describing: {% if method.type == 'webAuth' %}success{% elseif method.type == 'location' %}byteBuffer{% elseif method.responseModel | length == 0 %}result{% else %}{{ method.responseModel | caseCamel | escapeKeyword }}{% endif %})
-}
+{% endfor %}
```
stnguyen90, line 24: should this file be in this PR?

stnguyen90 (changes requested):
I didn't test any of this and I know barely anything about Continuations, but the syntax looks great! It's really clean! 

I'm only concerned about how much it will bother people with there being such a breaking change 😕

templates/swift/docs/example.md.twig
abnegate: Not explicitly, this PR depends on https://github.com/appwrite/sdk-generator/pull/575, once that is merged, this change will be gone from here

abnegate:
> I'm only concerned about how much it will bother people with there being such a breaking change 😕

@stnguyen90 It's not breaking 😉 `CoroutineCallback` extends `Continuation` so you can pass either

stnguyen90 (approved):

PR #583, opened 2022-10-24
Fix the `InputFIle.fromBlob` method in NodeJS SDK

arnu515:
## What does this PR do?

This PR awaits the `blob.arrayBuffer()` function in the `InputFile.fromBlob` method, as a result, making `fromBlob` async. This is required because the `blob.arrayBuffer()` returns `Promise<ArrayBuffer>` instead of just `ArrayBuffer`, which will cause a type error when it is used without `await`.

Here's the error screenshot:
The error reads "The "string" argument must be of type string or an instance of Buffer or ArrayBuffer. Received an instance of Promise"

> The "string" argument must be of type string or an instance of Buffer or ArrayBuffer. Received an instance of Promise

I also suggest making other `InputFile` methods async so there will be no difference in calling this one and the others.

## Test Plan

I have not written any tests, seeing as there aren't any tests for `fromBlob` to begin with.

## Related PRs and Issues

None

Yes

### DIFF
--- a/templates/node/lib/inputFile.js.twig
+++ b/templates/node/lib/inputFile.js.twig
@@ -21,8 +21,12 @@ class InputFile {
 
   static fromBlob = async (blob, filename) => {
     const buffer = await blob.arrayBuffer();
-    const stream = Readable.from(buffer.toString());
-    const size = Buffer.byteLength(buffer);
+    const stream = new Readable({
+      read() {
+        this.push(buffer);
+        this.push(null);
+      }
+    });
     return new InputFile(stream, filename);
   };
 
--- a/templates/node/lib/inputFile.js.twig
+++ b/templates/node/lib/inputFile.js.twig
@@ -19,8 +19,8 @@ class InputFile {
     return new InputFile(stream, filename, size);
   };
 
-  static fromBlob = (blob, filename) => {
-    const buffer = blob.arrayBuffer();
+  static fromBlob = async (blob, filename) => {
+    const buffer = await blob.arrayBuffer();
     const stream = Readable.from(buffer.toString());
     const size = Buffer.byteLength(buffer);
     return new InputFile(stream, filename);


arnu515:
image
I have also noticed that the `size` parameter is unused, and there are no typings for the `new InputFile` constructor, only its methods.

I can fix those if you want. Just LMK :)

arnu515:
I also saw that the method used to convert buffers to stream doesn't work anymore, so I've fixed it with the help of [this stackoverflow answer](https://stackoverflow.com/a/54136803). I think you may need to do this with every method

arnu515:
I'd recommend fixing just the `fromBuffer` method, and making all other methods call that one instead of repeating it multiple times. 

This is how it will look like. I haven't added it yet

```js
const { Readable } = require('stream');
const fs = require('fs');
const { promisify } = require('util');

class InputFile {
  stream; // Content of file, readable stream
  size; // Total final size of the file content
  filename; // File name

  static fromPath = (filePath, filename) => {
    const stream = fs.createReadStream(filePath);
    const { size } = fs.statSync(filePath);
    return new InputFile(stream, filename, size);
  };

  // modified
  static fromBuffer = (buffer, filename) => {
    const stream = new Readable({
      read() {
        this.push(Buffer.from(buffer));
        this.push(null);
      }
    });
    const size = Buffer.byteLength(buffer);
    return new InputFile(stream, filename, size);
  };

  // modified
  static fromBlob = async (blob, filename) => {
    const buffer = await blob.arrayBuffer();
    return InputFile.fromBuffer(buffer, filename);
  };

  static fromStream = (stream, filename, size) => {
    return new InputFile(stream, filename, size);
  };

  // modified
  static fromPlainText = (content, filename) => {
    const buffer = Buffer.from(content, "utf-8");
    return InputFile.fromBuffer(buffer, filename);
  };

  constructor(stream, filename, size) {
    this.stream = stream;
    this.filename = filename;
    this.size = size;
  }
}

module.exports = InputFile;
```

PineappleIOnic:
Thank you so much for the PR 🤩. We're adding the `hacktoberfest-accepted` label to ensure this PR counts towards your Hacktoberfest contributions count. With that said, please stay active on this PR to address any comments once you receive a review. Happy Hacktoberfest! 🎃

templates/node/lib/inputFile.js.twig
```diff
@@ -19,10 +19,14 @@ class InputFile {
...
-    const size = Buffer.byteLength(buffer);
+  static fromBlob = async (blob, filename) => {
+    const buffer = await blob.arrayBuffer();
+    const stream = new Readable({
+      read() {
+        this.push(buffer);
+        this.push(null);
+      }
+    });
     return new InputFile(stream, filename);
```
stnguyen90, line 30: If a stream is needed, why not use [`blob.stream()`](https://nodejs.org/docs/latest-v18.x/api/buffer.html#blobstream)?

```suggestion
  static fromBlob = async (blob, filename) => {
    const stream = blob.stream();
    const size = blob.size;
    return new InputFile(stream, filename, size);
```

stnguyen90:
> I also saw that the method used to convert buffers to stream doesn't work anymore

What code doesn't work?

templates/node/lib/inputFile.js.twig
arnu515: Yeah it could be used I guess, but it was only recently added
image

This is how I usually convert blobs to streams, so that's what I've put here, but your method is probably better IMO.

arnu515:
> > I also saw that the method used to convert buffers to stream doesn't work anymore
> 
> What code doesn't work? This?
> 
> ```js
>   static fromBuffer = (buffer, filename) => {
>     const stream = Readable.from(buffer.toString());
>     const size = Buffer.byteLength(buffer);
>     return new InputFile(stream, filename, size);
>   };
> ```

Yes this one. It gives an error

stnguyen90:
> Yes this one. It gives an error

@arnu515, that one didn't really work for a while, but it didn't give an error. It just uploaded an incorrectly encoded file. What error are you seeing now?

arnu515:
Oh yeah that's what it did. It didn't give an error, it just uploaded a bad file. My bad

-------- Original Message --------

templates/node/lib/inputFile.js.twig
stnguyen90: Hmm looks like Node 14 is EOL 2023-04-30 so we might be safe just support Node 16... @lohanidamodar, what are your thoughts?

stnguyen90:
@arnu515 thanks for confirming! 🙏🏼 Would you like to verify the [workaround](https://github.com/appwrite/sdk-for-node/issues/36#issuecomment-1249527671) still works and maybe add the fix to this PR?

arnu515:
Yeah, just give me a few minutes

arnu515:
Yes, looks like the workaround that is given does work

arnu515:
image

With this code, the file does get uploaded

stnguyen90:
> With this code, the file does get uploaded. But I tested using buffers, not blobs

@arnu515, `InputFile.fromBuffer()` expects a buffer rather than blob, though, so that should be fine.

stnguyen90:
@arnu515 what's the error you're seeing?

templates/node/lib/inputFile.js.twig
stnguyen90: We discussed internally and decided to make sure we support node 14 at least until EOL.

templates/node/lib/inputFile.js.twig
```diff
@@ -19,10 +19,14 @@ class InputFile {
     return new InputFile(stream, filename, size);
   };
 
-  static fromBlob = (blob, filename) => {
-    const buffer = blob.arrayBuffer();
-    const stream = Readable.from(buffer.toString());
-    const size = Buffer.byteLength(buffer);
+  static fromBlob = async (blob, filename) => {
```
stnguyen90, line 22: So, I tested this like:


```javascript
const buffer = require('buffer');
const fs = require('fs');
const { InputFile } = require('node-appwrite');

const filepath = './resources/nature.jpg';
const filename = path.basename(filepath);
const fileBuffer = fs.readFileSync(filepath);
const fileBlob = new buffer.Blob(fileBuffer);
const inputFile = await InputFile.fromBlob(fileBlob, filename);
const response = await storage.createFile(
    bucketId,
    ID.unique(),
    inputFile,
    [
        Permission.read(Role.any()),
        Permission.update(Role.users()),
        Permission.delete(Role.users()),
    ]
);
```

but I'm getting this error:

```
Uncaught TypeError TypeError [ERR_INVALID_ARG_TYPE]: The "chunk" argument must be of type string or an instance of Buffer or Uint8Array. Received an instance of ArrayBuffer
    at __node_internal_captureLargerStackTrace (internal/errors:465:5)
    at NodeError (internal/errors:372:5)
    at readableAddChunk (internal/streams/readable:260:13)
    at Readable.push (internal/streams/readable:228:10)
    at read /appwrite-playground-for-node/node_modules/node-appwrite/lib/inputFile.js:26:14)
    at Readable.read (internal/streams/readable:487:10)
    at resume_ (internal/streams/readable:987:12)
    at processTicksAndRejections (internal/process/task_queues:83:21)
    --- TickObject ---
    at init (internal/inspector_async_hook:25:19)
    at emitInitNative (internal/async_hooks:201:43)
    at emitInitScript (internal/async_hooks:506:3)
    at nextTick (internal/process/task_queues:133:5)
    at onDestroy (internal/streams/destroy:96:15)
    at Readable._destroy (internal/streams/readable:216:3)
    at _destroy (internal/streams/destroy:102:25)
    at destroy (internal/streams/destroy:64:5)
    at errorOrDestroy (internal/streams/destroy:206:12)
    at readableAddChunk (internal/streams/readable:266:5)
    at Readable.push (internal/streams/readable:228:10)
    at read /appwrite-playground-for-node/node_modules/node-appwrite/lib/inputFile.js:26:14)
    at Readable.read (internal/streams/readable:487:10)
    at resume_ (internal/streams/readable:987:12)
    at processTicksAndRejections (internal/process/task_queues:83:21)
    --- TickObject ---
    at init (internal/inspector_async_hook:25:19)
    at emitInitNative (internal/async_hooks:201:43)
    at emitInitScript (internal/async_hooks:506:3)
    at nextTick (internal/process/task_queues:133:5)
    at resume (internal/streams/readable:980:13)
    at Readable.resume (internal/streams/readable:971:5)
    at Readable.on (internal/streams/readable:885:12)
    at Readable.pipe (internal/streams/readable:751:7)
    at <anonymous> /appwrite-playground-for-node/node_modules/node-appwrite/lib/services/storage.js:442:25)
    at createFile /appwrite-playground-for-node/node_modules/node-appwrite/lib/services/storage.js:389:22)
    at uploadFile /appwrite-playground-for-node/src/app.js:322:40)
    at processTicksAndRejections (internal/process/task_queues:96:5)
    --- async function ---
    at runAllTasks /appwrite-playground-for-node/src/app.js:539:11)
    at processTicksAndRejections (internal/process/task_queues:96:5)
    --- async function ---
    at <anonymous> /appwrite-playground-for-node/src/app.js:562:1)
    at Module._compile (internal/modules/cjs/loader:1105:14)
    at Module._extensions..js (internal/modules/cjs/loader:1159:10)
    at Module.load (internal/modules/cjs/loader:981:32)
    at Module._load (internal/modules/cjs/loader:822:12)
    at executeUserEntryPoint (internal/modules/run_main:77:12)
    at <anonymous> (internal/main/run_main_module:17:47)
```

stnguyen90 (changes requested):

stnguyen90:
Closing stale PR.

PR #577, opened 2022-10-20
Feat graphql

abnegate:
## What does this PR do?

- Add `x-sdk-graphql` header for methods with `graphql` method type, to allow Appwrite to extract parameters from a parameterless route
- Allow list _or_ object responses for strongly typed languages that did not already (Android/Kotlin/Swift/Apple), for GraphQL batched requests which return a list of objects
- Add GraphQL language
- Add GraphQL example template
- Fix REST example template
- Add abstract base class HTTP which REST and GraphQL extend

## Test Plan

Existing tests

## Related PRs and Issues

https://github.com/appwrite/docs/pull/287
https://github.com/appwrite/homepage/pull/113
https://github.com/appwrite/appwrite/pull/974

### Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?

Yes

PR #584, opened 2022-10-26
Fix http examples

abnegate:
## What does this PR do?

Fix HTTP examples
- File uploads
- Differentiate method `security` and method `auth` to allow correct `set` methods in SDK examples and correct headers in HTTP examples. Rename existing usages to `auth` to align with what is represented by the spec.

## Test Plan

Existing tests + manual docs checks

## Related PRs and Issues

https://github.com/appwrite/sdk-generator/pull/577

Yes

### DIFF
--- a/src/Spec/Swagger2.php
+++ b/src/Spec/Swagger2.php
@@ -137,13 +137,15 @@ class Swagger2 extends Spec
 
         foreach ($paths as $pathName => $path) {
             foreach ($path as $methodName => $method) {
-                $auth = $method['x-appwrite']['auth'] ?? [];
+                $methodAuth = $method['x-appwrite']['auth'] ?? [];
+                $methodSecurity = $method['security'][0] ?? [];
 
                 if (isset($method['tags']) && is_array($method['tags']) && in_array($service, $method['tags'])) {
-                    if (isset($auth) && is_array($auth)) {
-                        foreach ($auth as $i => $node) {
-                            $auth[$i] = (array_key_exists($i, $security)) ? $security[$i] : [];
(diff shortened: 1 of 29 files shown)


src/Spec/Swagger2.php
```diff
@@ -169,7 +171,8 @@ public function getMethods($service)
                         'packaging' => $method['x-appwrite']['packaging'] ?? false,
                         'title' => $method['summary'] ?? '',
                         'description' => $method['description'] ?? '',
-                        'security' => [$auth] ?? [],
+                        'auth' => [$methodAuth] ?? [],
+                        'security' => [$methodSecurity] ?? [],
```
lohanidamodar, line 175: we are adding auth and security, are we using `security` for anything?

lohanidamodar (commented):
Looks good, just a query.

src/Spec/Swagger2.php
abnegate: To clarify: we already had `security`, but it was using `auth` from the spec. I have added `auth` and made this use `auth` from the spec, and changed the existing `security` to use `security` from the spec.

We are only using `security` in the HTTP examples here: https://github.com/appwrite/sdk-generator/blob/969bfa308f9504b7aaa45fa064f9e2ca83858343/templates/http/docs/example.md.twig#L10

lohanidamodar: Allright.

lohanidamodar (approved):
Looks good. Make sure we are not missing the `security` => `auth` change anywhere

PR #600, opened 2022-12-29
Fix Node InputFile

stnguyen90:
## What does this PR do?

* Fix InputFile helpers to avoid `toString()` because it can change the actual data and corrupt the upload
* Fix typing for InputFile to expose the constructor and use correct parameter types
* Fix chunk slicing because `slice()` takes a start and end rather than start and length
* Fix incorrect `Response` return type

## Test Plan

Manually tested using playground-for-node and updated the `uploadFile()` function to:

```javascript
const { Readable } = require('stream');
const buffer = require('buffer');

const uploadFile = async () => {
    console.log(chalk.greenBright('Running Upload File API'));

    // const filepath = './resources/nature.jpg';
    const filepath = './resources/big.png';
    const filename = path.basename(filepath);
    const fileBuffer = fs.readFileSync(filepath);
    const fileBlob = new buffer.Blob([fileBuffer]);

    const inputFiles = [];

    inputFiles.push(InputFile.fromPath(filepath, filename));
    inputFiles.push(InputFile.fromBuffer(fileBuffer, filename));
    inputFiles.push(await InputFile.fromBlob(fileBlob, filename));
    inputFiles.push(InputFile.fromPlainText('Hello World!', 'plain.txt'));

    for (const inputFile of inputFiles) {
        const response = await storage.createFile(
            bucketId,
            ID.unique(),
            inputFile,
            [
                Permission.read(Role.any()),
                Permission.update(Role.users()),
                Permission.delete(Role.users()),
            ]
        );
        fileId = response.$id;
        console.log(response);
    }

}
```

where big.png is an image bigger than 5MB.

The result was 3 correctly uploaded images and the plaintext file:

image

## Related PRs and Issues

* https://github.com/appwrite/sdk-for-node/issues/36
* https://github.com/appwrite/sdk-for-node/issues/48

Yes

### DIFF
--- a/templates/node/index.d.ts.twig
+++ b/templates/node/index.d.ts.twig
@@ -237,7 +237,7 @@ declare module "{{ language.params.npmPackage|caseDash }}" {
      * @throws {{ '{' }}{{ spec.title | caseUcfirst}}Exception}
      * @returns {Promise}
      */
-    {{ method.name | caseCamel }}{% if generics %}<{{generics}}>{% endif %}({% for parameter in method.parameters.all %}{{ parameter.name | caseCamel | escapeKeyword }}{% if not parameter.required %}?{% endif %}: {{ parameter | typeName }}{% if not loop.last %}, {% endif %}{% endfor %}): Promise<{% if method.type == 'location' %}Buffer{% else %}{% if method.responseModel and method.responseModel != 'any' %}{% if not spec.definitions[method.responseModel].additionalProperties %}Models.{% endif %}{{method.responseModel | caseUcfirst}}{% if generics_return %}<{{generics_return}}>{% endif %}{% else %}Response{% endif %}{% endif %}>;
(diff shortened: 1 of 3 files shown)


lohanidamodar (approved):

templates/node/index.d.ts.twig
```diff
@@ -131,11 +131,13 @@ declare module "{{ language.params.npmPackage|caseDash }}" {
 
     static fromBuffer(buffer: Buffer, filename: string): InputFile;
 
-    static fromBlob(blob: Blob, filename: string): InputFile;
+    static fromBlob(blob: buffer.Blob, filename: string): Promise<InputFile>;
```
Meldiron, line 134: Above we do `buffer: Buffer`, so here I would expect `blob: Buffer.Blob` (with uppercase B). Is it lowercase here on purpose?

Meldiron, line 134: We are adding `Promise` here but not into other functions  (fromStream, fromPlainText, ...). Is that on purpose?

stnguyen90: No type complaints when using `buffer`. Also `Buffer` is the class while `buffer` is the module.

stnguyen90: Yes, because this is the only async function.

Meldiron (approved):

PR #604, opened 2023-01-11
Fix Flutter websocket issue

lohanidamodar:
## What does this PR do?

- Propagate websocket close and error to realtime subscription

## Test Plan

## Related PRs and Issues

https://github.com/appwrite/sdk-for-flutter/issues/108

(Write your answer here.)

### DIFF
--- a/templates/flutter/lib/src/realtime_mixin.dart.twig
+++ b/templates/flutter/lib/src/realtime_mixin.dart.twig
@@ -76,6 +76,23 @@ mixin RealtimeMixin {
             }
             break;
         }
+      }, onDone: () {
+        for (var list in _channels.values) {
+          for (var stream in list) {
+            stream.close();
+          }
+        }
+        _closeConnection();
+      }, onError: (err, stack) {
+        for (var list in _channels.values) {
+          for (var stream in list) {
+            stream.sink.addError(err, stack);
+          }
+        }
+        if (_websok?.closeCode != null && _websok?.closeCode != 1008) {
+          debugPrint("Reconnecting in one second.");
+          Future.delayed(Duration(seconds: 1), _createSocket);
+        }
       });
     } catch (e) {
       if (e is {{spec.title | caseUcfirst}}Exception) {
--- a/templates/flutter/lib/src/realtime_mixin.dart.twig
+++ b/templates/flutter/lib/src/realtime_mixin.dart.twig
@@ -82,6 +82,7 @@ mixin RealtimeMixin {
             stream.close();
           }
         }
+        _channels = {};
         _closeConnection();
       }, onError: (err, stack) {
         for (var list in _channels.values) {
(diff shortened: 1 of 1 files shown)


templates/flutter/lib/src/realtime_mixin.dart.twig
```diff
@@ -76,6 +76,23 @@ mixin RealtimeMixin {
             }
             break;
         }
+      }, onDone: () {
+        for (var list in _channels.values) {
+          for (var stream in list) {
+            stream.close();
+          }
+        }
+        _closeConnection();
```
stnguyen90, line 85: I think we should try to reconnect if there are channels or something similar to how we do for [web](https://github.com/appwrite/sdk-for-web/blob/master/src/client.ts#L257)

My server times out realtime connections after 2 minutes and using web, it reconnects automatically. Using the flutter SDK, it doesn't try to reconnect.

stnguyen90, line 85: I think we should try to reconnect if there are channels or something similar to how we do for [web](https://github.com/appwrite/sdk-for-web/blob/master/src/client.ts#L257)

My server times out realtime connections after 2 minutes and using web, it reconnects automatically. Using the flutter SDK, it doesn't try to reconnect.

lohanidamodar: yes i'm struggling with that, can you try ?

stnguyen90: Try what? I was suggesting to reconnect in this callback if there were still channels instead of just closing.

lohanidamodar: It might be counterproductive to try to reconnect in the callback if the connection was terminated when there was no internet connection?

stnguyen90: I think there may be other reasons why a connection might close, though.

stnguyen90: I guess we can revise this later.

lohanidamodar:
@stnguyen90 if there's nothing else, I would love to get this merged and released?

templates/flutter/lib/src/realtime_mixin.dart.twig
stnguyen90, line 85: Is it a problem if the subscription isn't removed from `_channels`? If a new subscription is created, the old subscription could also be used, resulting in an error like:

```
Error: Bad state: Cannot add new events after calling close
```

lohanidamodar: I reset channel once done, this should probably solve this.

stnguyen90 (approved):

PR #607, opened 2023-01-31
Flutter fix test for latest version

lohanidamodar:
## What does this PR do?

- Fixes Flutter test for recent Flutter version. Issue arose from recent change to Flutter (https://docs.flutter.dev/development/packages-and-plugins/plugins-in-tests)

## Test Plan

## Related PRs and Issues

YES

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,6 @@ env:
   - SDK=FlutterBeta
   - SDK=Go112
   - SDK=Go118
-#  - SDK=FlutterBeta
   - SDK=KotlinJava8
   - SDK=KotlinJava11
   - SDK=KotlinJava17
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,7 @@ env:
   - SDK=DartStable
   - SDK=Deno1171
   - SDK=FlutterStable
+  - SDK=FlutterBeta
   - SDK=Go112
   - SDK=Go118
 #  - SDK=FlutterBeta
--- a/templates/flutter/lib/src/client_io.dart.twig
+++ b/templates/flutter/lib/src/client_io.dart.twig
@@ -80,7 +80,7 @@ class ClientIO extends ClientBase with ClientMixin {
   @override
   String get endPoint => _endPoint;
 
-  Future<Directory> getCookiePath() async {
+  Future<Directory> _getCookiePath() async {
     final directory = await getApplicationDocumentsDirectory();
     final path = directory.path;
     final Directory dir = Directory('$path/cookies');
@@ -133,7 +133,7 @@ class ClientIO extends ClientBase with ClientMixin {
   Future init() async {
     if(_initProgress) return;
     _initProgress = true;
-    final Directory cookieDir = await getCookiePath();
+    final Directory cookieDir = await _getCookiePath();
     _cookieJar = PersistCookieJar(storage: FileStorage(cookieDir.path));
     _interceptors.add(CookieManager(_cookieJar));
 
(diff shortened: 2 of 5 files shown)


templates/flutter/lib/src/client_io.dart.twig
```diff
@@ -80,7 +80,7 @@ class ClientIO extends ClientBase with ClientMixin {
   @override
   String get endPoint => _endPoint;
 
-  Future<Directory> _getCookiePath() async {
+  Future<Directory> getCookiePath() async {
```
stnguyen90, line 83: do we really want to make this public?

.travis.yml
```diff
@@ -29,6 +29,7 @@ env:
   - SDK=DartStable
   - SDK=Deno1171
   - SDK=FlutterStable
+  - SDK=FlutterBeta
```
stnguyen90, line 32: We should probably delete the commented out flutter beta below.

tests/languages/flutter/tests.dart
```diff
@@ -1,10 +1,21 @@
 import 'package:flutter/material.dart';
 import '../lib/packageName.dart';
+import '../lib/client_io.dart';
 import '../lib/models.dart';
+import 'dart:io';
+
+class MockClient extends ClientIO {
+  Future<Directory> getCookiePath() async {
```
stnguyen90, line 8: Is this done because of path_provider? Perhaps we should be replacing the [PathProviderPlatform.instance](https://github.com/flutter/plugins/blob/8fcff8756c9888ac112afa12baacbb186b07f982/packages/path_provider/path_provider/test/path_provider_test.dart#LL24C52-L24C52) instead?

templates/flutter/lib/src/client_io.dart.twig
lohanidamodar: Well, due to recent change in Flutter, the plugins don't work in Flutter test, in order to run test successfully, I need to overwrite this method and can't do that without making it public

tests/languages/flutter/tests.dart
stnguyen90: Yes, but the suggestion was to create a mocked [PathProviderPlatform.instance](https://github.com/flutter/plugins/blob/8fcff8756c9888ac112afa12baacbb186b07f982/packages/path_provider/path_provider/test/path_provider_test.dart#LL24C52-L24C52).

templates/flutter/lib/src/client_io.dart.twig
stnguyen90: We can discuss this in the other [thread](https://github.com/appwrite/sdk-generator/pull/607/files#r1092632600).

.travis.yml
stnguyen90: The commented out line is still down below

stnguyen90 (changes requested):

tests/languages/flutter/tests.dart
lohanidamodar: Only mocking might not work, because we need actual directory, as we have cookie test, that requires actually saving the cookie in cache, Otherwise, I'll have to mock cookie jar and other cookie related stuff including the interceptors

stnguyen90: @lohanidamodar, sorry, I don't quite understand the problem. What's the problem with mocking [PathProviderPlatform.instance](https://github.com/flutter/plugins/blob/8fcff8756c9888ac112afa12baacbb186b07f982/packages/path_provider/path_provider/test/path_provider_test.dart#LL24C52-L24C52) and having `getApplicationDocumentsDirectory()` return `.`. Then, the following

```dart
    final directory = await getApplicationDocumentsDirectory();
    final path = directory.path;
    final Directory dir = Directory('$path/cookies');
    await dir.create();
    return dir;
```

would be the same as what you have here, right?

lohanidamodar: Updated!

stnguyen90 (approved):

PR #614, opened 2023-02-15
Feat: Add path keyword

Meldiron:
## What does this PR do?

We use `path` as variable name inside function. Such parameter cannot be provided, otherwise there will be an error like

```ts
src/services/functions.ts:583:17 - error TS2300: Duplicate identifier 'path'.

583             let path = '/functions/{functionId}/executions'.replace('{functionId}', functionId);
                    ~~~~
```

## Test Plan

- [x] QA with branch that needs path parameter

## Related PRs and Issues

x

Yes


src/SDK/Language/Dart.php
```diff
@@ -103,7 +103,8 @@ public function getKeywords(): array
             "yield",
             "required",
             "extension",
-            "late"
+            "late",
+            "path"
```
lohanidamodar, line 107: Not a keyword for Dart language

lohanidamodar (changes requested):
@Meldiron Please add it to only relevant languages and also add it to ` abstract public function getIdentifierOverrides(): array;` or the relevant language to give it a proper replacement like `urlPath` or similar.

Meldiron:
Converting to draft. Will become relevant again near internal Functions G4 testing

Meldiron:
CleanShot 2023-05-11 at 22 30 06@2x

Should be good for now 😅

lohanidamodar:
@Meldiron can you also add this to identifier overrides so that we can have a override sucha as `filePath` or something instead of the default `xpath` when the keyword arises?
https://github.com/appwrite/sdk-generator/blob/3bd342fbfa771d573fbf877af8ec977e5c7e9194/src/SDK/Language/Dart.php#LL115C8-L115C8

lohanidamodar (approved):
this is good to merge, my last comment as I discussed with matej, as path can refer to file path, code path or something else, there is no proper identifier override that we can manually do so `xpath` is okay for now

PR #632, opened 2023-03-23
Add new query operators

abnegate:
## What does this PR do?

Add query support for:
- isNull
- isNotNull
- between
- startsWith
- endsWith
- select

## Test Plan

Updated query tests

## Related PRs and Issues

Yes


templates/android/library/src/main/java/io/appwrite/Query.kt.twig
```diff
@@ -16,6 +16,22 @@ class Query {
...
 
+    fun isNull(attribute: String) = "isNull(\"${attribute}\")"
+
+    fun isNotNull(attribute: String) = "isNotNull(\"${attribute}\")"
+
+    fun between(attribute: String, start: Int, end: Int) = Query.addQuery(attribute, "between", listOf(start, end))
+
+    fun between(attribute: String, start: Double, end: Double) = Query.addQuery(attribute, "between", listOf(start, end))
+
+    fun between(attribute: String, start: String, end: String) = Query.addQuery(attribute, "between", listOf(start, end))
```
lohanidamodar, line 27: how does between with string work

lohanidamodar (approved):
LGTM

templates/android/library/src/main/java/io/appwrite/Query.kt.twig
abnegate: It uses the letters like a numerical value, e.g. B is between A and C. Similarly, Boat is between Apple and Cat. 

If multiple letters match it will check the following letter. E.g. Boat is between Bat and Cat.

Issue #652, opened 2023-05-01
🚀 Feature: Allow user-defined fetch function for web sdk


SilverMira:
### 🔖 Feature description

Currently, the web SDK is importing and using `fetch` from `cross-fetch`  inside `Client.call`, is it possible to introduce a `Client.setFetch(fetchFn)` for users to provide the implementation to use?

I would be happy to create a PR for this feature

### 🎤 Pitch

When using appwrite with SvelteKit and SSR, there's a need to use SvelteKit's special fetch function on page load. This special fetch function provides additional functionalities as described [here](https://kit.svelte.dev/docs/load#making-fetch-requests). One of them being ability of reusing what the server has fetched from appwrite instead of refetching them on the client during hydration.

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

joeyouss:
Hi
This is a cool feature, you could definitely try making a PR, but it would be to our SDK-generator, which can be a bit hard to do.

SilverMira:
Hi, I have created the PR, hope everything goes smoothly 🚀

PR #656, opened 2023-05-12
feat: setFetchFunction for web

SilverMira:
## What does this PR do?

Introduce feature on Web SDK to accept user-defined fetch function when performing request, detailed pitch at #652 

## Test Plan

I have run the sdk-generator and bundled the generated Web SDK, linked to an existing SvelteKit project as a local package, and was able to reuse fetched responses by the server when SSR is used

## Related PRs and Issues

* #652 

Yes

### DIFF
--- a/templates/web/src/client.ts.twig
+++ b/templates/web/src/client.ts.twig
@@ -94,6 +94,7 @@ class Client {
     config = {
         endpoint: '{{ spec.endpoint }}',
         endpointRealtime: '',
+        fetchFunction: fetch,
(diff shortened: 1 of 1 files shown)


SilverMira:
Hi, just dropping by to see if there are any updates, running Appwrite on Sveltekit SSR wouldn't even work properly for authenticated request during `load()` without this functionality, as the only way to forward auth cookies on the server would be using `handleFetch` which proxies Sveltekit's special `fetch` function

PR #670, opened 2023-06-15
Fix: fix type conversion of strings for all models in .NET SDK

adityaoberai:
## What does this PR do?

Corrects type conversion of string type by shifting from `(string)` type cast to using `.ToString()` method in the `From()` function in all the Models

Example:  

- Earlier (generated SDK)

image

- Now (generated SDK)

image

## Test Plan

- Run `php example.php` in the repo
- View the `From()` function in all Models in the generated .NET SDK 

## Related PRs and Issues

#320 

Yes


templates/dotnet/src/Appwrite/Models/Model.cs.twig
```diff
@@ -40,7 +40,7 @@ namespace {{ spec.title | caseUcfirst }}.Models
...
+            {{ property.name | caseCamel | escapeKeyword | removeDollarSign }}: {% if property.sub_schema %}{% if property.type == 'array' %}((JArray)map["{{ property.name }}"]).ToObject<List<Dictionary<string, object>>>().Select(it => {{property.sub_schema | caseUcfirst | overrideIdentifier}}.From(map: it)).ToList(){% else %}{{property.sub_schema | caseUcfirst | overrideIdentifier}}.From(map: ((JObject)map["{{ property.name }}"]).ToObject<Dictionary<string, object>>()!){% endif %}{% else %}{% if property.type == 'array' %}((JArray)map["{{ property.name }}"]).ToObject<{{ property | typeName }}>(){% else %}{% if property.type == "integer" or property.type == "number" %}{% if not property.required %}map["{{ property.name }}"] == null ? null : {% endif %}Convert.To{% if property.type == "integer" %}Int64{% else %}Double{% endif %}(map["{{ property.name }}"]){% else %}{% if property.type == "bool" %}({{ property | typeName }}{% if not property.required %}?{% endif %})map["{{ property.name }}"]{% endif %}map["{{ property.name }}"].ToString(){% endif %}{% endif %}{% endif %}{% if not loop.last or (loop.last and definition.additionalProperties) %},{% endif %}
```
abnegate, line 43: > map["{{ property.name }}"].ToString()

Let's add a check for if the property is required and use a null-conditional to call `ToString()`

adityaoberai: @abnegate made the change

abnegate: We can make the code a bit more compact if the output is `map["{{ property.name }}"]?.ToString()` instead of `map["{{ property.name }}"] == null ? null : map["{{ property.name }}"].ToString()` since they are technically equivalent

adityaoberai: Makes sense
I was trying to retain the same styling as earlier (for numbers)


But that's a fair point 💯

abnegate: Let's make it so the `?` is only present if the parameter is _not_ required. If the parameter is required, we should be able to safely call `ToString()` without as it will always be non-null (if it is null, it indicates an API issue). So we should have:
- `map["{{ property.name }}"].ToString()` for required parameters
- `map["{{ property.name }}"]?.ToString()` for optional parameters

adityaoberai: Alright 👍

adityaoberai: Made this change @abnegate

abnegate (approved):

PR #426, opened 2022-04-20
Flutter file examples apart from image

lohanidamodar:
```dart
// generated code
import 'package:appwrite/appwrite.dart';

void main() { // Init SDK
  Client client = Client();
  Storage storage = Storage(client);

  client
    .setEndpoint('https://[HOSTNAME_OR_IP]/v1') // Your API Endpoint
    .setProject('5df5acd0d48c2') // Your project ID
  ;
  // downloading file
  Future result = storage.getFileDownload(
    bucketId: '[BUCKET_ID]',
    fileId: '[FILE_ID]',
  ).then((bytes) {
    final file = File('path_to/file.pdf');
    file.writeAsBytesSync(bytes)
  }).catchError((error) {
      print(error.response);
  })
}

//displaying image preview
FutureBuilder(
  future: storage.getFileDownload(
    bucketId: '[BUCKET_ID]',
    fileId: '[FILE_ID]',
  ), //works for both public file and private file, for private files you need to be logged in
  builder: (context, snapshot) {
    return snapshot.hasData && snapshot.data != null
      ? Image.memory(
          snapshot.data,
        )
      : CircularProgressIndicator();
  },
);

```

PR #674, opened 2023-06-19
feat: Enum Whitelist

2002Bishwajeet:
## What does this PR do?

Generates enum for the given whitelist

## Test Plan

N/A

## Related PRs and Issues

Part of [#426](https://github.com/appwrite/appwrite/issues/426)

yess


src/SDK/Language.php
```diff
@@ -97,4 +97,25 @@ protected function toCamelCase($str): string
         $str = str_replace(" ", "", $str);
         return lcfirst($str);
     }
+
+    public function getEnumType(string $service, string $method): string
+    {
+        switch ($service) {
```
eldadfux, line 103: We should avoid Appwrite specific things. If we need namespaces they should either be auto-built or provided by Appwrite's spec.

src/SDK/SDK.php
```diff
@@ -634,6 +637,28 @@ public function generate(string $target): void
                         }
                     }
                     break;
+                case 'enum':
+                    foreach ($this->spec->getServices() as $key => $service) {
+                        $methods = $this->spec->getMethods($key);
+
```
eldadfux, line 643:
```suggestion
```

Extra line is not needed.

src/Spec/Swagger2.php
```diff
@@ -227,6 +227,7 @@ public function getMethods($service)
                                 $output['parameters']['header'][] = $param;
                                 break;
                             case 'path':
+                                $param['enum'] = $parameter['enum'] ?? null;
```
eldadfux, line 230: Is this done according to the spec? is our swagger2 file still valid?

templates/dart/lib/enums.dart.twig
```diff
@@ -0,0 +1,5 @@
+library {{ language.params.packageName }}.enums;
+
+{% for enum in spec.enum %}
```
eldadfux, line 3: Let's make sure indentation on all templates work properly.

src/Spec/Swagger2.php
christyjacob4: Yes it is 👍  There are a few warnings, we will address those

src/SDK/SDK.php
```diff
@@ -634,6 +637,27 @@ public function generate(string $target): void
...
+                        foreach ($methods as $method) {
+                            $params['method'] = $method;
+                            $parameters = $method['parameters']['path'];
+
+                            foreach ($parameters as $parameter) {
+                                // Check if the enum field is defined
+                                if ($parameter['enum'] != null) {
+                                    $params['enum'] = [
+                                        'name' =>  $parameter['enum-name'],
```
christyjacob4, line 652: Isn't this supposed to be `x-enum-name` ? @2002Bishwajeet

2002Bishwajeet: I did this mapping [here](https://github.com/appwrite/sdk-generator/blob/fde85780a860a278e7db3b17b76827b0f7e2863c/src/Spec/Swagger2.php#L231) so hence the `enum-name`

src/SDK/Language.php
2002Bishwajeet: Its been removed and shifted to appwrite spec file (https://github.com/appwrite/appwrite/pull/5726)

templates/dart/lib/enums.dart.twig
lohanidamodar, line 1: enums doesn't require separate export we should add this to `package.dart.twig` for both Dart and Flutter SDK

lohanidamodar (changes requested):

templates/dart/lib/enums.dart.twig
2002Bishwajeet: Done

2002Bishwajeet (commented):
Looks good just one minor change

src/SDK/Language/DotNet.php
```diff
@@ -394,6 +394,11 @@ public function getFiles(): array
                 'scope'         => 'definition',
                 'destination'   => '/src/{{ spec.title | caseUcfirst }}/Models/{{ definition.name | caseUcfirst | overrideIdentifier }}.cs',
                 'template'      => 'dotnet/src/Appwrite/Models/Model.cs.twig',
+            ],
+            [
+                'scope'         => 'enum',
+                'destination'   => '/src/{{ spec.title | caseUcfirst }}/Enums/{{ definition.name | caseUcfirst | overrideIdentifier }}.cs',
```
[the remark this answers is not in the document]
adityaoberai: I forgot to save this change before committing 🤦
Committing the suggestion from here itself

templates/dart/lib/package.dart.twig
```diff
@@ -27,3 +27,7 @@ part 'id.dart';
 {% for service in spec.services %}
 part 'services/{{service.name | caseDash}}.dart';
 {% endfor %}
+
```
lohanidamodar, line 30:
```suggestion
```

lohanidamodar (approved):
Dart and Flutter now look good. Would love to see the generated outputs.

2002Bishwajeet:
@lohanidamodar , Created a web sdk repository . Each branch has different SDK platform(client,console,server)
https://github.com/2002Bishwajeet/sdk-for-web/ . You can expect the similar pattern for all other languages as well😉

templates/android/library/src/main/java/io/appwrite/enums/Enum.kt.twig
```diff
@@ -0,0 +1,10 @@
+package {{ sdk.namespace | caseDot }}.enums
+
+enum class {{ enum.name | caseUcfirst | overrideIdentifier }}(val value: String) {
+{% for value in enum.enum %}
+    {% set key = enum.keys is empty ? value : enum.keys[loop.index0] %}
+    {{ key | caseUpper | replace({'-': '',' ':'_'})}}('{{value}}'){% if not loop.last %},
```
abnegate, line 6: > '{{value}}'

We need to use double quotes instead of single for Kotlin

templates/dotnet/src/Appwrite/Enums/Enums.cs.twig
```diff
@@ -0,0 +1,12 @@
+using System;
+
+namespace {{ spec.title | caseUcfirst }}.Enums
+{
+    public static class {{ enum.name | caseUcfirst | overrideIdentifier }}
+    {
+        {% for value in enum.enum %}
+            {% set key = enum.keys is empty ? value : enum.keys[loop.index0] %}
+        public static string {{ key | caseUcfirst | replace({'-': ''}) }} = "{{ value }}",
```
abnegate, line 9: Let's make this a readonly property and skip the last separator:

```suggestion
        public static string {{ key | caseUcfirst | replace({'-': ''}) }} => "{{ value }}"{% if not loop.last %},{% endif %}
```

templates/android/library/src/main/java/io/appwrite/enums/Enum.kt.twig
abnegate, line 6: I think we should strip spaces instead of replacing with an underscore

templates/dart/docs/example.md.twig
```diff
@@ -22,7 +22,7 @@ void main() { // Init SDK
 
 {% endif %}
 {% if parameter.required %}
-    {{ parameter.name | caseCamel | overrideIdentifier }}: {{ parameter | paramExample | replace({'<field-name>': (parameter.name | caseCamel) }) | raw }},
+    {{ parameter.name | caseCamel | overrideIdentifier }}:{% if parameter.enumValues | length > 0%} {{parameter.enumName}}.{{(parameter.enumKeys[0] ?? parameter.enumValues[0]) | caseCamel | replace({'-': ''}) }}.value{% else%}{{ parameter | paramExample | replace({'<field-name>': (parameter.name | caseCamel) }) | raw }} {% endif %},
```
abnegate, line 25: Let's make a global twig filter `stripEnum` that will handle the replacements for us so we don't need to remember them for every template

templates/kotlin/src/main/kotlin/io/appwrite/enums/Enum.kt.twig
```diff
@@ -0,0 +1,10 @@
+package {{ sdk.namespace | caseDot }}.enums
+
+enum class {{ enum.name | caseUcfirst | overrideIdentifier }}(val value: String) {
+{% for value in enum.enum %}
+    {% set key = enum.keys is empty ? value : enum.keys[loop.index0] %}
+    {{ key | caseUpper | replace({'-': '',' ': '_'})}}('{{value}}'){% if not loop.last %},
```
abnegate, line 6: We need the double quotes instead of single here too

templates/swift/Sources/Enums/Enum.swift.twig
```diff
@@ -0,0 +1,5 @@
+enum {{ enum.name | caseUcfirst | overrideIdentifier }} : String {
+{% for value in enum.enum %}
+  {% set key = enum.keys is empty ? value : enum.keys[loop.index0] %}
+  case {{ key | caseCamel | replace({'-': ''})}} = '{{ value }}'
```
abnegate, line 4: Since we defined the raw type as `String`, we can remove the ` = '{{ value }}'` as the value can be inferred by the name

abnegate, line 4: We need to add a comma after every key except the last

src/SDK/Language/Swift.php
```diff
@@ -284,6 +284,11 @@ public function getFiles(): array
                 'destination'   => '/Sources/{{ spec.title | caseUcfirst}}Models/{{ definition.name | caseUcfirst }}.swift',
                 'template'      => '/swift/Sources/Models/Model.swift.twig',
             ],
+            [
+                'scope' => 'enum',
+                'destination' => '/Sources/{{ spec.title | caseUcfirst}}/Enums/{{ enum.name | caseUcfirst }}.swift',
```
abnegate, line 289: We should make this into another module, so the path would be:
```php
                'destination' => '/Sources/{{ spec.title | caseUcfirst}}Enums/{{ enum.name | caseUcfirst }}.swift',
```
Additionally, the Package.swift would need to be updated to declare the new target and add it as a dependency to the main Appwrite module. The service class would need to `import AppwriteEnums`

abnegate (commented):
Templates look good, we still need to update the services to accept the new enum types as parameters. E.g. [here](https://github.com/2002Bishwajeet/sdk-for-web/blob/7d7f20ce75740dfd677bf39d2cc53a5981efaf9a/src/services/account.ts#L655) the `provider` param should be typed `Provider` instead of `string`

templates/swift/Sources/Enums/Enum.swift.twig
2002Bishwajeet: It's necessary cause the `key`  and `value` have different casing. key has `camelCase` while the value could be lowercase, snakeCase , kebabCase etc

2002Bishwajeet: I couldn't see the format here about adding a comma (https://docs.swift.org/swift-book/documentation/the-swift-programming-language/enumerations/#Raw-Values) . Do you still want me to add it?

abnegate: Ah yes true :+1:

abnegate: No need, my mistake

abnegate:
@2002Bishwajeet We still need to change the service function parameters types to the enum type where applicable

E.g. [Here](https://github.com/2002Bishwajeet/sdk-for-web/blob/7d7f20ce75740dfd677bf39d2cc53a5981efaf9a/src/services/account.ts#L655) the provider param should be typed `Provider` instead of `string`

templates/dotnet/src/Appwrite/Enums/Enums.cs.twig
```diff
@@ -0,0 +1,13 @@
+using System;
+
+namespace {{ spec.title | caseUcfirst }}.Enums
+{
+    public static class {{ enum.name | caseUcfirst | overrideIdentifier }}
+    {
+        {% for value in enum.enum %}
+            {% set key = enum.keys is empty ? value : enum.keys[loop.index0] %}
+        public static readonly string {{ key | caseUcfirst | replace({'-': ''}) }} => "{{ value }}"
+            {% if not loop.last %},{% endif %}
```
adityaoberai, line 10:
```suggestion
            {% if not loop.last %}{% endif %}
```
Replaced the `,` with a `;` above, since these are [properties](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/properties)

adityaoberai (changes requested):
@2002Bishwajeet add a couple of comments

templates/dotnet/src/Appwrite/Enums/Enums.cs.twig
```diff
@@ -0,0 +1,13 @@
+using System;
+
+namespace {{ spec.title | caseUcfirst }}.Enums
+{
+    public static class {{ enum.name | caseUcfirst | overrideIdentifier }}
+    {
+        {% for value in enum.enum %}
+            {% set key = enum.keys is empty ? value : enum.keys[loop.index0] %}
+        public static string {{ key | caseUcfirst | replace({'-': ''}) }} => "{{ value }}";
+            {% if not loop.last %}{% endif %}
```
adityaoberai, line 10: Is this necessary here @2002Bishwajeet ?

adityaoberai (changes requested):
@2002Bishwajeet one more comment

templates/dotnet/src/Appwrite/Enums/Enums.cs.twig
2002Bishwajeet: Not Anymore

adityaoberai (approved):

abnegate (approved):

PR #681, opened 2023-06-22
Feat cli console project

stnguyen90:
## What does this PR do?

1. Specify the console project when checking endpoint
2. Test the CLI against Appwrite instead of homepage

## Test Plan

Manual

## Related PRs and Issues

* https://github.com/appwrite/appwrite/pull/5639

Yes


templates/cli/lib/commands/generic.js.twig
```diff
@@ -82,6 +82,7 @@ const client = new Command("client")
                 }
 
                 let client = new Client().setEndpoint(endpoint);
+                client.setProject('console');
```
lohanidamodar, line 85: Why do we need this here though?

tests/languages/cli/test.js
```diff
@@ -1,6 +1,6 @@
 const { exec, execSync } = require('child_process');
 
-execSync("node index client --endpoint 'https://stage.appwrite.io/v1' --projectId console --key=35y3h5h345 --selfSigned true", { stdio: 'inherit' });
+execSync("node index client --endpoint 'https://demo.appwrite.io/v1' --projectId console --key=35y3h5h345 --selfSigned true", { stdio: 'inherit' });
```
lohanidamodar, line 3: We need to decide on which instance we want to use for test, also that hast to have latest features even before release. We might want to maintain a separate instance I'm not sure yet but we need a better strategy.

stnguyen90: It should be suffice to use any Appwrite instance since this only tests the `/v1/health/version` endpoint. Anyways, it's better to get this working than to have it broken.

lohanidamodar (approved):

PR #684, opened 2023-06-23
Create a build script

stnguyen90:
## What does this PR do?

The [example.php](https://github.com/appwrite/sdk-generator/blob/66ae91ffd298bce9eb534d956879e58b7cce64d5/example.php) we provide requires fiddling to select which SDK to build. This PR introduces a build script where you can pass arguments for which SDKs you want to build:

## Test Plan

```shell

# install dependencies
$ docker run --rm --interactive --tty \
  --volume $PWD:/app \
  composer update --ignore-platform-reqs --optimize-autoloader --no-plugins --no-scripts --prefer-dist
  
# check usage
$ docker run --rm -v $(pwd):/app -w /app php:8.1-cli php build.php
Usage: php example.php <platform> [<sdk1> <sdk2> ...]

  <platform>  The platform you want to generate the SDK for.
              Supported platforms: client, console, server

  <sdk>       The SDK you want to generate. Supported SDKs:
              android, apple, flutter, web, cli, dart, deno, dotnet, kotlin, node, php, python, ruby, swift
              If no SDKs are provided, all SDKs for the specified platform will be generated.

# build dart and dotnet server SDKs
$ docker run --rm -v $(pwd):/app -w /app php:8.1-cli php build.php server dotnet dart
Example SDKs generated successfully

$ ls -la examples
total 0
drwxr-xr-x   4 steven  staff   128 Jun 23 14:41 .
drwxr-xr-x  33 steven  staff  1056 Jun 23 14:41 ..
drwxr-xr-x  13 steven  staff   416 Jun 23 14:41 dart
drwxr-xr-x   9 steven  staff   288 Jun 23 14:41 dotnet
```

## Related PRs and Issues

None

None

### DIFF
--- a/build.php
+++ b/build.php
@@ -1,0 +1,599 @@
+<?php
+
+include_once 'vendor/autoload.php';
+
+use Appwrite\SDK\Language\GraphQL;
+use Appwrite\Spec\Swagger2;
+use Appwrite\SDK\SDK;
+use Appwrite\SDK\Language\Web;
+use Appwrite\SDK\Language\Node;
+use Appwrite\SDK\Language\CLI;
+use Appwrite\SDK\Language\PHP;
+use Appwrite\SDK\Language\Python;
+use Appwrite\SDK\Language\Ruby;
+use Appwrite\SDK\Language\Dart;
+use Appwrite\SDK\Language\Go;
+use Appwrite\SDK\Language\Deno;
+use Appwrite\SDK\Language\REST;
+use Appwrite\SDK\Language\Swift;
+use Appwrite\SDK\Language\Apple;
+use Appwrite\SDK\Language\DotNet;
+use Appwrite\SDK\Language\Flutter;
+use Appwrite\SDK\Language\Android;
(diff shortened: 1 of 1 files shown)


stnguyen90:
@lohanidamodar, what are your thoughts on replacing our example.php with this?

lohanidamodar:
> @lohanidamodar, what are your thoughts on replacing our example.php with this?

I'm okay, if by default, when we don't provide any arguments it generates all possible SDKs like example.php. Can we make that happen?

lohanidamodar:
> > I'm okay, if by default, when we don't provide any arguments it generates all possible SDKs like example.php. Can we make that happen?
> 
> @lohanidamodar, the example.php we currently have makes all sdks, but does it make sense to do that because you need to use the right specs for each sdk, right (server specs for node, console specs for CLI, etc)?

It makes sense to quickly to check all the SDKs, in build.php we could do is setup in a way that appropriate spec is used to generate all the SDKs.

lohanidamodar:
@stnguyen90 let's sync with latest test changes

PR #713, opened 2023-09-07
fix for realtime calling subscription multiple time - Appwrite 1.4.x

lohanidamodar:
## What does this PR do?

- Fixes issue with calling subscribe multiple times
- Not syncing with main so this remains, if we want to release Flutter SDK with this fix for Appwrite 1.4.x
- For latest PR check: https://github.com/appwrite/sdk-generator/pull/801
- We shouldn't merge this, rather use this if we want to release (as main has few new features I think not suitable for generating with spec for 1.4.x) if not we can close this issue

## Test Plan

- Manually testing the generated SDK

## Related PRs and Issues

- https://github.com/appwrite/sdk-for-flutter/issues/156

YES


templates/flutter/lib/src/realtime_mixin.dart.twig
```diff
@@ -118,43 +122,49 @@ mixin RealtimeMixin {
...
-          if(_channels.isNotEmpty) {
+          _cleanup(channels);
+
+          if (_channels.isNotEmpty) {
             await Future.delayed(Duration.zero, () => _createSocket());
           } else {
             await _closeConnection();
           }
         });
+    _subscriptions[counter] = subscription;
```
stnguyen90, line 151: I think using a counter can lead to problems. For example:

0. start (counter = 0, subscriptions map is empty)
1. subscribe() (counter = 1, subscriptions[0] is set)
2. subscribe() (counter = 2, subscriptions[0] and [1] are set)
3. the first one gets removed (counter = 1, subscriptions[1] is set)
4. a new subscription added (counter = 2, subscriptions[1] is set and replaces the subscription from step 2)

lohanidamodar: Thanks, I've updated to use the time in microseconds to overcome this. please let me know what you think.

stnguyen90:
Closing as this is no longer needed. It has been added in later versions.

PR #717, opened 2023-09-11
Fix apple xpc interrupt

abnegate:
## What does this PR do?

- Fix XPC connection interrupted when app in background during OAuth login for Apple

## Test Plan

Manual test

## Related PRs and Issues

https://github.com/appwrite/appwrite/issues/6206

Yes

### DIFF
--- a/templates/swift/Sources/Services/Service.swift.twig
+++ b/templates/swift/Sources/Services/Service.swift.twig
@@ -35,10 +35,10 @@ open class {{ service.name | caseUcfirst }}: Service {
         {%~ if 'multipart/form-data' in method.consumes %}
         onProgress: ((UploadProgress) -> Void)? = nil
         {%~ endif %}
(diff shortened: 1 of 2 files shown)


ItzNotABug:
I don't think this would fix the:\
`Accessing Environment<OpenURLAction>'s value outside of being installed on a View. This will always read the default value and will not update` warning.

It may not be something major, its just pretty annoying.

abnegate:
> I don't think this would fix the: `Accessing Environment<OpenURLAction>'s value outside of being installed on a View. This will always read the default value and will not update` warning.
> 
> It may not be something major, its just pretty annoying.

@ItzNotABug Correct, it won't fix that warning - we'll address that one separately in future :+1:

stnguyen90 (approved):
I don't know Swift and I didn't generate the SDK to confirm things, but the template syntax looks okay.

Issue #680, opened 2023-06-22
🚀 Feature: Unit Tests


stnguyen90:
### 🔖 Feature description

Add unit tests for each SDK.

### 🎤 Pitch

asdf

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

tewarig:
can i work on this ? would love to add test for web

lohanidamodar:
@tewarig sure you can get started.

prathamesh3707:
i would like to work

leonyangela:
Hi @Haimantika, can I work on node?? Thank you

Haimantika:
> i would like to work

Hi @prathamesh3707 have assigned the issue to you! Thank you for showing interest in contributing to Appwrite! Happy Hacktoberfest 🎃

Notes:

Please update us with your progress every 3 days, so that we know that you are working on it.
Join us on Discord - https://appwrite.io/discord to chat about Hacktoberfest and Appwrite!

Haimantika:
> Hi @Haimantika, can I work on node?? Thank you

Hi, we are assigning issues on first-come first-serve basis. If @prathamesh3707 drops/is unresponsive, will assign it to you.

Haimantika:
Hi @prathamesh3707 any update? This issue will soon be reassigned to the next person due to inactivity

prathamesh3707:
sorry, i saw it just now. Thanks for assigning it to me! I am working on it

Haimantika:
> sorry, i saw it just now. Thanks for assigning it to me! I am working on it

Hi, let us know in [Discord](https://appwrite.io/discord) if you have any questions :)

prathamesh3707:
Ya sure , I need a little help in getting started , I'll ask there. Is there any deadline for this?

geisterfurz007:
Heyyo! I would like to pick up either dotnet or node if that was possible. Probably dotnet because node might be more popular for others to work on but I would leave that to your judgement. Let me know and I'll get started soon :)

airwakz:
@Haimantika can i work on this

ujju200:
Can I work on this issue?

Haimantika:
@leonyangela you are up next, would you like to work on it?

[...]

PR #735, opened 2023-10-31
Unit Tests - Deno

geisterfurz007:
## What does this PR do?

This PR adds unit tests for the Deno SDK.

## Test Plan

Generate the SDK, then run `deno test` in the generated SDK's directory.

## Related PRs and Issues
#680 

### Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?

Yup

PR #725, opened 2023-10-11
feat: ssr

loks0n:
## What does this PR do?

- Replaces node dependencies `form-data` and `axios` with `undici`.
- Drops support for Node 14 and adds Node 20

### Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?

Yes

PR #736, opened 2023-10-31
Unit Tests - Node

geisterfurz007:
## What does this PR do?

This PR adds generated unit tests to the Node SDK.

## Test Plan

Generate the SDK, then run run `yarn install` and `yarn jest` (or the equivalents for your favourite package manager) in the directory of the generated SDK.

## Related PRs and Issues

#680 

Yup


lohanidamodar (commented):
From brief review, its looking good. Would test with generation later.

templates/node/package.json.twig
```diff
@@ -11,7 +11,8 @@
     "url": "{{ sdk.gitURL }}"
   },
   "devDependencies": {
-    "@types/node": "^18.16.1"
+    "@types/node": "^18.16.1",
+    "jest": "^29.7.0"
```
loks0n, line 15: Not a blocker but just noting Node actually has built in test runner for a while now. We could use it in the future

templates/node/test/services/service.test.js.twig
```diff
@@ -0,0 +1,38 @@
+const Client = require("../../lib/client");
+const InputFile = require("../../lib/inputFile");
+const {{ service.name | caseUcfirst }} = require("../../lib/services/{{ service.name | caseCamel }}");
+
+const mockedAxios = require("axios");
+jest.mock('axios', () => jest.fn());
```
loks0n, line 6: We are migrating to use `import {fetch} from 'undici'` as part of SSR support changes. (see https://github.com/appwrite/sdk-generator/pull/725)

We will need to mock that instead

loks0n (changes requested):

templates/node/package.json.twig
geisterfurz007: Yeah, I saw that mentioned somewhere the other day but didn't look into it further; experimental in Node 18, stable since 20; I could look into migrating this if needed!

Point of concern, since I am not too familiar with distributing libraries like this: Wouldn't tests in the long term not also be run on older Node versions than 18 to ensure they are compatible with those versions? Or are there plans to drop anything below 18 anyway since 16 LTS has reached EOL earlier this year?

loks0n: The tests only have to work on CI and developer environment so that's fine. 
The generated SDK needs to support 16 and above. We're a little bit behind node EOL.

If you are interested to look into it that'd be awesome, but it's not a blocker :)

geisterfurz007: 'ello! I took a peek and for the most part, just dropping jest and adding `const { describe, test } = require("node:test");` should be sufficient. I did stumble over one issue though and that is mocking of modules. Specifically, there doesn't seem to be a sane way of mocking axios in this instance. After some research, I stumbled over https://github.com/nodejs/node/issues/51164 with still ongoing discussion on whether / how module level mocking is supported at the moment.

Given that there doesn't seem to be a documented example on how to properly achieve this, I think I'd rather hold off on this for now but I am down to possibly revisit this in the future!

loks0n: Makes sense, sounds like a good plan

abnegate:
Are you make these test run automatically with the existing tests?

geisterfurz007:
@abnegate No, and I wasn't aware that was part of the task, sorry! Same answer goes for the other PRs from me you commented on. Since there currently seems to be a fairly central point of testing which is broken up by these PRs, it might make sense to overhaul it in one go at some point? Two of the Unit Tests PRs ([PHP](https://github.com/appwrite/sdk-generator/pull/737) and [Deno](https://github.com/appwrite/sdk-generator/pull/735)) are also affected.

gewenyu99:
Hey there! There were a lot of big PRs during this Hacktoberfest, and we wanted to give everyone ample time to collaborate with our engineering team. If you were able to merge your PRs during October, amazing. If it’s still not merged, don’t worry about it either. Either way, we’ve got your Hacktoberfest swag minted and ready to ship. 

Please comment with your Discord username here so we can contact you about your shipping information to deliver your Hacktoberfest swag.

lohanidamodar:
@abnegate these tests are to be run on `appwrite/sdk-for-node` repository instead.

stnguyen90:
> these tests are to be run on `appwrite/sdk-for-node` repository instead.

@lohanidamodar, for Flutter and Dart, we run them in this repo. I think it makes sense to run them in this repo to catch any problems early before pushing to the `appwrite/sdk-for-node` repo 🤔

mvarendorff:
@ChiragAgg5k I _think_ it should be doable to get these polished up again - frankly, I am lacking the resources to bring everything to the most up-to-date state (e.g. include potentially new / more in-depth test-cases) but I am down to put some best-effort into these.

This PR I have mostly gotten ready, getting all tests fixed, I am going to throw together a quick operator.test.ts in the next couple days and then look into the other SDKs (Android/Kotlin and Python are both still on the list - .NET is superseeded by a more recent PR, I closed mine already).

Since that had been a comment in the past: I am not sure entirely how to set up automatic testing. In case that's required in this PR, I'd appreciate some guidance for that!

And another thing before I sink too much time into these: Do you think the PRs would be merged in the weeks after it's done? My schedule is going to be very tight later this year to the point where I cannot support the PRs anymore and I don't have a rough idea when I'd be free again and it would be unlucky for the time to go wasted.

ChiragAgg5k (approved):

mvarendorff:
@ChiragAgg5k alright, did that. However the last point of my previous post still stands:

There are already scripts for all SDKs in their respective languages that print all the function calls' results to stdout which is then matched against an expected list of strings (like `tests/languages/node/test.js` which calls a bunch of functions on client, Query and others) which appears to be run from `tests/Node*Test.php` in the `tests.yml` workflow.

There are two things that confuse me about this:

1) There are now two places where tests are written (the SDK templates themselves on one hand and a PHP file + language specific script on the other) and two workflows where they are executed (sdk-build-validation.yml and tests.yml respectively). Is this really intended or am I adding to potential tech debt here?

2) I noticed this since I was using the Dart SDK as my baseline: The Dart and Flutter SDK are not tested in the same workflow that I now added the Jest run to. How comes? This adds onto the feeling that the generated tests are largely handled as an afterthought, possibly *because* those other tests already exist and are more present?

ChiragAgg5k:
> There are now two places where tests are written (the SDK templates themselves on one hand and a PHP file + language specific script on the other) and two workflows where they are executed (sdk-build-validation.yml and tests.yml respectively). Is this really intended or am I adding to potential tech debt here?

sdk-generator is meant to be a generic generator that takes in any swagger spec and converts it into a sdk. i admit we are not doing very good on that front, but the idea remains. so the tests which print to stdout use a generic test spec for generating the tests and try to print out consistent outputs for testing.

in sdk build validation file we are doing appwrite specific sdk generation and running unit tests for those services specifically, giving us more confidence for out sdks

> I noticed this since I was using the Dart SDK as my baseline: The Dart and Flutter SDK are not tested in the same workflow that I now added the Jest run to. How comes? This adds onto the feeling that the generated tests are largely handled as an afterthought, possibly because those other tests already exist and are more present?

you can feel that way, and if you have ideas for improvements there feel free to create separate issue to solve them

PR #743, opened 2023-11-19
Unit Tests - Dotnet

geisterfurz007:
## What does this PR do?

This PR adds generated unit tests to the Dotnet SDK.

It also contains a deserialization fix that caused a `KeyNotFoundException` when a key was missing from the response map (I am not sure what the behaviour from the server is, whether the value is returned as null or doesn't exist in the map). The error came up when testing the User-related services that failed deserialization because the `password` field (amongst others) was not present in the mocked response.

## Test Plan

Generate the SDK, then run `dotnet test` in the generated SDK's `src` directory.

## Related PRs and Issues

#680

Yup

### DIFF
--- a/src/SDK/Language/DotNet.php
+++ b/src/SDK/Language/DotNet.php
@@ -333,6 +333,11 @@ class DotNet extends Language
             ],
             [
                 'scope'         => 'default',
+                'destination'   => '/src/{{ spec.title | caseUcfirst }}.Tests/{{ spec.title | caseUcfirst }}.Tests.csproj',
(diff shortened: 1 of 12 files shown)


geisterfurz007:
I have a question on this particular one. Some SDKs contain tests for serializing using the respective Model classes. I did work on those in here as well but faced a fairly annoying problem: The output of `ToMap` fails when fed into `From` due to the casting to JObject / JArray. Thus the question: How should I address that? I'd rather not meddle with the (de-)serialization too much if possible but I am not sure if leaving the tests out is a good idea either.

PR #763, opened 2024-01-29
Feat android messaging

abnegate:
## What does this PR do?

Add Android push notification handling

## Test Plan

## Related PRs and Issues

(Write your answer here.)


src/SDK/Language/Android.php
```diff
@@ -318,79 +348,13 @@ public function getFiles(): array
             ],
             [
                 'scope'         => 'copy',
-                'destination'   => '/example/.gitignore',
```
lohanidamodar, line 321: are we removing these example files?

templates/android/build.gradle.twig
```diff
@@ -2,17 +2,20 @@ apply plugin: 'io.github.gradle-nexus.publish-plugin'
...
         google()
         mavenCentral()
     }
     dependencies {
-        classpath "com.android.tools.build:gradle:4.2.2"
+        classpath "com.android.tools.build:gradle:8.2.1"
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
-        classpath 'io.github.gradle-nexus:publish-plugin:1.1.0'
+        classpath 'io.github.gradle-nexus:publish-plugin:1.3.0'
+        classpath 'com.google.gms:google-services:4.4.0'
```
lohanidamodar, line 18: I'm not comfortable with adding this by default, isn't there a way to avoid this and allow user to add it in their app build.gradle file instead of adding in the SDK? Doing this will complicate Flutter as we will have to add and maintain Firebase library for all Flutter platforms

src/SDK/Language/Android.php
abnegate: Just for the Java specific example, they are only used for dev testing while adding new features, and I never use the Java one

templates/android/build.gradle.twig
abnegate: Yep removed this, it's redundant as users have to add it to their gradle build anyway. We will still need to add the Firebase BOM + messaging library as a dependency to the SDK though

PR #766, opened 2024-02-01
Feat json queries

loks0n:


src/SDK/Language/Web.php
```diff
@@ -185,6 +185,9 @@ public function getTypeName(array $parameter, array $method = []): string
         if (!empty($parameter['enumValues'])) {
             return \ucfirst($parameter['name']);
         }
+        if (!empty($parameter['name']) && $parameter['name'] === 'queries') {
+            return 'Query[]';
+        }
```
abnegate, line 190: I think we might need something like this for Android and Flutter too @lohanidamodar, is there a way to do it without coupling to the parameter name?

templates/php/src/Query.php.twig
```diff
@@ -2,112 +2,145 @@
...
+
+    public function jsonSerialize()
+    {
+        return array_filter([
+            'method' => $this->method,
+            'attribute' => $this->attribute,
+            'values' => $this->values,
+        ], function ($value) {
+            return !is_null($value);
+        });
```
abnegate, line 37: You can omit the callback here, the default is to remove null values

templates/ruby/lib/container/query.rb.twig
```diff
@@ -1,94 +1,127 @@
+require 'json'
+
 module {{spec.title | caseUcfirst}}
     class Query
+        def initialize(method, attribute = nil, values = nil)
+            @method = method
+            @attribute = attribute
+
+            if (values != nil)
```
abnegate, line 9: Let's omit braces for if conditions

templates/ruby/lib/container/query.rb.twig
```diff
@@ -1,94 +1,127 @@
...
+
+            if (@attribute != nil)
+                query["attribute"] = @attribute
+            end
+
+            if (@values != nil)
+                query["values"] = @values
+            end
+
+            return query.to_json(*args)
```
abnegate, line 31:
```suggestion
            {
                method: @method,
                attribute: @attribute,
                values: @values
            }.compact.to_json(*args)
```

templates/dotnet/src/Appwrite/Query.cs.twig
```diff
@@ -1,139 +1,145 @@
...
 
 namespace Appwrite
 {
-    public static class Query
+    public class Query
     {
-        public static string Equal(string attribute, object value)
+        public string method;
+        public string? attribute;
+        public object? values;
```
abnegate, line 13: We can use `List<object?>` here

templates/swift/Sources/Query.swift.twig
```diff
@@ -1,104 +1,272 @@
...
+
     }
 
-    public static func isNull(_ attribute: String) -> String {
-        "isNull(\"\(attribute)\")"
+    public static func equal(_ attribute: String, value: Any) -> Query {
+        return Query(
+            method: "equal",
+            attribute: attribute,
+            values: [value]
```
abnegate, line 96: We need to make sure `value` isn't already an array for methods that accept `Any`

templates/web/src/client.ts.twig
```diff
@@ -359,6 +359,9 @@ class Client {
         } else {
             switch (headers['content-type']) {
                 case 'application/json':
+                    if ('queries' in params && Array.isArray(params['queries'])) {
+                        params['queries'] = params['queries'].map((query) => query.toString());
+                    }
```
abnegate, line 364: We need to do this in the GET case, routes with queries will always be GET

templates/deno/src/client.ts.twig
```diff
@@ -85,6 +85,9 @@ export class Client {
 
             body = formData;
         } else {
+            if ('queries' in params && Array.isArray(params['queries'])) {
+                params['queries'] = params['queries'].map((query) => query.toString());
+            }
```
abnegate, line 90: We need to do this in the GET case, routes with queries will always be GET

src/SDK/Language/Web.php
loks0n: We need to put something else in the specs. This was my workaround for now

templates/ruby/lib/container/query.rb.twig
loks0n: done

templates/php/src/Query.php.twig
loks0n: `json_encode` does not remove the null values. Are you referring to something else? I think it's safer to keep this

templates/ruby/lib/container/query.rb.twig
loks0n: done

templates/web/src/client.ts.twig
loks0n: Done inside flatten() function instead.

templates/deno/src/client.ts.twig
loks0n: Done inside flatten() function instead.

templates/dotnet/src/Appwrite/Query.cs.twig
loks0n: done

templates/php/src/Query.php.twig
abnegate: From the `array_filter`, we can just have it like:

```suggestion
        ]);
```

abnegate (approved):

PR #768, opened 2024-02-05
feat: oauth return location header

loks0n:
We want to allow server-side SDKs to use `webAuth` type endpoints such as `createOAuth2Session` and `createOAuth2Token`. 

These endpoints should return the string value of the location header on server SDKs.

for example:
```js
const redirectUrl = await account.createOAuth2Token("github", successUrl, failureUrl);`

console.log(redirectUrl) // outputs: https://github.com/oauth2?query=asdkfjhasdkjfhasdhkf?....
```

- [x] Node
- [x] Deno
- [x] Kotlin
- [x] Dart
- [x] DotNet
- [x] PHP
- [x] Python
- [x] Ruby
- [x] Swift


src/SDK/Language/Kotlin.php
```diff
@@ -447,7 +447,7 @@ public function getFilters(): array
     protected function getReturnType(array $method, array $spec, string $namespace, string $generic = 'T'): string
     {
         if ($method['type'] === 'webAuth') {
-            return 'Bool';
+            return 'String';
```
abnegate, line 450: We'll need to override this function and still return `Bool` for Android

src/SDK/Language/Swift.php
```diff
@@ -480,7 +480,7 @@ public function getFilters(): array
     protected function getReturnType(array $method, array $spec, string $generic): string
     {
         if ($method['type'] === 'webAuth') {
-            return 'Bool';
+            return 'String';
```
abnegate, line 483: We'll need to override this function and still return `Bool` for Apple

templates/swift/Sources/Services/Service.swift.twig
```diff
@@ -22,9 +22,6 @@ open class {{ service.name | caseUcfirst | overrideIdentifier }}: Service {
     /// @throws Exception
     /// @return array
     ///
-    {%~ if method.type == "webAuth" %}
-    @available(iOS 14.0, macOS 11.0, tvOS 14.0, watchOS 7.0, *)
-    {%~ endif %}
```
abnegate, line 27: We'll need to duplicate this file into the apple templates and keep this in the duplicate

templates/swift/base/requests/oauth.twig
abnegate, line 1: We'll need to duplicate this file into the apple templates and keep this in the duplicate

PR #774, opened 2024-02-15
fix: exports enums js

loks0n:
- Adds export of enums in the Node SDK index.d.ts
- Use readonly objects with `Object.freeze` for node
- Simplify enums object in template params


abnegate:
@loks0n Lets keep the TS enums:
> - Enums prevent usage of the string literal value which is a breaking change.

The breaking change is okay for SDKs, this will be the case for all typed languages, since we now expect the enum type instead of a string

> - Enums are not available is in vanilla JS, meaning we would have to (badly) emulate the implementation for the node SDK.

Node can continue using plain objects instead of TS enums

> Use UPPER_SNAKE_CASE for js enums keys

If this is a convention we are using elsewhere let's do it, otherwise in my experience they should be pascal case

loks0n:
> The breaking change is okay for SDKs, this will be the case for all typed languages, since we now expect the enum type instead of a string
Not a fan of this DX personally but OK

> Node can continue using plain objects instead of TS enums
OK

> If this is a convention we are using elsewhere let's do it, otherwise in my experience they should be pascal case
Both are acceptable just personally preferred UPPER_SNAKE, reverted.

abnegate (approved):

PR #780, opened 2024-03-06
fix: cli missing description

loks0n:


templates/cli/lib/parser.js.twig
```diff
@@ -170,6 +170,7 @@ const commandDescriptions = {
     "logout": `The logout command allows you to logout of your {{ spec.title|caseUcfirst }} account.`,
     "console" : `The console command allows gives you access to the APIs used by the Appwrite console.`,
     "assistant": `The assistant command allows you to interact with the Appwrite Assistant AI`,
+    "messaging": `The messaging command allows you to send and receive messages.`,
```
abnegate, line 173: Let's remove the receive part

PR #800, opened 2024-03-13
Fix 30 add null value handling in user model

theemaster:
## What does this PR do?

Fix `User` Model to handle null values for `password`, `hash` and `hashOptions`.

## Test Plan

When using a JWT to authenticate and use `Account.Get()` there should no longer be an error.

## Related PRs and Issues

[🐛 Bug Report: Bug acquiring users on server side](https://github.com/appwrite/sdk-for-dotnet/issues/30)

## Credit

Credit goes to @joaquingrech.


nove1398:
I am experiencing this same error, was debugging a bit and couldn't find why on my side

joaquingrech:
> I am experiencing this same error, was debugging a bit and couldn't find why on my side

It is because they never merged the fix. It is also in the latest 1.5. I'm surprised it passed release since if you don't use a password and login using Google or any other auth system, all queries for that user will always fail.

theemaster:
Because the fix hasn't been reviewed yet and merged/released I have forked the nuget package and created a temporary package with the package name `theemaster.appwrite`. You could use it temporarily until the fix has been released.

nove1398:
How can we get some more wires on this issue? The changes seem small enough where it could get applied as a patch.

nove1398:
I can confirm for me the password key error stops with your changes, but now i am getting it for mfa as well lol, i think the dictionary values should all be checked by TryGet in this User class

adityaoberai (approved):
Can confirm that these are now nullable @lohanidamodar 
image

LGTM :+1:

adityaoberai:
> I can confirm for me the password key error stops with your changes, but now i am getting it for mfa as well lol, i think the dictionary values should all be checked by TryGet in this User class

Only the Mfa field or Totp also?

nove1398:
> > I can confirm for me the password key error stops with your changes, but now i am getting it for mfa as well lol, i think the dictionary values should all be checked by TryGet in this User class
> 
> Only the Mfa field or Totp also?

Totp too, however I know what my problem is, I was using cloud with the fix meaning the response is trying to parse data the Server is not providing. When I use an older version of the fix it works as expected, and I am able to continue my dev work. Thanks for addressing this so quickly

adityaoberai:
> > > I can confirm for me the password key error stops with your changes, but now i am getting it for mfa as well lol, i think the dictionary values should all be checked by TryGet in this User class
> > 
> > 
> > Only the Mfa field or Totp also?
> 
> Totp too, however I know what my problem is, I was using cloud with the fix meaning the response is trying to parse data the Server is not providing. When I use an older version of the fix it works as expected, and I am able to continue my dev work. Thanks for addressing this so quickly

Ah yes. That would work fine as soon as cloud is upgraded to 1.5.*

@lohanidamodar, in that case we should be good to merge this

nove1398:
> > > > I can confirm for me the password key error stops with your changes, but now i am getting it for mfa as well lol, i think the dictionary values should all be checked by TryGet in this User class
> > > 
> > > 
> > > Only the Mfa field or Totp also?
> > 
> > 
> > Totp too, however I know what my problem is, I was using cloud with the fix meaning the response is trying to parse data the Server is not providing. When I use an older version of the fix it works as expected, and I am able to continue my dev work. Thanks for addressing this so quickly
> 
> Ah yes. That would work fine as soon as cloud is upgraded to 1.5.*
> 
> @lohanidamodar, in that case we should be good to merge this

Thanks #TeamAppwrite ! I am happy to know .Net is not the forgotten language lol :) <3 I look forward to the cloud migration to 1.5 I have some clients I am recommending to come onboard to the platform. #DevRelations

abnegate (approved):

PR #802, opened 2024-03-17
appwrite cli login issue - Update client.js.twig

StarZeus:
This change enables appwrite-cli to capture all cookies in 'set-cookie' response headers.

Fixes https://github.com/appwrite/sdk-for-cli/issues/110

### DIFF
--- a/templates/cli/lib/client.js.twig
+++ b/templates/cli/lib/client.js.twig
@@ -154,11 +154,7 @@ class Client {
 
     let cookies = response.headers.getSetCookie();
     if (cookies && cookies.length > 0) {
-      let cookieString = "";
-      cookies.forEach((cookie) => {
-          cookieString += cookie.split(";")[0] + ";";
-      });        
-      globalConfig.setCookie(cookieString);
+      globalConfig.setCookie(cookies.join(";"));
     }
 
     const text = await response.text();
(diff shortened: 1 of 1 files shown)


templates/cli/lib/client.js.twig
```diff
@@ -154,7 +154,11 @@ class Client {
 
     let cookies = response.headers.getSetCookie();
     if (cookies && cookies.length > 0) {
-      globalConfig.setCookie(cookies[0]);
+      let cookieString = "";
+      cookies.forEach((cookie) => {
+          cookieString += cookie.split(";")[0] + ";";
```
stnguyen90, line 159: Why do we need to do this split?

lohanidamodar: I think this makes sense, if first cookie is set to something else, doing this will save all the kookies.

Meldiron:
```suggestion
      globalConfig.setCookie(cookies.join(";"));
```
Would this do the job? 🤔

StarZeus: That sounds simpler :). Thanks @Meldiron. I'll update the PR.

StarZeus:
@stnguyen90 - Updated the PR. Please let me know if you have any questions.

stnguyen90 (approved):

stnguyen90:
LGTM!

image

PR #803, opened 2024-03-20
✅ Fix query unit tests + Bump package_info_plus package

istornz:
## What does this PR do?

Rework query unit tests.
Upgrade package_info_plus dep to 5.0.1 (as is blocking on some project depending on it)

⚠️ I think I found a bug in the SDK, if someone could check if it's voluntary or not: In the query_test.dart file, line 86.
I think the correct behavior is to return a simple list ['a', 'b', 'c'] like others but actually it return [['a', 'b', 'c']].

### DIFF
--- a/templates/dart/test/query_test.dart.twig
+++ b/templates/dart/test/query_test.dart.twig
@@ -19,10 +19,6 @@ class BasicFilterQueryTest<V, E> {
   });
 }
 
-extension StringJSON on String {
-  Map<String, dynamic> toJson() => jsonDecode(this);
-}
-
 void main() {
   group('basic filter tests', () {
     final tests = [
@@ -61,7 +57,7 @@ void main() {
     group('equal()', () {
       for (var t in tests) {
         test(t.description, () {
-          final query = Query.equal('attr', t.value).toJson();
+          final query = jsonDecode(Query.equal('attr', t.value));
           expect(query['attribute'], 'attr');
           expect(query['values'], t.expectedValues);
           expect(query['method'], 'equal');
@@ -69,7 +65,7 @@ void main() {
       }
 
       test('with a list', () {
-        final query = Query.equal('attr', ['a', 'b', 'c']).toJson();
+        final query = jsonDecode(Query.equal('attr', ['a', 'b', 'c']));
         print(query);
         expect(query['attribute'], 'attr');
         expect(query['values'], ['a', 'b', 'c']);
@@ -80,7 +76,7 @@ void main() {
     group('notEqual()', () {
       for (var t in tests) {
         test(t.description, () {
-          final query = Query.notEqual('attr', t.value).toJson();
+          final query = jsonDecode(Query.notEqual('attr', t.value));
           expect(query['attribute'], 'attr');
           expect(query['values'], t.expectedValues);
           expect(query['method'], 'notEqual');
@@ -88,7 +84,7 @@ void main() {
       }
 
       test('with a list', () {
-        final query = Query.notEqual('attr', ['a', 'b', 'c']).toJson();
+        final query = jsonDecode(Query.notEqual('attr', ['a', 'b', 'c']));
         print(query);
         expect(query['attribute'], 'attr');
         expect(query['values'], [['a', 'b', 'c']]); // Is there a bug here?
@@ -99,7 +95,7 @@ void main() {
     group('lessThan()', () {
       for (var t in tests) {
         test(t.description, () {
-          final query = Query.lessThan('attr', t.value).toJson();
+          final query = jsonDecode(Query.lessThan('attr', t.value));
           expect(query['attribute'], 'attr');
           expect(query['values'], t.expectedValues);
           expect(query['method'], 'lessThan');
@@ -107,7 +103,7 @@ void main() {
       }
 
       test('with a list', () {
-        final query = Query.lessThan('attr', ['a', 'b', 'c']).toJson();
+        final query = jsonDecode(Query.lessThan('attr', ['a', 'b', 'c']));
         print(query);
         expect(query['attribute'], 'attr');
         expect(query['values'], ['a', 'b', 'c']);
@@ -118,7 +114,7 @@ void main() {
     group('lessThanEqual()', () {
       for (var t in tests) {
         test(t.description, () {
-          final query = Query.lessThanEqual('attr', t.value).toJson();
+          final query = jsonDecode(Query.lessThanEqual('attr', t.value));
           expect(query['attribute'], 'attr');
           expect(query['values'], t.expectedValues);
           expect(query['method'], 'lessThanEqual');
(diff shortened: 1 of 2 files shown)


templates/dart/test/query_test.dart.twig
```diff
@@ -17,39 +19,44 @@ class BasicFilterQueryTest {
...
-      BasicFilterQueryTest(
-        description: 'with a list',
-        value: ['a', 'b', 'c'],
-        expectedValues: ["a","b","c"],
-      ),
+      // FIXME: seems to be not working for now... bug?
+      // BasicFilterQueryTest<List<String>>(
+      //   description: 'with a list',
+      //   value: ['a', 'b', 'c'],
+      //   expectedValues: [['a', 'b', 'c']],
```
stnguyen90, line 58: What's not working? 

This should be:

```
expectedValues: ['a', 'b', 'c'],
```

templates/dart/test/query_test.dart.twig
```diff
@@ -1,14 +1,16 @@
...
 import 'package:flutter_test/flutter_test.dart';
 {% endif %}
 
-class BasicFilterQueryTest {
+class BasicFilterQueryTest<V> {
   final String description;
-  final dynamic value;
-  final String expectedValues;
+  final V value;
+  final List<V> expectedValues;
```
stnguyen90, line 13: This isn't always the case. If you pass a string, expected value is List of strings. If you pass in a list of strings, expected value is still list of strings.

templates/dart/test/query_test.dart.twig
```diff
@@ -61,6 +68,14 @@ void main() {
           expect(query['method'], 'equal');
         });
       }
+
+      test('with a list', () {
+        final query = Query.equal('attr', ['a', 'b', 'c']).toJson();
```
stnguyen90, line 73: You can't call toJson() here.

templates/flutter/pubspec.yaml.twig
```diff
@@ -22,7 +22,7 @@ dependencies:
   device_info_plus: ^9.0.2
   flutter_web_auth_2: ^3.0.0
   http: '>=0.13.6 <2.0.0'
-  package_info_plus: ^4.0.2
+  package_info_plus: ^5.0.1
```
stnguyen90, line 25: This is already handled in another PR.

stnguyen90 (changes requested):

templates/dart/test/query_test.dart.twig
stnguyen90: Because `Query.equal()` returns a `String` and `String`'s don't have a `toJson()` method.

istornz: I created a String extension to handle it (line 22) 👍

stnguyen90: Adding the extension would make people question what it is and force people to go look at the implementation...why not just call `jsonDecode()`?

stnguyen90:
@istornz, also why do these tests need to be reworked?

templates/dart/test/query_test.dart.twig
stnguyen90: unresolved

templates/dart/test/query_test.dart.twig
stnguyen90: unresolved

stnguyen90 (changes requested):

templates/dart/test/query_test.dart.twig
istornz: Done

templates/dart/test/query_test.dart.twig
istornz: Done

istornz:
> @istornz, also why do these tests need to be reworked?

Because tests was broken & can't compile/run

templates/flutter/pubspec.yaml.twig
istornz: Downgraded, where's the PR which handle this change?

templates/dart/test/query_test.dart.twig
```diff
@@ -20,195 +22,243 @@ class BasicFilterQueryTest {
...
+        expect(query['values'], ['a', 'b', 'c']);
+        expect(query['method'], 'lessThanEqual');
+      });
     });
 
     group('greaterThan()', () {
       for (var t in tests) {
         test(t.description, () {
-          final query = Query.greaterThan('attr', t.value).toJson();
+          final query = jsonDecode(Query.greaterThan('attr', t.value).toJson());
```
stnguyen90, line 136: missed one.

templates/dart/test/query_test.dart.twig
```diff
@@ -20,195 +22,243 @@ class BasicFilterQueryTest {
...
           expect(query['values'], t.expectedValues);
           expect(query['method'], 'notEqual');
         });
       }
+
+      test('with a list', () {
+        final query = jsonDecode(Query.notEqual('attr', ['a', 'b', 'c']));
+        print(query);
+        expect(query['attribute'], 'attr');
+        expect(query['values'], [['a', 'b', 'c']]); // Is there a bug here?
```
stnguyen90, line 90: Hmm I think the idea with this was that you'd never pass a list to this because it just doesn't make sense.
```
attr != a OR attr != b OR attr != C
```
would always be true.

That said, the implementation makes the behavior weird... @lohanidamodar, what do you think?

templates/dart/test/query_test.dart.twig
```diff
@@ -20,195 +22,243 @@ class BasicFilterQueryTest {
...
         });
       }
+
+      test('with a list', () {
+        final query = jsonDecode(Query.equal('attr', ['a', 'b', 'c']));
+        print(query);
+        expect(query['attribute'], 'attr');
+        expect(query['values'], ['a', 'b', 'c']);
+        expect(query['method'], 'equal');
+      });
```
stnguyen90, line 73: Isn't this already covered in the `tests` array?

stnguyen90 (changes requested):

templates/dart/test/query_test.dart.twig
lohanidamodar: I think it's the database implementation, we don't support arrays for values in few methods. @abnegate might be able to shed some light.

abnegate: Not equal should only provide [a single value](https://github.com/utopia-php/database/blob/a6c1cf81e2822e78bbfcc5b07bdd31d53db6a22d/src/Database/Query.php#L336), the problem is it's hard to represent in languages without union types. This should actually [throw an exception](https://github.com/utopia-php/database/blob/main/src/Database/Validator/Query/Filter.php#L266-L277)

stnguyen90: @istornz, let's leave the list test for [these](https://github.com/utopia-php/database/blob/main/src/Database/Validator/Query/Filter.php#L266-L277) methods out for now. We'll figure out tests and whatnot for these in the future.

stnguyen90 (changes requested):
Please also try to run the tests locally to make sure it all passes.

PR #816, opened 2024-04-23
Test removing okhttp BOM

abnegate:
## What does this PR do?

Fix Android SDK compile issues with Android Gradle Plugin version greater than 8.2

## Test Plan

## Related PRs and Issues

(Write your answer here.)

### DIFF
--- a/templates/android/library/build.gradle.twig
+++ b/templates/android/library/build.gradle.twig
@@ -59,10 +59,7 @@ dependencies {
     api("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.7.1")
     api("org.jetbrains.kotlinx:kotlinx-coroutines-android:1.7.1")
 
-    api(platform("com.squareup.okhttp3:okhttp-bom:4.12.0"))
-    api("com.squareup.okhttp3:okhttp")
-    implementation("com.squareup.okhttp3:okhttp-urlconnection")
-    implementation("com.squareup.okhttp3:logging-interceptor")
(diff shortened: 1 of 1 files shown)


templates/android/library/build.gradle.twig
```diff
@@ -59,10 +59,7 @@ dependencies {
     api("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.7.1")
     api("org.jetbrains.kotlinx:kotlinx-coroutines-android:1.7.1")
 
-    api(platform("com.squareup.okhttp3:okhttp-bom:4.12.0"))
-    api("com.squareup.okhttp3:okhttp")
-    implementation("com.squareup.okhttp3:okhttp-urlconnection")
-    implementation("com.squareup.okhttp3:logging-interceptor")
```
lohanidamodar, line 65: Why did we add these previously? removing these will not cause any issues?

abnegate: > api("com.squareup.okhttp3:okhttp")

We're still using this, just in `implementation` scope instead of `api`. Means that consumers of the SDK won't have direct access to this libraries code. I think I had added this for developing realtime.

>  implementation("com.squareup.okhttp3:okhttp-urlconnection")

Not sure why we had this, we don't need it. This let's you build a Java native `HttpUrlConnection` using okhttp, which we never do.

> implementation("com.squareup.okhttp3:logging-interceptor")

This is a logger for HTTP requests, also added for development testing.

lohanidamodar (approved):

PR #825, opened 2024-05-04
Add missing export for InputFile

stnguyen90:
## What does this PR do?

InputFile wasn't exported so it wasn't possible to import it:

image

This exports it so it can be imported.

## Test Plan

Testing using the playground-for-node:

image

## Related PRs and Issues

None

Yes

### DIFF
--- a/templates/node/src/index.ts.twig
+++ b/templates/node/src/index.ts.twig
@@ -7,6 +7,7 @@ export type { QueryTypes, QueryTypesList } from './query';
 export { Permission } from './permission';
 export { Role } from './role';
 export { ID } from './id';
+export { InputFile } from './inputFile';
 {% for enum in spec.enums %}
 export { {{ enum.name | caseUcfirst  }} } from './enums/{{enum.name | caseDash}}';
 {% endfor %}


loks0n (approved):

templates/node/src/index.ts.twig
```diff
@@ -7,6 +7,7 @@ export type { QueryTypes, QueryTypesList } from './query';
 export { Permission } from './permission';
 export { Role } from './role';
 export { ID } from './id';
+export { InputFile } from './inputFile';
```
loks0n, line 10: It's now on a seperate entry point. So instead you do 

```
import { InputFile } from 'node-appwrite/file'
```

If we put it on this entrypoint edge runtimes will break as inputfile depends on node fs module

loks0n (changes requested):

templates/node/src/index.ts.twig
stnguyen90: Is that a separate package? I wasn't getting any suggestions on what to import:

image

That import also didn't work for me:

image

Where is this documented?

loks0n: https://github.com/appwrite/sdk-generator/blob/3ade92cbc7f6498c024314199f31237cfe13c80d/templates/node/package.json.twig#L23

Should be defined here in the package.json

Not sure why you don't have the types

stnguyen90: @loks0n, and it works for you? Can you try on the https://github.com/appwrite/playground-for-node repo?

stnguyen90: FYI, if I just put /, I get this as autocomplete:

image

loks0n: Works for me with CJS and ESM. Are you 100% sure you have the rc version installed?
Screenshot 2024-05-10 at 10 36 45

stnguyen90: Ahh looks like I had a bad tsconfig.json file. It had `module` set to `commonjs` and maybe something else that caused problems. I created a fresh project and was able to resolve the dependency fine 🎉

PR #828, opened 2024-05-08
fix: esm import extensions

loks0n:
uses workaround from https://github.com/egoist/tsup/issues/953

Related
* https://github.com/appwrite/sdk-for-node/issues/86


stnguyen90 (approved):
It's pretty hacky but at least it works.

image

TorstenDittmann (approved):

PR #830, opened 2024-05-10
Fix Apple realtime with multiple subscriptions

abnegate:
## What does this PR do?

- **BREAKING CHANGE**: `Realtime.subscribe` is now async
- Fixes https://github.com/appwrite/sdk-for-apple/issues/55
- Fixes https://github.com/appwrite/sdk-for-apple/issues/56

## Test Plan

TBD

## Related PRs and Issues

* https://github.com/appwrite/sdk-for-apple/issues/55
* https://github.com/appwrite/sdk-for-apple/issues/56

Yes

### DIFF
--- a/templates/swift/Sources/Services/Realtime.swift.twig
+++ b/templates/swift/Sources/Services/Realtime.swift.twig
@@ -13,6 +13,8 @@ open class Realtime : Service {
     private var activeChannels = Set<String>()
     private var activeSubscriptions = [Int: RealtimeCallback]()
 
+    let connectSync = DispatchQueue(label: "ConnectSync")
+
     private var subCallDepth = 0
     private var reconnectAttempts = 0
     private var subscriptionsCounter = 0
@@ -120,7 +122,9 @@ open class Realtime : Service {
             with: callback
         )
 
-        subCallDepth += 1
+        connectSync.sync {
+            subCallDepth+=1
+        }
 
         try await Task.sleep(nanoseconds: UInt64(DEBOUNCE_NANOS))
 
@@ -128,7 +132,9 @@ open class Realtime : Service {
             try await self.createSocket()
         }
 
-        self.subCallDepth -= 1
+        connectSync.sync {
+            self.subCallDepth -= 1
+        }
 
         return RealtimeSubscription {
             self.activeSubscriptions[self.subscriptionsCounter] = nil
--- a/templates/swift/Sources/Services/Realtime.swift.twig
+++ b/templates/swift/Sources/Services/Realtime.swift.twig
@@ -55,11 +55,12 @@ open class Realtime : Service {
             return
         }
 
-        let promise = group.any().makePromise(of: Void.self)
+        if (client.isConnected) {
+            let promise = group.any().makePromise(of: Void.self)
+            client.close(promise: promise)
+            try await promise.futureResult.get()
+        }
 
-        client.close(promise: promise)
-
-        try await promise.futureResult.get()
         try await group.shutdownGracefully()
     }
 
--- a/templates/swift/Sources/Services/Realtime.swift.twig
+++ b/templates/swift/Sources/Services/Realtime.swift.twig
@@ -39,9 +39,9 @@ open class Realtime : Service {
         if (socketClient != nil) {
             reconnect = false
             closeSocket()
-        } else {
-            socketClient = WebSocketClient(url, tlsEnabled: !client.selfSigned, delegate: self)!
         }
+    
+        socketClient = WebSocketClient(url, tlsEnabled: !client.selfSigned, delegate: self)!
 
         try! socketClient?.connect()
     }
--- a/templates/swift/Sources/Services/Realtime.swift.twig
+++ b/templates/swift/Sources/Services/Realtime.swift.twig
@@ -7,24 +7,21 @@ open class Realtime : Service {
 
     private let TYPE_ERROR = "error"
     private let TYPE_EVENT = "event"
-    private let DEBOUNCE_MILLIS = 1
+    private let DEBOUNCE_NANOS = 1_000_000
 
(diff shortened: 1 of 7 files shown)


stnguyen90 (changes requested):
A contributor tested something like this and the error that [this](https://github.com/appwrite/sdk-generator/pull/284/commits/8a3221699821698a01e321739dda74ca35f3dfa5) was supposed to prevent came back:

```
NIOCore/EventLoop.swift:1168: Fatal error: BUG DETECTED: syncShutdownGracefully() must not be called when on an EventLoop.
Calling syncShutdownGracefully() on any EventLoop can lead to deadlocks.
Current eventLoop: SelectableEventLoop { selector = Selector { descriptor = 5 }, thread = NIOThread(name = NIO-ELT-0-#0) }
```

templates/swift/Sources/Services/Realtime.swift.twig
```diff
@@ -38,17 +37,30 @@ open class Realtime : Service {
...
         }
 
-        try! socketClient?.connect()
+        socketClient = WebSocketClient(
+            url,
+            tlsEnabled: !client.selfSigned,
+            delegate: self
+        )
+
+        try await socketClient?.connect()
```
stnguyen90, line 49: I'm having trouble testing this. I updated the playground for apple's subscribe call [here](https://github.com/appwrite/playground-for-apple-swiftui/blob/50c601345421346dfd6fe330e201537a929cc96e/PlaygroundForApple/PlaygroundViewModel.swift#L138-L144) to be:

```swift
    func subscribe() async {
        do {
            try await realtime.subscribe(channel: "databases.\(databaseId).collections.\(collectionId).documents") { event in
                DispatchQueue.main.async {
                    self.message = String(describing: event.payload!)
                }
            }
    
            try await realtime.subscribe(channel: "files") { event in
                DispatchQueue.main.async {
                    self.message = String(describing: event.payload!)
                }
            }
        } catch {
            dialogText = error.localizedDescription
        }
    }
```

but it seems like the files event isn't coming through. The document one does, though. And if I swap them like:

```swift
    func subscribe() async {
        do {
            try await realtime.subscribe(channel: "files") { event in
                DispatchQueue.main.async {
                    self.message = String(describing: event.payload!)
                }
            }

            try await realtime.subscribe(channel: "databases.\(databaseId).collections.\(collectionId).documents") { event in
                DispatchQueue.main.async {
                    self.message = String(describing: event.payload!)
                }
            }
        } catch {
            dialogText = error.localizedDescription
        }
    }
```

The files one will come through, but the documents one won't. 

Any ideas on what I'm doing wrong?

templates/swift/Sources/Services/Realtime.swift.twig
```diff
@@ -98,36 +110,38 @@ open class Realtime : Service {
...
+        }
+
+        connectSync.sync {
+            self.subCallDepth -= 1
         }
 
         return RealtimeSubscription {
-            self.activeSubscriptions[counter] = nil
+            self.activeSubscriptions[count] = nil
             self.cleanUp(channels: channels)
```
lohanidamodar, line 143: I think this part is causing the issue, when the second subscribe comes in, the channel from first subscribe is removed by the cleanup as it's not in the list of current channels.

lohanidamodar: We might need to update the cleanup method to also check for active subscriptions or perhaps we can do the cleanup on unsubscribe/close only?

lohanidamodar (changes requested):
Left comment

templates/swift/Sources/Services/Realtime.swift.twig
stnguyen90: Not sure about this because when I tested, I had breakpoints set and I saw that the 2nd create socket call had all the channels. I also didn't test closing subscriptions.

templates/swift/Sources/Services/Realtime.swift.twig
```diff
@@ -98,36 +110,38 @@ open class Realtime : Service {
...
+            try await self.createSocket()
+        }
+
+        connectSync.sync {
+            self.subCallDepth -= 1
         }
 
         return RealtimeSubscription {
-            self.activeSubscriptions[counter] = nil
+            self.activeSubscriptions[count] = nil
```
lohanidamodar, line 142: This too troubles me, we use active subscription to send back callback (https://github.com/appwrite/sdk-generator/blob/f67b1a4b953e82d32cfe842452f842867711c36d/templates/swift/Sources/Services/Realtime.swift.twig#L218-L219)

templates/swift/Sources/Services/Realtime.swift.twig
lohanidamodar: 🤔 Then I might be wrong here.

abnegate: The cleanup is only run when this `close` function is called, and the `count` and `chanels` are captured by value, so it will only remove the subscription and channels that were added for that particular call to `subscribe`

templates/swift/Sources/Services/Realtime.swift.twig
abnegate: This is expected, the lambda passed to the `RealtimeSubscription` is the `close` function. So when close is called, we remove it from the activeSubscriptions

templates/swift/Sources/Services/Realtime.swift.twig
abnegate: Hmm in my tests I had a delay between subscribes, might be an issue with the debounce. Will check

abnegate: Fixed

lohanidamodar (approved):

stnguyen90 (approved):
I was able to get multiple events through 🎉

PR #833, opened 2024-05-13
Feat(cli): Deploying function when none was selected

byawitz:
If no function was selected using the deployment action, then all the functions will be deployed.

Also, a note will be displayed to the user for using the `-all` flag.

> A small `tip` utility was added as tip console log helper.

### DIFF
--- a/templates/cli/lib/commands/deploy.js.twig
+++ b/templates/cli/lib/commands/deploy.js.twig
@@ -240,27 +240,21 @@ const deployFunction = async ({ functionId, all, yes } = {}) => {
     let response = {};
 
     const functionIds = [];
+    const configFunctions = localConfig.getFunctions();
 
     if (functionId) {
         functionIds.push(functionId);
     } else if (all) {
-        const functions = localConfig.getFunctions();
-        if (functions.length === 0) {
+        if (configFunctions.length === 0) {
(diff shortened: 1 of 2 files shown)


byawitz:
Extracted all logic to a `selectedOrAll` function.

Attached is a photo of successful run of that new feature
image

templates/cli/lib/commands/deploy.js.twig
```diff
@@ -907,6 +905,22 @@ const deployTeam = async ({ all, yes } = {}) => {
...
+    const ids = []
+
+    const answers = await inquirer.prompt(questions);
+    ids.push(...answers[resourceName]);
+
+    if (ids.length <= 0) {
+        const all = data instanceof Function ? data() : data.map((resource) => resource.$id)
+
+        ids.push(...all);
+        tip(`Deploying All available ${resourceName}. Use the '--all' flag to execute this action intentionally\n`)
```
loks0n, line 918: Not sure I understand the message.

Shouldn't it be:
`Deploying all available ${resourceName} as none specified. Use the `--all` flag to disable this warning.`

loks0n: @gewenyu99 any suggestions?

PR #841, opened 2024-05-17
Fix error caused by variable named as default in .NET SDK

adityaoberai:
## What does this PR do?

Fixes error caused by variable named as `default` in .NET SDK (`default` is keyword in .NET and can't be used as a variable name)

## Test Plan

Generate the SDK using `example.php`
image

## Related PRs and Issues

No

Yes


templates/dotnet/src/Appwrite/Models/Model.cs.twig
```diff
@@ -40,7 +40,7 @@ namespace {{ spec.title | caseUcfirst }}.Models
...
...overrideIdentifier}}.From(map: it)).ToList(){% else %}{{property.sub_schema | caseUcfirst | overrideIdentifier}}.From(map: ((JObject)map["{{ property.name }}"]).ToObject<Dictionary<string, object>>()!){% endif %}{% else %}{% if property.type == 'array' %}((JArray)map["{{ property.name }}"]).ToObject<{{ property | typeName }}>(){% else %}{% if property.type == "integer" or property.type == "number" %}{% if not property.required %}map["{{ property.name }}"] == null ? null : {% endif %}Convert.To{% if property.type == "integer" %}Int64{% else %}Double{% endif %}(map["{{ property.name }}"]){% else %}{% if property.type == "boolean" %}({{ property | typeName }}{% if not property.required %}?{% endif %})map["{{ property.name }}"]{% else %}map{% if not property.required %}.TryGetValue("{{ property.name }}", out var {% if property.name == "default" %}x{% endif %}{{ property.name }}) ? {% if property.name == "default" %}x{% endif %}{{ property.name }}.ToString() : null{% else %}["{{ property.name }}"]{% if not property.required %}?{% endif %}.ToString(){% endif %}{% endif %}{% endif %}{% endif %}{% endif %}{% if not loop.last or (loop.last and definition.additionalProperties) %},{% endif %}
```
stnguyen90, line 43: hmm we have a way to handle keywords...  maybe we need to do something like:

```suggestion
            {{ property.name | caseCamel | escapeKeyword | removeDollarSign }}: {% if property.sub_schema %}{% if property.type == 'array' %}((JArray)map["{{ property.name }}"]).ToObject<List<Dictionary<string, object>>>().Select(it => {{property.sub_schema | caseUcfirst | overrideIdentifier}}.From(map: it)).ToList(){% else %}{{property.sub_schema | caseUcfirst | overrideIdentifier}}.From(map: ((JObject)map["{{ property.name }}"]).ToObject<Dictionary<string, object>>()!){% endif %}{% else %}{% if property.type == 'array' %}((JArray)map["{{ property.name }}"]).ToObject<{{ property | typeName }}>(){% else %}{% if property.type == "integer" or property.type == "number" %}{% if not property.required %}map["{{ property.name }}"] == null ? null : {% endif %}Convert.To{% if property.type == "integer" %}Int64{% else %}Double{% endif %}(map["{{ property.name }}"]){% else %}{% if property.type == "boolean" %}({{ property | typeName }}{% if not property.required %}?{% endif %})map["{{ property.name }}"]{% else %}map{% if not property.required %}.TryGetValue("{{ property.name }}", out var {{ property.name | caseCamel | escapeKeyword | removeDollarSign }}) ? {{ property.name | caseCamel | escapeKeyword | removeDollarSign }}.ToString() : null{% else %}["{{ property.name }}"]{% if not property.required %}?{% endif %}.ToString(){% endif %}{% endif %}{% endif %}{% endif %}{% endif %}{% if not loop.last or (loop.last and definition.additionalProperties) %},{% endif %}
```

adityaoberai: Oh okay, I wasn't aware of that. 'default' is present in our list of keywords though, so this should ideally work 👍

adityaoberai: Lemme do a quick test run

adityaoberai: image

Works fine @stnguyen90 👍 
 
 Committing the suggestion

stnguyen90 (approved):

abnegate (approved):

PR #848, opened 2024-05-20
Feat create resources

byawitz:
## What does this PR do?
Interactive resources adding.

Already exists
- Create project
- Create function

Added
- Buckets
- Collections
- Topics

## Test Plan
Tested against Appwrite version `1.5.5`

image


templates/cli/lib/commands/pull.js.twig
```diff
@@ -26,142 +23,77 @@ const pull = new Command("pull")
...
-    let sdk = await sdkForConsole();
-    if (answers.start === "new") {
-        response = await teamsCreate({
-            teamId: 'unique()',
-            name: answers.project,
-            sdk,
-            parseOutput: false
-        })
+    try {
+        let response = await projectsList({
```
Meldiron, line 27: Can we use `projectsGet` instead?

templates/cli/lib/commands/pull.js.twig
```diff
@@ -26,142 +23,77 @@ const pull = new Command("pull")
...
-    })
+        const localFunction = localFunctions.find((localFunc) => localFunc['$id'] === func['$id']);
 
-    fs.mkdirSync(functionDir, "777");
+        if (localFunction['deployment'] === '') {
+            continue
+        }
 
-    let gitInitCommands = "git clone -b v3 --single-branch --depth 1 --sparse https://github.com/{{ sdk.gitUserName }}/functions-starter ."; // depth prevents fetching older commits reducing the amount fetched
+        const compressedFileName = `${+new Date()}.tar.gz`
```
Meldiron, line 77: Let's include function ID too

templates/cli/lib/commands/pull.js.twig
```diff
@@ -26,142 +23,77 @@ const pull = new Command("pull")
...
+            overrideForCli: true,
+            parseOutput: false
+        })
 
-    /* Force use CMD as powershell does not support && */
-    if (process.platform === 'win32') {
-        gitInitCommands = 'cmd /c "' + gitInitCommands + '"';
-        gitPullCommands = 'cmd /c "' + gitPullCommands + '"';
-    }
+        tar.extract({
```
Meldiron, line 87: Not sure if relevant, but we not only tar, we also zip (tar.gz). We may need to unzip too, or provide some parameter to tar to tell it to unzip too.

If you did QA and it worked well, you can ignore this comment

templates/cli/lib/commands/pull.js.twig
```diff
@@ -253,19 +185,21 @@ const pullMessagingTopic = async () => {
 
 pull
     .command("project")
-    .description("Pulling your {{ spec.title|caseUcfirst }} project")
+    .description("Pulling your Appwrite project name")
```
Meldiron, line 188: Let's stick to `{{ spec.title|caseUcfirst }}`, unless this was recommended by Christy

templates/cli/lib/commands/pull.js.twig
```diff
@@ -253,19 +185,21 @@ const pullMessagingTopic = async () => {
...
-    .action(actionRunner(pullFunction))
+    .description(`Pulling your Appwrite functions`)
+    .option(`--yes`, `Flag to confirm all warnings`)
+    .option(`--all`, `Flag to pull all functions`)
+    .action(actionRunner(pullFunctions));
 
 pull
     .command("collection")
-    .description("Pulling your {{ spec.title|caseUcfirst }} collections")
+    .description("Pulling your Appwrite collections")
```
Meldiron, line 200: Let's stick to `{{ spec.title|caseUcfirst }}`, unless this was recommended by Christy

templates/cli/lib/commands/pull.js.twig
```diff
@@ -26,142 +23,77 @@ const pull = new Command("pull")
...
+    for (let func of selectedFunctions) {
+        const functionExistLocally = localFunctions.find((localFunc) => localFunc['$id'] === func['$id']) !== undefined;
 
-    if (!answers.runtime.commands) {
-        log(`Installation command for this runtime not found. You will be asked to configure the install command when you first push the function.`);
-    }
+        if (!overridingLocalChanges && functionExistLocally) {
+            log(`Skipping locally found implementation of ${func['name']}`)
+            continue;
+        }
```
Meldiron, line 63: Not sure this makes sense. We dotn have `overridingLocalChanges` in other commands, so I would guess this is saying if source code should be pulled or not.

This `continue` is before pulling info into localConfig. I think it should be after.

Also, copy is a bit inconsistent with Appwrite's terminology. Let's use terms like build, deployment, function settings, ..

templates/cli/lib/commands/pull.js.twig
```diff
@@ -253,19 +185,21 @@ const pullMessagingTopic = async () => {
...
+    .option(`--all`, `Flag to pull all functions`)
+    .action(actionRunner(pullFunctions));
 
 pull
     .command("collection")
-    .description("Pulling your {{ spec.title|caseUcfirst }} collections")
+    .description("Pulling your Appwrite collections")
     .option(`--databaseId <databaseId>`, `Database ID`)
-    .option(`--all`, `Flag to pullialize all databases`)
+    .option(`--all`, `Flag to pull all databases`)
```
Meldiron, line 202: Previous was `pullialize`. Let's do global search to make sure it isn't used anywhere else.

templates/cli/lib/questions.js.twig
```diff
@@ -212,11 +208,35 @@ const questionsPullProject = [
...
+                    name: `${func.name} (${func.$id})`,
+                    value: func
+                }
+            });
         }
+    },
+    {
+        type: "input",
+        name: "override",
+        message: `Are you sure you want to override local functions code and definition? ${chalk.red('All local changes will be lost!')} Type "YES" to confirm.`
```
Meldiron, line 235: No is also valid answer here, but we make it sound like it's not. Instead of `are you sure` we could ask `do you want to ...`, or something in that tone.

templates/cli/lib/commands/init.js.twig
```diff
@@ -0,0 +1,261 @@
...
+        });
+    } catch (e) {
+        log('You must login first')
+        await loginCommand()
+    }
+
+    const answers = await inquirer.prompt(questionsInitProject)
+    if (!answers.project || !answers.organization) {
+        process.exit(1)
+    }
```
Meldiron, line 42: Is this needed? Can we instead mark question as required? Please also do global search for process.exit and see if it can be fixed in other places too

templates/cli/lib/commands/init.js.twig
```diff
@@ -0,0 +1,261 @@
...
+        localConfig.setProject(answers.project.id, answers.project.name);
+    }
+
+    success();
+}
+
+const initBucket = async () => {
+    let response = {}
+    const answers = await inquirer.prompt(questionsCreateBucket)
+    if (!answers.bucket || !answers.id || !answers.fileSecurity) process.exit(1)
```
Meldiron, line 63: Same comment as above

templates/cli/lib/commands/init.js.twig
```diff
@@ -0,0 +1,261 @@
...
+            name: answers.collection,
+            documentSecurity: answers.documentSecurity.toLowerCase() === 'yes',
+            enabled: true,
+            parseOutput: false
+        })
+
+        localConfig.addCollection(response);
+        success();
+    } catch (e) {
+        error(e.getMessage ?? 'Unknown error occurred. Please try again');
```
Meldiron, line 99: i dont think `getMessage` works, I think its `e.message`. Please test, and replace if needed. If replaced, please check all places with global search

templates/cli/lib/commands/pull.js.twig
Meldiron: Better idea, just pull without asking questions - both code and settings

templates/cli/lib/commands/init.js.twig
byawitz: It's needed as the user can decide not to override the project, though the checking can be changed

Meldiron (approved):

templates/cli/lib/commands/push.js.twig
```diff
@@ -255,6 +256,19 @@ const pushResources = async ({ all, yes } = {}) => {
     }
 };
 
+const pushProject = async () => {
+    try {
+        await projectsUpdate({
+            projectId: localConfig.getProject().projectId,
+            name: localConfig.getProject().projectName,
+            parseOutput: false
```
christyjacob4, line 264: This should also push the project settings right?

byawitz: Yes. But pushing is done manually per service and auth

byawitz: The changes them self for pushing are in the other PR

templates/cli/lib/commands/init.js.twig
```diff
@@ -0,0 +1,281 @@
...
+        response = await projectsCreate({
+            projectId: answers.id,
+            name: answers.project,
+            teamId: answers.organization.id,
+            parseOutput: false
+        })
+
+        localConfig.setProject(response['$id'], response.name);
+    } else {
+        localConfig.setProject(answers.project.id, answers.project.name);
```
christyjacob4, line 59: shouldn't we also get the project settings and set them too?

christyjacob4 (approved):
Looks good. Just a question

PR #853, opened 2024-05-23
Feat cli sudo

byawitz:
## What does this PR do?
- Refactor `--yes` to `--force`
- Adding two aliases `-a` to `--all` and `-f` for `--force`
- Making these flags available globally 

## Test Plan

image


templates/cli/index.js.twig
```diff
@@ -28,12 +28,24 @@ program
   .version(version, "-v, --version")
   .option("--verbose", "Show complete error log")
   .option("--json", "Output in JSON format")
+  .option("-f,--force", "Flag to confirm all warnings")
+  .option("-a,--all", "Flag to push all resources")
+  .option("--ids [id...]", "Flag to pass list of ids for a giving action")
```
Meldiron, line 33: Hmm not sure if it should be `--id` or `--ids`. Christy to decide

templates/cli/lib/commands/push.js.twig
```diff
@@ -1023,14 +1024,14 @@ const pushTeam = async ({ all, yes } = {}) => {
...
+const pushMessagingTopic = async () => {
     let response = {};
 
     let topicsIds = [];
     const configTopics = localConfig.getMessagingTopics();
-    let overrideExisting = yes;
+    let overrideExisting = cliConfig.force;
 
-    if (all) {
+    if (cliConfig.all) {
```
Meldiron, line 1034: As part of this PR, can we also simplify all actions? For example here, this is the flow:

- If all, push all from cofig to local array
- If not all, ask question, then set answer to local array

What I was thinking instead:
- Ask question

The logic for `all` or `force` or others will all live inside `questions.js`. There, we could wrap question in functions like `allowAll`, `allowForce`.. Let's imagine in future we have this like of questions:

- What is your name?
- Select gender
- Want to save profile?

and I ran question with `--yes --id="male`, then I should be asked name, and that's it. Instead of asking me rest of questions, it can pre-fill values  from tags. Also, if all questions are automated, it should not ask at all - pre-fill and submit everything.

templates/cli/index.js.twig
Meldiron: Prefer:
appwrite push function --id createPayment --id deletePayment

If not supported, then
appwrite push function --ids createPayment deletePayment

templates/cli/lib/commands/push.js.twig
```diff
@@ -864,30 +862,20 @@ const pushCollection = async ({ all, yes } = {}) => {
 
         success(`Created ${relationships.length} relationship attributes`);
     }
+
+    process.exit(0);
```
Meldiron, line 866: Why do we need to add those exists?

byawitz: Because `inquirer js` doesn't support auto-answer, the workaround for enabling this feature sometimes leaves the console hanging. For that, `process.exit(0)` will ensure that the process finishes.

templates/cli/lib/questions.js.twig
```diff
@@ -306,165 +324,344 @@ const questionsLogin = [
...
+    }
 ];
 
+const questionsGetEntrypoint = [
+    {
+        type: "input",
+        name: "entrypoint",
+        message: "Enter the entrypoint",
+        default: null,
+        validate(value) {
```
christyjacob4, line 656: Is there a way to add an example param here like a hint? 

Else maybe we can do `message: "Enter the entrypoint (example: src/index.js )",`

templates/cli/lib/commands/pull.js.twig
```diff
@@ -164,26 +164,12 @@ const pullFunction = async () => {
...
-    if (databaseId) {
-        databaseIds.push(databaseId);
-    } else if (all) {
-        let allDatabases = await databasesList({
-            parseOutput: false
-        })
-
-        databaseIds.push(...allDatabases.databases.map((d) => d.$id));
-    }
+    for (let dbID of databasesIds) {
```
christyjacob4, line 170:
```suggestion
    for (let id of databasesIds) {
```
`id` or `databaseId`

templates/cli/lib/commands/pull.js.twig
```diff
@@ -198,14 +184,14 @@ const pullCollection = async ({ all, databaseId } = {}) => {
...
-        collections.forEach(async collection => {
+        await Promise.all(collections.map(async collection => {
             log(`Fetching ${collection.name} ...`);
             localConfig.addCollection({
                 ...collection,
                 '$createdAt': undefined,
                 '$updatedAt': undefined,
             });
-        });
+        }))
```
christyjacob4, line 194: why does this need to be in a promise? Even the `async` may not be required

templates/cli/lib/commands/push.js.twig
```diff
@@ -312,8 +310,8 @@ const pushFunction = async ({ functionId, all, yes, async } = {}) => {
                     func.pushVariables = true;
                 } else if (total > 0 && !func.pushVariables) {
                     log(`The function ${func.name} has remote variables setup`);
-                    const variableAnswers = await inquirer.prompt(questionsPushFunctions[1])
-                    func.pushVariables = variableAnswers.override.toLowerCase() === "yes";
+                    const variableAnswers = await inquirer.prompt(questionsPushFunctions.slice(2, 4));
+                    func.pushVariables = variableAnswers.setupoverride ? variableAnswers.setupoverride : variableAnswers.override.toLowerCase() === "yes";
```
christyjacob4, line 314: Same here .what is the need for the slice ?

templates/cli/lib/questions.js.twig
byawitz: For that we will need to fetch the runtime from the appwrite instance.

templates/cli/lib/commands/push.js.twig
```diff
@@ -241,42 +252,31 @@ const pushResources = async ({ all, yes } = {}) => {
...
+
+    process.exit(0);
 };
 
-const pushFunction = async ({ functionId, all, yes, async } = {}) => {
+const pushFunction = async ({ async } = {}) => {
     let response = {};
 
-    const functionIds = [];
+    const functionIds = (await inquirer.prompt(questionsPushFunctions.slice(0, 2))).functions;
```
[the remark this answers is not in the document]
byawitz: As it changes to auto answer, the first question `setup` the answer in case the second question is not neccesery

templates/cli/lib/commands/push.js.twig
byawitz: The same, in case the user passes the `--force` flag, then we won't ask the question, this is done like so, to make the flag control from the question instead of the code-flow itself.

templates/cli/lib/paginate.js.twig
```diff
@@ -5,11 +5,15 @@ const paginate = async (action, args = {}, limit = 100, wrapper = '') => {
...
-
+        const additionalQueries = [];
+        if (args.queries) {
+            additionalQueries.push(...args.queries);
+        }
         // Merge the limit and offset into the args
         const response = await action({
             ...args,
             queries: [
+                ...additionalQueries,
```
christyjacob4, line 16: Why do we have this code here as part of this PR? Is this an existing bug ?

christyjacob4 (approved):

christyjacob4 (approved):

PR #854, opened 2024-05-23
Feat non destructive db actions

byawitz:
## What does this PR do?

- `appwrite.json` can be used as a db migration tool.
- Nondestructive changes will execute on the spot, such as removing or adding attributes and indexes.
- Destructive will display a table of changes and wait for the user's decision.
- Parallel creation of database & collection
- Attributes are created in a serialized fashion, ensuring attribute order.

## Test Plan

The Apprwrite instance can be seen in the background, and this is what happening while I'm changing the `appwrite.json` file
1. Fields are being created. Existing fields don't touch.
2. When destructive actions must be taken, a visualized table is displayed to the user, noting the differences.
3. Removed amount fields get removed.

https://github.com/appwrite/sdk-generator/assets/316103/2343bfcb-8f6f-465f-9ec4-aeb12d3ea058


templates/cli/lib/questions.js.twig
```diff
@@ -368,9 +369,9 @@ const questionsPushCollections = [
     },
     {
         type: "input",
-        name: "override",
-        message: 'Are you sure you want to override this collection? This can lead to loss of data! Type "YES" to confirm.'
-    },
+        name: "changes",
+        message: `These are the pending destructive changes. To create the new fields, all data in the old ones will be ${chalk.red('deleted')}. Type "YES" to confirm`
```
Meldiron, line 373:
```suggestion
        message: `Changes above will cause recreation of an attribute. All existing documents will lose data in those attributes. Type "YES" to confirm`
```

templates/cli/lib/commands/push.js.twig
```diff
@@ -645,9 +679,92 @@ const createAttribute = async (databaseId, collectionId, attribute) => {
...
+}
+
+const deepSimilar = (remote, local, collection) => {
+    if (local === undefined) {
+        return undefined;
+    }
+
+    const key = `${chalk.yellow(local.key)} in ${collection.name} (${collection['$id']})`;
 
+    if (remote.type !== local.type) {
```
Meldiron, line 700: Hmm if we add something new in future, we will surely forget to add it here.
Can we make JSON comparison instead? And if needed we can ignore some keys like $id,name,key.. If relevant.

templates/cli/lib/commands/push.js.twig
```diff
@@ -645,9 +679,92 @@ const createAttribute = async (databaseId, collectionId, attribute) => {
...
+const findMatch = (attribute, attributes) => attributes.find((attr) => attr.key === attribute.key);
+
+const updatedList = async (remoteAttributes, localAttributes, collection) => {
+    const deleting = remoteAttributes.filter((attribute) => !findMatch(attribute, localAttributes));
+    const changes = remoteAttributes.map((attribute) => deepSimilar(attribute, findMatch(attribute, localAttributes), collection)).filter(attribute => attribute !== undefined);
+    let changedAttributes = [];
+
+    if (changes.length > 0) {
+        log('There is a conflict in your collection deployment');
+        drawTable(changes.map((change) => {
```
Meldiron, line 736: I like the direction, but I think we should take it 1 step further.

Instead of showing only changes that require attribute recreation, we should show all changes. Including new/deleted indexes/attributes.

We could have extra columns to explain how to difference will be applied. For example, new attribute isn't destructive, but editing length or deleting index are.

Then, always ask them to confirm those changes. No matter if they are destructive or not.

byawitz:
image

Full changes table

templates/cli/lib/commands/push.js.twig
```diff
@@ -666,204 +793,154 @@ const pushCollection = async ({ all, yes } = {}) => {
             collections.push(collection);
         })
     }
+    const databases = Array.from(new Set(collections.map(c => c['databaseId'])));
```
christyjacob4, line 796:
```suggestion
    const databases = Array.from(new Set(collections.map(collection => collection['databaseId'])));
```
lets always use the full names to be more verbose.

templates/cli/lib/commands/push.js.twig
```diff
@@ -666,204 +793,154 @@ const pushCollection = async ({ all, yes } = {}) => {
...
 
-    for (let collection of collections) {
-        log(`Pushing collection ${collection.name} ( ${collection['databaseId']} - ${collection['$id']} )`)
-
-        let databaseId;
-
-        const localDatabase = localConfig.getDatabase(collection.databaseId);
+    // Parallel db actions
+    await Promise.all(databases.map(async (dbId) => {
```
christyjacob4, line 799:
```suggestion
    await Promise.all(databases.map(async (databaseId) => {
```
Lets use the full name here or just `id` and not use the short form `db`

templates/cli/lib/commands/push.js.twig
```diff
@@ -666,204 +793,154 @@ const pushCollection = async ({ all, yes } = {}) => {
...
+        });
+    }
 
-        await Promise.all(relationships.map(attribute => {
-            return createAttribute(collection['databaseId'], collection['$id'], attribute);
-        }));
+    const result = await awaitPools.expectIndexes(
+        collection['databaseId'],
+        collection['$id'],
+        indexes.map(attribute => attribute.key)
```
christyjacob4, line 917:
```suggestion
        indexes.map(index => index.key)
```

templates/cli/lib/questions.js.twig
```diff
@@ -368,9 +369,9 @@ const questionsPushCollections = [
     },
     {
         type: "input",
-        name: "override",
-        message: 'Are you sure you want to override this collection? This can lead to loss of data! Type "YES" to confirm.'
-    },
+        name: "changes",
+        message: `Changes above will cause recreation of an attribute. All existing documents will lose data in those attributes. Type "YES" to confirm`
```
christyjacob4, line 373: Previously the message was about overriding the collection. Now the message has been changes to something specific like attribute creation. Is this correct? 

What if there is a deletion of attribute, or changes to indexes? This `message` does not talk about those scenarios

templates/cli/lib/commands/push.js.twig
```diff
@@ -666,204 +793,154 @@ const pushCollection = async ({ all, yes } = {}) => {
...
-            await databasesUpdateCollection({
-                databaseId,
-                collectionId: collection['$id'],
-                name: collection.name,
-                documentSecurity: collection.documentSecurity,
-                permissions: collection['$permissions'],
-                enabled: collection.enabled,
-                parseOutput: false
-            })
+            collection.isExisted = true;
```
christyjacob4, line 849: What is the function of this attribute `isExisted` . Is this saved in the appwrite.json file?

templates/cli/lib/commands/push.js.twig
```diff
@@ -645,9 +681,100 @@ const createAttribute = async (databaseId, collectionId, attribute) => {
...
+
+    if (changing.length === 0) {
+        return changedAttributes;
+    }
+    log('There are pending changes in your collection deployment');
+
+    drawTable(changing.map((change) => {
+        return { Key: change.key, Action: change.action, Reason: change.reason, };
+    }));
+
```
christyjacob4, line 745:
```suggestion
    
    const deleting = remoteAttributes.filter((attribute) => !findMatch(attribute, localAttributes)).map((attr) => mapChangeAttribute(attr, collection, false));
    const adding = localAttributes.filter((attribute) => !findMatch(attribute, remoteAttributes)).map((attr) => mapChangeAttribute(attr, collection, true));
    const conflicts = remoteAttributes.map((attribute) => deepSimilar(attribute, findMatch(attribute, localAttributes), collection)).filter(attribute => attribute !== undefined);
    
    let changedAttributes = [];
    const changing = [...deleting, ...adding, ...conflicts]
    if (changing.length === 0) {
        return changedAttributes;
    }
    log('There are pending changes in your collection deployment');
    drawTable(changing.map((change) => {
        return { Key: change.key, Action: change.action, Reason: change.reason, };
    }));

```

templates/cli/lib/commands/push.js.twig
```diff
@@ -140,6 +143,39 @@ const awaitPools = {
...
+    deleteAttributes: async (databaseId, collectionId, attributeKeys, iteration = 1) => {
+        if (iteration > pollMaxDebounces) {
+            return false;
+        }
+
+        let steps = Math.max(1, Math.ceil(attributeKeys.length / STEP_SIZE));
+        if (steps > 1 && iteration === 1) {
+            pollMaxDebounces *= steps;
+
+            log('Found a large number of deleting attributes, increasing timeout to ' + (pollMaxDebounces * POLL_DEBOUNCE / 1000 / 60) + ' minutes')
```
christyjacob4, line 155:
```suggestion
            log('Found a large number of attributes to be deleted. Increasing timeout to ' + (pollMaxDebounces * POLL_DEBOUNCE / 1000 / 60) + ' minutes')
```

templates/cli/lib/commands/push.js.twig
```diff
@@ -645,9 +681,100 @@ const createAttribute = async (databaseId, collectionId, attribute) => {
...
 
+    await databasesDeleteAttribute({
+        databaseId: collection['databaseId'],
+        collectionId: collection['$id'],
+        key: attribute.key,
+        parseOutput: false
+    });
+}
+
+const deepSimilar = (remote, local, collection) => {
```
christyjacob4, line 695: Can we come up with a better name for this function ? It is not clear from its name what it does

templates/cli/lib/commands/push.js.twig
```diff
@@ -645,9 +681,100 @@ const createAttribute = async (databaseId, collectionId, attribute) => {
...
+            reason += `${bol}${key} changed from ${chalk.red(remote[key])} to ${chalk.green(local[key])}`;
+        }
+    }
+
+    return reason === '' ? undefined : { key: keyName, attribute: remote, reason, action };
+}
+
+const findMatch = (attribute, attributes) => attributes.find((attr) => attr.key === attribute.key);
+
+const mapChangeAttribute = (attribute, collection, isAdding) => {
```
christyjacob4, line 719: Same here. can we have a better name for this function

templates/cli/lib/commands/push.js.twig
```diff
@@ -645,9 +681,100 @@ const createAttribute = async (databaseId, collectionId, attribute) => {
...
+    return {
+        key: `${chalk.yellow(attribute.key)} in ${collection.name} (${collection['$id']})`,
+        attribute,
+        reason: isAdding ? 'Field doesn\'t exist on the remote server' : 'Field doesn\'t exist in appwrite.json file',
+        action: isAdding ? chalk.green('adding') : chalk.red('deleting')
+    };
+
+};
+
+const updatedList = async (remoteAttributes, localAttributes, collection) => {
```
christyjacob4, line 729: Same here. can we have a better name for this function

templates/cli/lib/commands/push.js.twig
```diff
@@ -645,9 +681,100 @@ const createAttribute = async (databaseId, collectionId, attribute) => {
...
+    if (answers.changes.toLowerCase() !== 'yes') {
+        return changedAttributes;
+    }
+
+    if (conflicts.length > 0) {
+        changedAttributes = conflicts.map((change) => change.attribute);
+
+        await Promise.all(changedAttributes.map((changed) => deleteAttribute(collection, changed)));
+
+        remoteAttributes = remoteAttributes.filter((attribute) => !findMatch(attribute, changedAttributes))
```
christyjacob4, line 757:
```suggestion
        changedAttributes = conflicts.map((change) => change.attribute);
        await Promise.all(changedAttributes.map((changed) => deleteAttribute(collection, changed)));
        remoteAttributes = remoteAttributes.filter((attribute) => !findMatch(attribute, changedAttributes))
```

templates/cli/lib/commands/push.js.twig
```diff
@@ -645,9 +681,100 @@ const createAttribute = async (databaseId, collectionId, attribute) => {
...
+    }
+
+    const deletingAttributes = deleting.map((change) => change.attribute);
+
+    await Promise.all(deletingAttributes.map((attribute) => deleteAttribute(collection, attribute)));
+
+    const attributeKeys = [...remoteAttributes.map(attribute => attribute.key), ...deletingAttributes.map(attribute => attribute.key)]
```
christyjacob4, line 764:
```suggestion
    const deletingAttributes = deleting.map((change) => change.attribute);
    await Promise.all(deletingAttributes.map((attribute) => deleteAttribute(collection, attribute)));
    const attributeKeys = [...remoteAttributes.map(attribute => attribute.key), ...deletingAttributes.map(attribute => attribute.key)]
```

christyjacob4 (changes requested):
Please address the comments

templates/cli/lib/commands/push.js.twig
byawitz: This is just for the local flow as the collection is no longer saved to the `appwrite.json` file

christyjacob4 (approved):

PR #856, opened 2024-05-24
refactor(cli): Changing commands from singular to plural when applicable

byawitz:
## What does this PR do?

Changing command from singular to plural, for example: `appwrite pull collection` to `appwrite pull collections` .
CommanderJS will show errors like this automatically
```sh
error: unknown command 'collection'
(Did you mean collections?)
```
The change doesn't include `appwrite init` as this action is singular.

## Test Plan
image

### DIFF
--- a/templates/cli/lib/commands/push.js.twig
+++ b/templates/cli/lib/commands/push.js.twig
@@ -1117,8 +1117,10 @@ const pushMessagingTopic = async ({ all, yes } = {}) => {
 const push = new Command("push")
     .alias('deploy')
     .description(commandDescriptions['push'])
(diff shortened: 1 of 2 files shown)


templates/cli/lib/commands/push.js.twig
```diff
@@ -1127,7 +1129,7 @@ push
     .action(actionRunner(pushProject));
 
 push
-    .command("function")
+    .command("functions")
```
christyjacob4, line 1132: remember to add alias('function') as well

christyjacob4 (approved):

PR #860, opened 2024-05-28
feat(cli): Beta CLI releases

byawitz:
## What does this PR do?

Releases a new `beta` version for any new PR related to Appwrite CLI. 
You can see this command example
```
npx appwrite-cli-beta@0.16.0dd61a6df0ff32ed10e1eb539d53185c16c337140 --help
```
You'll notice change in the `json` flag allowing the short `-j` version as well.

### DIFF
--- a/.github/workflows/cli-beta-publish.yaml
+++ b/.github/workflows/cli-beta-publish.yaml
@@ -1,0 +1,47 @@
+name: CLI beta Publish
+on:
+  pull_request:
+    paths:
+      - 'templates/cli/**.twig'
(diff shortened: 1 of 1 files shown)


.github/workflows/cli-beta-publish.yaml
```diff
@@ -0,0 +1,47 @@
+name: CLI beta Publish
+on:
+  pull_request:
```
christyjacob4, line 3: I think you should do on `push` else it might only trigger the first time when you create a PR

christyjacob4 (approved):

PR #862, opened 2024-05-29
Feat: CLI Local development

Meldiron:
## What does this PR do?

Implements local development in `appwrite run function`.

## Test Plan

- [x] Manual QA

## Related PRs and Issues

x

Yes


templates/cli/lib/parser.js.twig
```diff
@@ -203,6 +203,7 @@ const commandDescriptions = {
     "databases": `The databases command allows you to create structured collections of documents, query and filter lists of documents.`,
     "init": `The init command provides a convenient wrapper for creating and initializing project, functions, collections, buckets, teams and messaging in Appwrite.`,
     "push": `The push command provides a convenient wrapper for pushing your functions, collections, buckets, teams and messaging.`,
+    "run": `The dev command allows you to run project locally to allow easy development and quick debugging.`,
```
christyjacob4, line 206:
```suggestion
    "run": `The run command allows you to run project locally to allow easy development and quick debugging.`,
```
Also we need to discuss if this needs to be `run` or `dev` as we need to keep some future products in mind as well

templates/cli/lib/commands/run.js.twig
```diff
@@ -0,0 +1,599 @@
...
+            // TODO: Once we have endpoint for this, use it
+            scopes: ["sessions.write","users.read","users.write","teams.read","teams.write","databases.read","databases.write","collections.read","collections.write","attributes.read","attributes.write","indexes.read","indexes.write","documents.read","documents.write","files.read","files.write","buckets.read","buckets.write","functions.read","functions.write","execution.read","execution.write","locale.read","avatars.read","health.read","providers.read","providers.write","messages.read","messages.write","topics.read","topics.write","subscribers.read","subscribers.write","targets.read","targets.write","rules.read","rules.write","migrations.read","migrations.write","vcs.read","vcs.write","assistant.read"],
+            duration: 60*60,
+            parseOutput: false
+        });
+        this.functionJwt = functionResponse.jwt;
+    }
+};
+
+const Queue = {
```
christyjacob4, line 150: can we move this Queue, JWTManager, systemtools etc to a folder lib/emulation ?

templates/cli/lib/commands/run.js.twig
```diff
@@ -0,0 +1,599 @@
...
+            fs.rmSync(appwritePath, { recursive: true, force: true });
+        }
+
+        const tempPath = path.join(process.cwd(), func.path, 'code.tar.gz');
+        if (fs.existsSync(tempPath)) {
+            fs.rmSync(tempPath, { force: true });
+        }
+    }
+}
+
```
christyjacob4, line 350: And lets move all the docker related functions to lib/emulation/docker.js

templates/cli/lib/questions.js.twig
```diff
@@ -768,7 +772,66 @@ const questionsMfaChallenge = [
...
+            })
+            return choices;
+        }
+    },
+    {
+        type: "input",
+        name: "port",
+        message: 'Which port would you like function to listen on?',
+        default: async () => {
+            let port = 3000;
```
christyjacob4, line 800: we should only show this prompt when the port 3000 is not available right? 
you can use the `when` condition of inquirer to check this

templates/cli/lib/questions.js.twig
```diff
@@ -768,7 +772,66 @@ const questionsMfaChallenge = [
...
+            }
+
+            return 3000;
+        },
+        validate: function(value) {
+            const done = this.async();
+
+            (async () => {
+                if (typeof value !== 'number' && isNaN(+value)) {
+                    throw Error(`You need to provide a number.`);
```
christyjacob4, line 817:
```suggestion
                    throw Error(`Port needs to be a valid integer between 1 and 65536`);
```
Even though ports 1 to 65536 are available in theory, some ranges are reserved for specific use cases. So lets research what is the allowed port range and modify the error message and condition to reflect it

templates/cli/lib/questions.js.twig
```diff
@@ -768,7 +772,66 @@ const questionsMfaChallenge = [
...
+                }
+
+                value = +value;
 
+                const taken = await isPortTaken(value);
+
+                if(taken) {
+                    throw Error(`Port ${value} is taken. Pick another one.`);
```
christyjacob4, line 825:
```suggestion
                    throw Error(`Port ${value} is in use by another process. Pick a different one.`);
```

templates/cli/lib/commands/run.js.twig
```diff
@@ -0,0 +1,599 @@
...
+    .action(actionRunner(async (_options, command) => {
+        command.help();
+    }));
+
+run
+    .command("function")
+    .alias("functions")
+    .description("Run functions in the current directory.")
+    .option(`--functionId <functionId>`, `Function ID`)
+    .option(`--port <port>`, `Local port`)
```
christyjacob4, line 591: lets fix a default port so user doesnt need to enter it

templates/cli/lib/commands/run.js.twig
```diff
@@ -0,0 +1,599 @@
...
+    }
+
+    if(!port) {
+        const answers = await inquirer.prompt(questionsRunFunctions[1]);
+        port = answers.port;
+    }
+
+    // Configuration: Engine
+    if(!systemHasCommand('docker')) {
+        return error("Please install Docker first: https://docs.docker.com/engine/install/");
```
christyjacob4, line 392:
```suggestion
        return error("Docker Engine is required for local development. Please install Docker using: https://docs.docker.com/engine/install/");
```

templates/cli/lib/commands/run.js.twig
```diff
@@ -0,0 +1,599 @@
...
+
+    if(isNaN(port)) {
+        port = null;
+    }
+
+    if(port) {
+        const taken = await isPortTaken(port);
+
+        if(taken) {
+            log(`Port ${port} is already used.`);
```
christyjacob4, line 380:
```suggestion
            log(`Port ${port} is already in use by another process.`);
```

templates/cli/lib/commands/run.js.twig
```diff
@@ -0,0 +1,599 @@
...
+    const settings = {
+        runtime: func.runtime,
+        entrypoint: func.entrypoint,
+        path: func.path,
+        commands: func.commands,
+    };
+
+    log("Local function configuration:");
+    drawTable([settings]);
+    log('If you wish to change local settings, update appwrite.json file and rerun the command. To deploy the function, run: appwrite push function');
```
christyjacob4, line 405:
```suggestion
    log('If you wish to change your local settings, update the appwrite.json file and rerun the `appwrite run` command. ');
```
I dont think the we need to mention about push here

templates/cli/lib/questions.js.twig
Meldiron: Well we could automatically also pick 3001 because its available. Should we not ask this question at all, and just tell them to use `--port` if they want to change it?

templates/cli/lib/questions.js.twig
Meldiron: I think 1024 and upwards should be allow for custom use, below are considered system ports

templates/cli/lib/commands/run.js.twig
Meldiron: It's optional param. If not provided ,question is asked (with default 3000 as placeholder)

templates/cli/lib/commands/run.js.twig
```diff
@@ -0,0 +1,283 @@
...
+        if(taken) {
+            error(`Port ${port} is already in use by another process.`);
+            return;
+        }
+    }
+
+    if(!port) {
+        let portFound = fale;
+        port = 3000;
+        while(port < 3100) {
```
christyjacob4, line 60: while very rare, why limit the search to 100 ports? Why not set it to something higher like 6000 ?

templates/cli/lib/commands/run.js.twig
```diff
@@ -0,0 +1,283 @@
...
+            if(!taken) {
+                portFound = true;
+                break;
+            }
+
+            port++;
+        }
+
+        if(!portFound) {
+            error('Could not auto-suggest an available port. Please configure port with `appwrite run --port YOUR_PORT` command.');
```
christyjacob4, line 71:
```suggestion
            error('Could not find an available port. Please select a port with `appwrite run --port YOUR_PORT` command.');
```

src/SDK/Language/CLI.php
```diff
@@ -245,6 +245,11 @@ public function getFiles(): array
                 'destination'   => 'lib/commands/push.js',
                 'template'      => 'cli/lib/commands/push.js.twig',
             ],
+            [
+                'scope'         => 'default',
+                'destination'   => 'lib/commands/run.js',
+                'template'      => 'cli/lib/commands/run.js.twig',
+            ],
```
christyjacob4, line 252: You need to add the new template files here

templates/cli/lib/commands/run.js.twig
Meldiron: No idea how long that might take. And as you said, very rare. I wouldn' worry too much until we get some complains.

src/SDK/Language/CLI.php
Meldiron: Thanks.

christyjacob4 (approved):

PR #865, opened 2024-05-29
Feat push pull all

byawitz:
## What does this PR do?
Adds the ability to pull resources at once 

## Test Plan

https://github.com/appwrite/sdk-generator/assets/316103/8718ff60-548a-464b-8156-2f360491f2da

### DIFF
--- a/templates/cli/lib/commands/pull.js.twig
+++ b/templates/cli/lib/commands/pull.js.twig
@@ -29,12 +29,13 @@ const pullResources = async () => {
         }
     } else {
         const answers = await inquirer.prompt(questionsPullResources[0]);
-        answers.resources.forEach((resource) => {
+
+        for (let resource of answers.resources) {
             const action = actions[resource];
             if (action !== undefined) {
-                action();
+                await action();
             }
-        })
+        }
     }
 };
 
@@ -112,10 +113,9 @@ const pullCollection = async () => {
 
     if (databases.length === 0) {
         if (cliConfig.all) {
-            databases = (await paginate(databasesList, { parseOutput: false }, 100, 'databases')).databases.map(database=>database.$id);
-        } else{
+            databases = (await paginate(databasesList, { parseOutput: false }, 100, 'databases')).databases.map(database => database.$id);
+        } else {
             databases = (await inquirer.prompt(questionsPullCollection)).databases;
-    }
         }
     }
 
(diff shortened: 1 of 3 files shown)


templates/cli/lib/commands/pull.js.twig
```diff
@@ -163,6 +193,14 @@ const pull = new Command("pull")
...
+    .command("all")
+    .description("Push all resource.")
+    .option(`--all`, `Flag to pull all functions`)
+    .option(`--yes`, `Flag to confirm all warnings`)
+    .action(actionRunner(pullResources));
+
+
 pull
     .command("project")
     .description("Pulling your {{ spec.title|caseUcfirst }} project name, services and auth settings")
```
christyjacob4, line 206: I would also change all these messages to `Pull` instead of `Pulling`

templates/cli/lib/commands/pull.js.twig
```diff
@@ -163,6 +193,14 @@ const pull = new Command("pull")
         helpWidth: process.stdout.columns || 80
     });
 
+pull
+    .command("all")
+    .description("Push all resource.")
+    .option(`--all`, `Flag to pull all functions`)
+    .option(`--yes`, `Flag to confirm all warnings`)
+    .action(actionRunner(pullResources));
+
```
christyjacob4, line 202: Not sure if this should be a new command. Have you considered using just an option `--all` to the base pull command itself? So it would be 

```sh
appwrite pull project
appwrite pull functions
appwrite pull --all
``` 
??

byawitz: The reason this was done, when using `appwrite pull --all` we won't get the error message like this for example
```sh
> appwrite pull function
error: unknown command 'function`
(Did you mean functions?)

```
As everything is the first command

templates/cli/lib/commands/pull.js.twig
```diff
@@ -163,6 +193,14 @@ const pull = new Command("pull")
         helpWidth: process.stdout.columns || 80
     });
 
+pull
+    .command("all")
+    .description("Push all resource.")
+    .option(`--all`, `Flag to pull all functions`)
```
christyjacob4, line 199:
```suggestion
    .option(`--all`, `Flag to pull all resources`)
```

templates/cli/lib/commands/pull.js.twig
christyjacob4: Alright 👍

templates/cli/lib/commands/pull.js.twig
byawitz: This `options` removed altogether, as it's handle globally

christyjacob4 (approved):

PR #868, opened 2024-05-31
Feat arg project init

byawitz:
## What does this PR do?

Allowing project creation using a single command

## Test Plan
image


templates/cli/lib/commands/init.js.twig
```diff
@@ -40,12 +40,35 @@ const initProject = async () => {
...
+                answers.start = 'new';
+                answers.id = answers.project.id;
+                answers.project = answers.project.name;
+            } else {
+                throw e;
+            }
+        }
+
+    } else {
+        answers = await inquirer.prompt(questionsInitProject)
```
Meldiron, line 65: Hmm. What if we did different approach?
We start with empty "answers" object.. We try to fill ith with params from CMD, only using what is provided. We then take `questionsInitProject` and filter out questions that we already have answer for. Then, we show remaining questions and  fill them into our answers object. If all answers were already given with params, there is no need to ask questions all-together

templates/cli/lib/commands/init.js.twig
```diff
@@ -40,12 +40,35 @@ const initProject = async () => {
         log('You must login first')
         await loginCommand();
     }
+    let answers = {};
 
-    const answers = await inquirer.prompt(questionsInitProject)
-    if (answers.override === false) {
-        process.exit(1)
+    if (organizationId && projectId && projectName) {
```
Meldiron, line 45: I think we should add ability to set organization ID, and nothing else. That way you are still asked questions, but not about organization. I think this would make sense for each param.

templates/cli/lib/commands/init.js.twig
```diff
@@ -23,7 +23,7 @@ const { accountGet } = require("./account");
 const { loginCommand } = require("./generic");
 const { sdkForConsole } = require("../sdks");
 
-const initProject = async () => {
+const initProject = async ({ organizationId, projectId, projectName } = {}) => {
```
Meldiron, line 26: We are no longer setting project name in this method, I believe. It's a separate method `appwrite pull proejct`

templates/cli/lib/commands/init.js.twig
byawitz: Addressed

templates/cli/lib/commands/init.js.twig
byawitz: Addressed

templates/cli/lib/questions.js.twig
```diff
@@ -350,7 +357,22 @@ const questionsCreateCollection = [
             }
 
             return choices;
-        }
+        },
+        when: (answers) => (answers.method ?? '').toLowerCase() === 'existing'
+    },
+    {
+        type: "input",
+        name: "database_name",
```
christyjacob4, line 365: we should use camel case here

templates/cli/lib/questions.js.twig
```diff
@@ -350,7 +357,22 @@ const questionsCreateCollection = [
...
+        message: "What would you like to name your database?",
+        default: "My Awesome Database",
+        when: (answers) => (answers.method ?? '').toLowerCase() !== 'existing'
+    },
+    {
+        type: "input",
+        name: "database_id",
```
christyjacob4, line 372: we should use camel case here

templates/cli/lib/utils.js.twig
```diff
@@ -67,7 +67,13 @@ function showConsoleLink(serviceName, action, open, ...ids) {
...
     if (open) {
         const start = (process.platform == 'darwin' ? 'open' : process.platform == 'win32' ? 'start' : 'xdg-open');
-        cp.exec(`${start} ${url}`);
+
+        cp.exec(`${start} ${url}`, (err, stdout, stderr) => {
+            if (err !== null) {
+                console.log('\n');
+                error('Opening in default browser. ' + err)
+            }
+        });
```
christyjacob4, line 76: Do we need this --open option? I think it might cause more problems than we are trying to solve. Does `xdg-open` work on all linux distros? 

I think we can save this trouble altogether by removing the `open` param. We can just display the link with the --console flag and people can just click it.

christyjacob4 (approved):

PR #874, opened 2024-06-07
Feat general improvements

byawitz:
## What does this PR do?

- Function pushing summary show good and bad results
- Tip to use `--verbose` or `--report`
- Print an error when `--console --open` didn't work
- Print endpoint in `whoami`
- Fix `Check cannot destrucutre selfHosted of undefined` bug
- Delete halfway logged-in accounts.

### DIFF
--- a/templates/cli/lib/client.js.twig
+++ b/templates/cli/lib/client.js.twig
@@ -4,10 +4,11 @@ const { fetch, FormData, Agent } = require("undici");
 const JSONbig = require("json-bigint")({ storeAsString: false });
 const {{spec.title | caseUcfirst}}Exception = require("./exception.js");
 const { globalConfig } = require("./config.js");
+const {log} = require('./parser');
 
 class Client {
   CHUNK_SIZE = 5*1024*1024; // 5MB
-  
+
   constructor() {
     this.endpoint = '{{spec.endpoint}}';
     this.headers = {
@@ -143,6 +144,13 @@ class Client {
         json = JSON.parse(text);
       } catch (error) {
         throw new {{spec.title | caseUcfirst}}Exception(text, response.status, "", text);
+      }
+
+      if(path !== '/account' && json.code === 401 && json.type ==='user_more_factors_required'){
+        log('Unusable account found, removing...')
+        const current = globalConfig.getCurrentLogin();
+        globalConfig.setCurrentLogin('');
+        globalConfig.removeLogin(current);
       }
       throw new {{spec.title | caseUcfirst}}Exception(json.message, json.code, json.type, json);
     }
--- a/templates/cli/lib/commands/init.js.twig
+++ b/templates/cli/lib/commands/init.js.twig
@@ -23,7 +23,7 @@ const { accountGet } = require("./account");
 const { loginCommand } = require("./generic");
 const { sdkForConsole } = require("../sdks");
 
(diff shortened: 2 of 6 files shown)


templates/cli/lib/utils.js.twig
```diff
@@ -67,7 +67,13 @@ function showConsoleLink(serviceName, action, open, ...ids) {
 
     if (open) {
         const start = (process.platform == 'darwin' ? 'open' : process.platform == 'win32' ? 'start' : 'xdg-open');
-        cp.exec(`${start} ${url}`);
+
+        cp.exec(`${start} ${url}`, (err, stdout, stderr) => {
+            if (err !== null) {
+                console.log('\n');
+                error('Opening in default browser. ' + err)
```
christyjacob4, line 74: what exactly are we doing here?why are we saying opening in default browser and appending the error?

templates/cli/lib/commands/init.js.twig
```diff
@@ -269,6 +269,7 @@ const init = new Command("init")
     .option("--organizationId <organizationId>", "{{ spec.title|caseUcfirst }} organization ID")
     .option("--projectId <projectId>", "{{ spec.title|caseUcfirst }} project ID")
     .option("--projectName <projectName>", "{{ spec.title|caseUcfirst }} project ID")
+    .option(`-sh, --self-hosted`, `Flag for enabling custom endpoint for self hosted instances for non-logged in users`)
```
christyjacob4, line 272: what are we trying to achieve here?

templates/cli/lib/client.js.twig
```diff
@@ -4,10 +4,11 @@ const { fetch, FormData, Agent } = require("undici");
 const JSONbig = require("json-bigint")({ storeAsString: false });
 const {{spec.title | caseUcfirst}}Exception = require("./exception.js");
 const { globalConfig } = require("./config.js");
+const {log} = require('./parser');
```
christyjacob4, line 7:
```suggestion
const { log } = require('./parser');
```

templates/cli/lib/client.js.twig
```diff
@@ -144,6 +145,13 @@ class Client {
       } catch (error) {
         throw new {{spec.title | caseUcfirst}}Exception(text, response.status, "", text);
       }
+
+      if(path !== '/account' && json.code === 401 && json.type ==='user_more_factors_required'){
```
christyjacob4, line 149:
```suggestion
      if (path !== '/account' && json.code === 401 && json.type ==='user_more_factors_required') {
```

templates/cli/lib/utils.js.twig
byawitz: When the browser failed to open the link for any reason the `error` function will preint
Error: Opening in default browser. ERROR MESSAGE

templates/cli/lib/commands/init.js.twig
byawitz: In case the user is not logged in, the project init will prompt the user to login,
By default, all logins are  targeted at the cloud, this will give the user the option to add an endpoint

christyjacob4: No lets not do this.. There is already a dedicated command for setting endpoint with appwrite client. 
We should make sure every command has one job and one job only

christyjacob4 (approved):

PR #877, opened 2024-06-11
Feat function init improvments

byawitz:
## What does this PR do?

- Fetching function templates using API.
- When the user reached API `rate-limit` then failing back to current folder-based fetching

### DIFF
--- a/templates/cli/lib/commands/init.js.twig
+++ b/templates/cli/lib/commands/init.js.twig
@@ -174,7 +174,7 @@ const initFunction = async () => {
     fs.mkdirSync(functionDir, "777");
     fs.mkdirSync(templatesDir, "777");
     const repo = "https://github.com/{{ sdk.gitUserName }}/templates";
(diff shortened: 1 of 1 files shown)


templates/cli/lib/commands/init.js.twig
```diff
@@ -172,10 +173,26 @@ const initFunction = async () => {
 
     fs.mkdirSync(functionDir, "777");
     fs.mkdirSync(templatesDir, "777");
+    const repo = "https://github.com/{{ sdk.gitUserName }}/templates";
+    const api = `https://api.github.com/repos/appwrite/templates/contents/${answers.runtime.name}`
```
christyjacob4, line 177:
```suggestion
    const api = `https://api.github.com/repos/{{ sdk.gitUserName }}/templates/contents/${answers.runtime.name}`
```

christyjacob4 (approved):

Issue #883, opened 2024-06-19
🐛 Bug Report: Fix Enum Casings


ItzNotABug:
### 👟 Reproduction steps

Swift -
![image](https://github.com/appwrite/sdk-generator/assets/20625965/fa4bda5d-7761-4d9b-a9fa-e12e66c52435)

Dart -
<img width="571" alt="Screenshot 2024-06-19 at 12 14 39 PM" src="https://github.com/appwrite/sdk-generator/assets/20625965/4bd8cd67-b370-4e25-a58f-fd9161d4854f">

---

I am open to work on this issue.

### 👍 Expected behavior

The enum casing should be `lowerCamelCase`.

### 👎 Actual Behavior

Enum casing uses `camelCase`, which isn't quite right here.

### 🎲 Appwrite version

Different version (specify in environment)

### 💻 Operating system

MacOS

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

PR #885, opened 2024-06-19
Fix Enum Casings For Swift & Dart

ItzNotABug:
## What does this PR do?

Enums in Swift & Dart should be in a `lowerCamelCase`.

## Test Plan

N/A.

## Related PRs and Issues

- #883.

Yes.


src/SDK/Language.php
```diff
@@ -106,6 +106,22 @@ protected function toCamelCase($str): string
...
+        // Remove accents and other residual non-ASCII characters
+        $str = preg_replace('/\p{M}/u', '', $str);
+
+        // Replace non-alphanumeric characters with a space
+        $str = preg_replace('/[^a-zA-Z0-9]+/', ' ', $str);
+        $str = ucwords(strtolower($str));
+        $str = lcfirst(str_replace(' ', '', $str));
+
+        return $str;
+    }
```
abnegate, line 123: I think we just need the `\strtolower` added to the existing `toCamelCase` function before the `\ucwords`.

ItzNotABug: Done.

abnegate (commented):
Let's add a value to the mock enum endpoint with all upper case and test against it to make sure the templating is correct for all languages

abnegate (approved):

PR #889, opened 2024-06-26
fix(cli): Apply non destructive changes

byawitz:
## What does this PR do?
Apply non-destructive changes

image

### DIFF
--- a/templates/cli/lib/commands/push.js.twig
+++ b/templates/cli/lib/commands/push.js.twig
@@ -25,6 +25,16 @@ const {
     databasesCreateUrlAttribute,
     databasesCreateIpAttribute,
     databasesCreateEnumAttribute,
+    databasesUpdateBooleanAttribute,
+    databasesUpdateStringAttribute,
+    databasesUpdateIntegerAttribute,
+    databasesUpdateFloatAttribute,
+    databasesUpdateEmailAttribute,
+    databasesUpdateDatetimeAttribute,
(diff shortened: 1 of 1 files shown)


templates/cli/lib/commands/push.js.twig
```diff
@@ -1063,7 +1212,7 @@ const pushCollection = async ({ returnOnZero } = { returnOnZero: false }) => {
             attributes = await attributesToCreate(collection.remoteVersion.attributes, collection.attributes, collection);
 
             if (Array.isArray(attributes) && attributes.length <= 0) {
-                log(`No changes has been detected. Skipping ${collection.name} ( ${collection['$id']} )`);
+                log(`No destructive changes has been detected. Skipping ${collection.name} ( ${collection['$id']} )`);
```
Meldiron, line 1215: Do we have some non-destructive ones?

Meldiron: If so, we shouldnt skip them. we should do them.
Also, in databases specifically, I would ALWAYS ask for confirmation, so they always see the table and can decide if all those changes should be applied - even if non-destructire. It's like having dry-run.

byawitz: The skipping is on the destructive changes, the non-destructive changes are being applied (as max)

byawitz: image

PR #908, opened 2024-07-16
feat(cli): Custom debounce limit

byawitz:
## What does this PR do?

(Provide a description of what this PR does.)

## Test Plan

## Related PRs and Issues

* https://github.com/appwrite/sdk-for-cli/issues/135

(Write your answer here.)

### DIFF
--- a/templates/cli/lib/commands/push.js.twig
+++ b/templates/cli/lib/commands/push.js.twig
@@ -1427,7 +1427,7 @@ push
     .command("collection")
     .alias("collections")
     .description("Push collections in the current project.")
-    .option(`-d, --debounce <maxDebounces>`, `Max number of debounces, default: 30.`)
+    .option(`-a, --attempts <numberOfAttempts>`, `Max number of attempts before timing out collection attribute operations, default: 30.`)
     .action(actionRunner(pushCollection));
 
 push
--- a/templates/cli/lib/commands/push.js.twig
+++ b/templates/cli/lib/commands/push.js.twig
@@ -68,8 +68,9 @@ const { checkDeployConditions } = require('../utils');
 const STEP_SIZE = 100;      // Resources
 const POLL_DEBOUNCE = 2000; // Milliseconds
 const POLL_MAX_DEBOUNCE = 1800; // Times of POLL_DEBOUNCE (1 hour)
+const POLL_DEFAULT_VALUE = 30;
 
-let pollMaxDebounces = 30;
+let pollMaxDebounces = POLL_DEFAULT_VALUE;
 
 const changeableKeys = ['status', 'required', 'xdefault', 'elements', 'min', 'max', 'default', 'error'];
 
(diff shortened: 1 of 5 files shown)


templates/cli/lib/commands/push.js.twig
```diff
@@ -90,11 +91,13 @@ const awaitPools = {
...
-        if (steps > 1 && iteration === 1) {
-            pollMaxDebounces *= steps;
+        if (pollMaxDebounces === POLL_DEFAULT_VALUE) {
+            let steps = Math.max(1, Math.ceil(total / STEP_SIZE));
+            if (steps > 1 && iteration === 1) {
+                pollMaxDebounces *= steps;
 
-            log('Found a large number of attributes, increasing timeout to ' + (pollMaxDebounces * POLL_DEBOUNCE / 1000 / 60) + ' minutes')
+                log('Found a large number of attributes, increasing timeout to ' + (pollMaxDebounces * POLL_DEBOUNCE / 1000 / 60) + ' minutes')
+            }
```
stnguyen90, line 100: So we're only going to log if the developer doesn't specify a custom debounce? Why is that?

templates/cli/lib/commands/push.js.twig
```diff
@@ -1410,6 +1427,7 @@ push
     .command("collection")
     .alias("collections")
     .description("Push collections in the current project.")
+    .option(`-d, --debounce <maxDebounces>`, `Max number of debounces, default: 30.`)
```
stnguyen90, line 1430: This description is pretty confusing for people because they don't know how the push code works. How can we rephrase this to be clearer?

templates/cli/lib/commands/push.js.twig
byawitz: If the developer specified the debounce limit, it will be used, and then it doesn't make sense to notify about it as it will be used anyway.

byawitz: If the developer specified the debounces limit, then there's no need to check if we're increasing the timeout as we won't increase it but set it to the manually one

templates/cli/lib/commands/push.js.twig
byawitz: Is this better?

stnguyen90 (approved):

templates/cli/lib/commands/push.js.twig
```diff
@@ -1410,6 +1427,7 @@ push
     .command("collection")
     .alias("collections")
     .description("Push collections in the current project.")
+    .option(`-a, --attempts <numberOfAttempts>`, `Max number of attempts before timing out collection attribute operations, default: 30.`)
```
stnguyen90, line 1430: attempts is definitely much better than debounce. The only thing is it's not just for attribute operations; it's for indexes too. So, maybe just:

Max number of attempts before timing out

christyjacob4 (approved):

PR #914, opened 2024-07-20
CLI: local variables, speed-up build

Meldiron:
## What does this PR do?

- Ensures build process doesn't see any ignored files
- Print env vars during function start, and take config from .env
- Fix render of arrays and objects in table

## Test Plan

- [x] Manual QA

## Related PRs and Issues

x

Yes


templates/cli/lib/commands/run.js.twig
```diff
@@ -110,7 +112,9 @@ const runFunction = async ({ port, functionId, noVariables, noReload, userId } =
         fs.writeFileSync(errorsPath, '');
     }
 
+    const userVariables = {};
```
engibeeer, line 115: I can't see the usability of this variable

engibeeer (approved):

TorstenDittmann (approved):
the file ignoring part is a bit weird to me, we should respect .gitignore or so

templates/cli/base/params.twig
```diff
@@ -17,6 +17,7 @@
     const func = localConfig.getFunction(functionId);
 
     ignorer.add('.appwrite');
+    ignorer.add('.DS_Store');
```
eldadfux, line 20: Let's leave this up for the dev to decide similar to how Git does.

PR #920, opened 2024-07-27
Fix incorrect parsing/rounding of `BigInt`

ItzNotABug:
## What does this PR do?

Fixes the issue where the big integers are incorrect due to precision loss.

Fixes https://github.com/appwrite/console/issues/1473

## Test Plan

## Related PRs and Issues

* https://github.com/appwrite/console/issues/1473

Yes


templates/node/package.json.twig
```diff
@@ -42,12 +42,14 @@
...
   "devDependencies": {
     "@types/node": "20.11.25",
+    "@types/json-bigint": "^1.0.4",
     "tsup": "7.2.0",
     "esbuild-plugin-file-path-extensions": "^2.0.0",
     "tslib": "2.6.2",
     "typescript": "5.4.2"
   },
   "dependencies": {
+    "json-bigint": "^1.0.0",
```
loks0n, line 52: Do we need a depencdency? is it possible without?

loks0n:
Can you share more about the issue? What situation does it occur?

ItzNotABug:
> Can you share more about the issue? What situation does it occur?

The values returned on console are different for min/max due to how PHP and JS round off the int/big int. Values should be -
`-9223372036854775808` & `9223372036854775807` but JS rounds off to `-9223372036854776000` and `-9223372036854776000`. This causes issues during updates iirc.

templates/node/package.json.twig
ItzNotABug: Yup. I don't think there's a built-in logic for managing big-ints.

ivanskodje:
> Fixes the issue where the big integers are incorrect due to precision loss.

Does these changes affect how the Console repo parses the number it receives from the backend? I ask, because in Console we are already receiving the correct number from the backend via JSON - but it is only after it parses it into a `number` object it gets rounded to an incorrect amount causing this issue.

Ref: https://github.com/appwrite/console/issues/1473

ItzNotABug:
@loks0n based on the new changes, we could either use that impl or might add a library to handle this scenario. `CLI` anyway uses the same library used previously to handle this scenario.

cc @TorstenDittmann.

PR #890, opened 2024-06-27
Fix variable names and logging in web and react-native docs templates #8328 

kunalArya1:
# Title: Fix Console Logging in Storage Client Example Templates

## Description:
This PR addresses an issue in the example templates for the Appwrite documentation where the console logging of API responses was incorrectly referencing the response variable instead of the result variable. This discrepancy could potentially confuse users following the examples.

## Changes Made:

Updated the console logging in the example templates for both the web and React Native clients to use the result variable consistently.

### Context:
The original templates incorrectly used response in the console.log statement, which does not align with the variable naming conventions in the example code provided. By changing this to result, we ensure clarity and accuracy in the displayed examples.

## Contributor Note:
This change aligns with the conventions observed throughout the documentation examples and improves the usability for developers integrating the Appwrite storage client in their applications.

### References:

Fixes: https://github.com/appwrite/appwrite/issues/8328

### Issue Link

* https://github.com/appwrite/appwrite/issues/8328)

This PR is ready for review. Your feedback is appreciated!

PR #925, opened 2024-07-30
Fix: Consistently Log ' result ' in API Call Template.

ParagGhatage:
### Summary
This pull request fixes the issue where the `console.log` statement inconsistently logged the `response` variable instead of the `result` variable in the API call template. This ensures that the output is always correct and avoids any confusion that might arise from logging an undefined `response` variable.

### Changes Made
- Updated the `console.log` statement to always log `result` instead of `response`.

### Reason for Changes
The previous implementation had an inconsistency in the variable being logged after an API call. Depending on the method type, the `console.log` statement sometimes attempted to log an undefined `response` variable, leading to potential confusion and errors. By consistently logging the `result` variable, the output remains accurate and predictable.

### Related  issues
- https://github.com/appwrite/appwrite/issues/8328#issue-2373904477

### Testing
- Verified that the `console.log` statement now correctly logs the `result` variable.
- Ensured that the change does not affect other parts of the code template.

### Impact
This change should have no negative impact on the existing functionality but will improve the clarity and correctness of the logging in the generated code template.

@stnguyen90

### DIFF
--- a/templates/react-native/docs/example.md.twig
+++ b/templates/react-native/docs/example.md.twig
@@ -22,5 +22,5 @@ const {{ service.name | caseCamel }} = new {{service.name | caseUcfirst}}(client
 {% endif %}
 
 {% if method.type != 'webAuth' %}
-console.log({% if method.type == 'location' %}result{% else %}response{% endif %});
+console.log(result);
 {% endif %}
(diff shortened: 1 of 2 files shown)


stnguyen90:
Closing as already addressed via https://github.com/appwrite/sdk-generator/pull/890

PR #930, opened 2024-08-05
feat(cli): Pull last deployment instead of local when available

byawitz:
## What does this PR do?
Try to fetch the last function deployment for local code pulling.
On any error, we will resort to the local value

## Test Plan
Locally

### DIFF
--- a/templates/cli/lib/commands/pull.js.twig
+++ b/templates/cli/lib/commands/pull.js.twig
@@ -6,7 +6,7 @@ const inquirer = require("inquirer");
 const { messagingListTopics } = require("./messaging");
 const { teamsList } = require("./teams");
 const { projectsGet } = require("./projects");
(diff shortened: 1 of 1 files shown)


templates/cli/lib/commands/pull.js.twig
```diff
@@ -108,10 +108,30 @@ const pullFunctions = async ({ code, withVariables }) => {
...
+                        parseOutput: false
+                    });
+
+                    if (fetchResponse['total'] > 0) {
+                        deploymentId = fetchResponse['deployments'][0]['$id'];
+                    }
+
+                } catch (e) {
+                    log("Can't get last deployment ID, using local value instead")
+                }
```
Meldiron, line 129: I dont think we need fallback. If there is no deployment in listDeployments, the active deployment will also never work

christyjacob4 (approved):

PR #940, opened 2024-08-14
feat: php multiparts

byawitz:
## What does this PR do?
Changes `createExecution` to send and parse multipart data 

Code example 👇
```php
<?php

use Appwrite\Client;
use Appwrite\Payload;
use Appwrite\Services\Functions;

$client = (new Client())
    ->setEndpoint('https://cloud.appwrite.io/v1') // Your API Endpoint
    ->setProject('&lt;YOUR_PROJECT_ID&gt;') // Your project ID
    ->setSession(''); // The user session to authenticate with

$functions = new Functions($client);

$result = $functions->createExecution(
    functionId: '<FUNCTION_ID>',
    body: Payload::fromJson([]), // optional
    async: false, // optional
    path: '<PATH>', // optional
    method: ExecutionMethod::GET(), // optional
    headers: '', // optional
    scheduledAt: '' // optional
);
```

## Test Plan

image


mock-server/app/http.php
```diff
@@ -317,12 +316,11 @@
     ->param('x', '', new Text(100), 'Sample string param')
     ->param('y', '', new Integer(true), 'Sample numeric param')
     ->param('z', null, new ArrayList(new Text(256), APP_LIMIT_ARRAY_PARAMS_SIZE), 'Sample array param')
-    ->param('file', [], new File(), 'Sample file param', skipValidation: true)
+    ->param('payload', [], new File(), 'Sample file param', skipValidation: true)
```
christyjacob4, line 319: this change is not required right? we can keep the param name as file itself?

tests/languages/php/test.php
```diff
@@ -118,6 +118,11 @@
 );
 echo $url . "\n";
 
+$response = $general->multipart();
+echo "{$response['x']}\n";
+$hash = md5($response['responseBody']->toBinary());
+echo "{$hash}\n";
+
```
christyjacob4, line 125: cool this is good, but this is an example of how we receive multipart data right? Don't we meed to have a test for  sending multipart as well ?

templates/php/base/params.twig
```diff
@@ -16,6 +20,9 @@
         {%~ for parameter in method.parameters.header %}
         $apiHeaders['{{ parameter.name }}'] = ${{ parameter.name | caseCamel | escapeKeyword }};
         {%~ endfor %}
+        {%~ if method.name | lower == "createexecution" %}
+        $apiHeaders['accept'] = 'multipart/form-data';
+        {%~ endif %}
```
christyjacob4, line 25: I think we should move this code to the client class, inside  the case for multi part 

Screenshot 2024-08-30 at 11 40 04 AM

templates/php/src/Payload.php.twig
```diff
@@ -0,0 +1,80 @@
...
+    {
+        $instance = new Payload();
+        $instance->path = null;
+        $instance->data = $data;
+        $instance->mimeType = $mimeType;
+        $instance->filename = $filename;
+        return $instance;
+    }
+
+    public static function fromJson(array $data) {
```
christyjacob4, line 52: missing return type. I think we can use `:self` right ?

templates/php/base/params.twig
christyjacob4: Or we should pass some sort of an attribute from the specs to make this check easier

templates/php/src/Payload.php.twig
```diff
@@ -0,0 +1,80 @@
...
+    {
+        return $this->mimeType;
+    }
+
+    public function getFilename(): ?string
+    {
+        return $this->filename;
+    }
+
+    public static function fromFile(string $path, ?string $mimeType = null, ?string $filename = null): Payload
```
christyjacob4, line 32:
```suggestion
    public static function fromFile(string $path, ?string $mimeType = null, ?string $filename = null): self
```

templates/php/src/Payload.php.twig
```diff
@@ -0,0 +1,80 @@
...
+    {
+        $instance = new Payload();
+        $instance->path = $path;
+        $instance->data = null;
+        $instance->mimeType = $mimeType;
+        $instance->filename = $filename;
+        return $instance;
+    }
+
+    public static function fromBinary(string $data, ?string $mimeType = null, ?string $filename = null): Payload
```
christyjacob4, line 42:
```suggestion
    public static function fromBinary(string $data, ?string $mimeType = null, ?string $filename = null): self
```

templates/php/src/Payload.php.twig
```diff
@@ -0,0 +1,80 @@
...
+        $instance->data = $data;
+        $instance->mimeType = $mimeType;
+        $instance->filename = $filename;
+        return $instance;
+    }
+
+    public static function fromJson(array $data) {
+        $instance = new Payload();
+        $instance->path = null;
+        $instance->data = json_encode($data);
```
christyjacob4, line 55: lets set the mime type to `application/json`

templates/php/src/Payload.php.twig
```diff
@@ -0,0 +1,80 @@
...
+        return $instance;
+    }
+
+    public static function fromString(string $data) {
```
christyjacob4, line 59: missing return type

templates/php/src/Payload.php.twig
```diff
@@ -0,0 +1,80 @@
...
+        $instance = new Payload();
+        $instance->path = null;
+        $instance->data = $data;
+        return $instance;
+    }
+
+    public function toBinary(): string
+    {
+        return $this->data;
```
christyjacob4, line 68: is binary represented as a string in PHP? is there no other data type like a buffer or uint8 array or something?

templates/php/src/Services/Service.php.twig
```diff
@@ -53,7 +53,7 @@ class {{ service.name | caseUcfirst }} extends Service
         );
 
         {{~ include('php/base/params.twig') -}}
-        {%~ if 'multipart/form-data' in method.consumes %}
+        {%~ if 'multipart/form-data' in method.consumes and method.name | lower != "createexecution" %}
```
christyjacob4, line 56: lets see if there is a better way to do this check

mock-server/app/http.php
loks0n: This a quick test to determine that we aren't using hardcoded checks against the `file` param name

tests/languages/php/test.php
loks0n: We already have a test for this, the `upload` function

templates/php/src/Client.php.twig
```diff
@@ -240,4 +246,35 @@ class Client
...
+            $matched = preg_match('/name="?(?<name>\w+)/s', $part, $matches);
+            if ($matched) {
+                array_shift($lines);
+                if(isset($lines[0]) && $lines[0] === 'Content-Type: application/json'){
+                    array_shift($lines);
+                    $headers = json_decode(implode($lines), true);
+                    $headers = array_combine(
+                        array_map(fn($header)=>  $header['name'], $headers),
+                        array_map(fn($header)=>  $header['value'], $headers)
+                    );
```
abnegate, line 266: We should support any JSON, not just objects with name/value

templates/php/src/Client.php.twig
```diff
@@ -240,4 +246,35 @@ class Client
...
+                    $data[$matches['name']] = $headers;
+                    continue;
+                }
+                $data[$matches['name']] = implode("\r\n",$lines) ?? '';;
+            }
+        }
+
+        $data['responseStatusCode'] = (int) ($data['responseStatusCode'] ?? '');
+        $data['duration'] = ((float) ($data['duration'] ?? ''));
+        $data['responseBody'] = Payload::fromString($data['responseBody'] ?? '');
```
abnegate, line 276: We should only set these if they already exist in $data

templates/php/src/Payload.php.twig
byawitz: It's a string, yes

christyjacob4 (approved):

PR #941, opened 2024-08-15
feat: ruby multipart

loks0n:


templates/ruby/lib/container/payload.rb.twig
```diff
@@ -0,0 +1,81 @@
...
+        mime_type
+      end
+    end
+
+    def self.from_path(path, filename: nil, mime_type: nil)
+      raise "File not found" unless File.exist?(path)
+      new(nil, path, filename, mime_type)
+    end
+
+    def self.from_data(data, filename: nil, mime_type: nil)
```
christyjacob4, line 37:
```suggestion
    def self.from_binary(data, filename: nil, mime_type: nil)
```

templates/ruby/lib/container/payload.rb.twig
```diff
@@ -0,0 +1,81 @@
...
+
+    def to_binary
+      @data || File.binread(@path)
+    end
+
+    def to_json
+      JSON.parse(@data)
+    end
+
+    def to_s
```
christyjacob4, line 57: is it not possible to define a to_string? it it a reserved function ?

templates/ruby/lib/container/multipart.rb.twig
```diff
@@ -0,0 +1,120 @@
...
+                else
+                    begin
+                        h[name] = part[:contents].force_encoding("utf-8")
+                    rescue
+                        h[name] = part[:contents]
+                    end
+                end
+            end
+
+            h
```
Meldiron, line 117: h? 😅

templates/ruby/lib/container/payload.rb.twig
```diff
@@ -0,0 +1,77 @@
+module Appwrite
+  class Payload
+    attr_reader :data
+    attr_reader :filename
+
+    def initialize(data: nil, path: nil, filename: nil)
```
Meldiron, line 6: Do we need path here?

templates/ruby/lib/container/payload.rb.twig
```diff
@@ -0,0 +1,77 @@
...
+      @path = path
+      @data = data
+      @filename = filename
+
+      @size = if @data then
+        @data.bytesize
+      else
+        File.size(@path)
```
Meldiron, line 14: Do we need this size?

templates/ruby/lib/container/payload.rb.twig
```diff
@@ -0,0 +1,77 @@
...
+      end
+    end
+
+    def read(offset: 0, length: nil)
```
Meldiron, line 18: Having generic read seems off. Would read work with strings or jsons? Is this needed specifically for files?

templates/ruby/lib/container/payload.rb.twig
```diff
@@ -0,0 +1,77 @@
...
+    # @param [String] path
+    # @param [String, nil] filename
+    # @return [Payload]
+    def self.from_file(path, filename: nil)
+        filename = if filename.nil? then
+          File.basename(path)
+        else
+          filename
+        end
+      new(nil, path, filename)
```
Meldiron, line 35: Hmm, data nil and we pass file?
Can we instead simply get data here, and pass it? That way constructor would probably not need path anymore.

templates/ruby/lib/container/payload.rb.twig
```diff
@@ -0,0 +1,77 @@
...
+    end
+
+    # @param [String] b
+    # @param [String, nil] filename
+    def self.from_binary(b, filename: nil)
```
Meldiron, line 40: lets do `bytes` instead of `b`

templates/ruby/lib/container/payload.rb.twig
loks0n: It is stored to enable loading the file in chunks

templates/ruby/lib/container/payload.rb.twig
loks0n: It is stored to enable loading the file in chunks

templates/ruby/lib/container/payload.rb.twig
loks0n: It's need for the chunking logic in the client

templates/ruby/lib/container/payload.rb.twig
loks0n: In the case of large files, you load everything into memory this way.

We can do it, but it's not how it was before

templates/ruby/lib/container/payload.rb.twig
loks0n: to_string is not reserved, but to_s is idiomatic 

I've gone with both

christyjacob4 (approved):

PR #942, opened 2024-08-15
feat: go multipart

byawitz:
## What does this PR do?

Adding multipart support for go

## Test Plan


templates/go/services/service.go.twig
```diff
@@ -95,7 +99,7 @@ func (srv *{{ service.name | caseUcfirst }}) {{ method.name | caseUcfirst }}({{
 	path := "{{ method.path }}"
 {% endif %}
 {{include('go/base/params.twig')}}
-{% if 'multipart/form-data' in method.consumes %}
+{% if 'multipart/form-data' in method.consumes and method.name | lower != "createexecution" %}
```
christyjacob4, line 102: I think we discussed to use method.type or so here

templates/go/base/params.twig
```diff
@@ -10,7 +10,11 @@
 	params["{{ parameter.name }}"] = {{ parameter.name | caseUcfirst }}
 {% else %}
 	if options.enabledSetters["{{ parameter.name | caseUcfirst}}"] {
+	{%~ if method.name | caseLower == "createexecution" and parameter.name == "body" %}
+		params["{{ parameter.name }}"] = string(options.{{ parameter.name | caseUcfirst }}.Data)
```
christyjacob4, line 14: this again feels too specific to createexecution... 

why dont we copy over the stuff from php ?

templates/go/services/service.go.twig
byawitz: This is to make sure it's not the `createexecution` method. the `method.type` won't cover all use-cases

templates/go/services/service.go.twig
```diff
@@ -1,12 +1,16 @@
 {%- set requireModelsPkg = false -%}
-{%- set requireFilesPkg = false -%}
+{%- set requirePayloadPkg = false -%}
+{%- set requireParsingLibes = false -%}
```
christyjacob4, line 3: where is this var used ? And what is it used for?

christyjacob4 (approved):

PR #948, opened 2024-08-19
Refactor go improvments

byawitz:
## What does this PR do?

- Handling upload of files smaller than the minimum chunk.
- Changing collection attributes to `[]map[string]any` aka dynamic object

### DIFF
--- a/src/SDK/Language/Go.php
+++ b/src/SDK/Language/Go.php
@@ -138,6 +138,9 @@ class Go extends Language
      */
     public function getTypeName(array $parameter, array $spec = []): string
     {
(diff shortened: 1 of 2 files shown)


src/SDK/Language/Go.php
```diff
@@ -138,6 +138,9 @@ public function getFiles(): array
      */
     public function getTypeName(array $parameter, array $spec = []): string
     {
+        if (str_contains($parameter['description'], 'Collection attributes') || str_contains($parameter['description'], 'List of attributes')) {
+            return '[]map[string]any';
+        }
```
christyjacob4, line 143: can we find a better way to do this?

PR #955, opened 2024-08-23
fix: cli login use endpoint

loks0n:


templates/cli/lib/commands/generic.js.twig
```diff
@@ -13,7 +13,8 @@ const DEFAULT_ENDPOINT = 'https://cloud.appwrite.io/v1';
 
 const loginCommand = async ({ email, password, endpoint, mfa, code }) => {
     const oldCurrent = globalConfig.getCurrentSession();
-    let configEndpoint = endpoint ?? DEFAULT_ENDPOINT;
+
+    const configEndpoint = endpoint ?? globalConfig.getEndpoint() ?? DEFAULT_ENDPOINT;
```
byawitz, line 17: Does it mean it will use the last endpoint?

byawitz (approved):

byawitz (approved):

templates/cli/lib/commands/generic.js.twig
loks0n: So it will select, in order of preference:
 - endpoint set with --endpoint on command
 - endpoint set with appwrite client --endpoint
 - "https://cloud.appwrite.io"

byawitz (approved):

Meldiron (approved):

PR #964, opened 2024-08-28
Multipart support changes for node

vermakhushboo:
## What does this PR do?

- Add payload class
- Rename older `Payload` to `JSONPayload`
- Remove `InputFile` and replace it with `Payload`

## Test Plan

## Related PRs and Issues

(Write your answer here.)


templates/node/src/client.ts.twig
```diff
@@ -1,8 +1,11 @@
 import { fetch, FormData, File } from 'node-fetch-native-with-agent';
 import { createAgent } from 'node-fetch-native-with-agent/agent';
 import { Models } from './models';
+import { Payload } from './Payload';
+import * as multipart from 'parse-multipart-data';
+const { buffer } = require('node:stream/consumers');
 
-type Payload = {
+type JSONPayload = {
```
loks0n, line 8: Would RequestPayload be a better name? It's not always in the context of json

templates/node/src/client.ts.twig
```diff
@@ -1,8 +1,11 @@
 import { fetch, FormData, File } from 'node-fetch-native-with-agent';
 import { createAgent } from 'node-fetch-native-with-agent/agent';
 import { Models } from './models';
+import { Payload } from './Payload';
```
byawitz, line 4: I think the file path should `./payload` lowercase as of the copy action [here](https://github.com/appwrite/sdk-generator/pull/964/files#diff-947d6880be0e5cb7ad5aaef122edb9500aad7a480bcc6eef441a1478f5ea0a94R167)

templates/node/src/Payload.ts.twig
```diff
@@ -0,0 +1,58 @@
...
+  public toString(): string {
+    return this.data.toString("utf-8");
+  }
+
+  public toFile(path: string): void {
+    writeFileSync(path, this.data);
+  }
+
+  public static fromBinary(bytes: Buffer, name?: string): Payload {
+    return new Payload(Buffer.from(bytes), name);
```
byawitz, line 38: Is this 👇 could work? as the data is expecting `buffer` already
```js
return new Payload(bytes, name);
```

tests/languages/node/test.js
```diff
@@ -65,18 +65,18 @@ async function start() {
     response = await general.redirect();
     console.log(response.result);
 
-    response = await general.upload('string', 123, ['string in array'], InputFile.fromPath(__dirname + '/../../resources/file.png', 'file.png'));
+    response = await general.upload('string', 123, ['string in array'], Payload.fromFile(__dirname + '/../../resources/file.png', 'file.png').toFile());
     console.log(response.result);
 
-    response = await general.upload('string', 123, ['string in array'], InputFile.fromPath(__dirname + '/../../resources/large_file.mp4', 'large_file.mp4'));
+    response = await general.upload('string', 123, ['string in array'], Payload.fromFile(__dirname + '/../../resources/large_file.mp4', 'large_file.mp4').toFile());
```
loks0n, line 71: We shouldn't need toFile here
fromFile and toFile seems bad dx

templates/node/src/Payload.ts.twig
```diff
@@ -0,0 +1,59 @@
+import { readFileSync } from "fs";
```
loks0n, line 1: We can't import from `fs` or `path` here as these are node only, and this will break Next.js middleware

loks0n (dismissed):

tests/languages/node/test.js
christyjacob4: Agreed. The toFile() conversion should happen inside the createFile() endpoint

src/SDK/Language/Node.php
```diff
@@ -164,8 +167,8 @@ public function getFiles(): array
             ],
             [
                 'scope'         => 'default',
-                'destination'   => 'src/inputFile.ts',
-                'template'      => 'node/src/inputFile.ts.twig',
+                'destination'   => 'src/payload.ts',
+                'template'      => 'node/src/Payload.ts.twig',
```
christyjacob4, line 171:
```suggestion
                'template'      => 'node/src/payload.ts.twig',
```
use lowercase files names

src/SDK/Language/Node.php
```diff
@@ -20,6 +20,9 @@ public function getTypeName(array $parameter, array $method = []): string
         if (!empty($parameter['enumValues'])) {
             return \ucfirst($parameter['name']);
         }
+        if (($parameter['name'] ?? '') === 'body') {
```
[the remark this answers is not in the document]
christyjacob4: this is risky.. if there is any other parameter in the Appwrite API with the name `body` its going to use Payload as the type which is not intended

templates/node/src/Payload.ts.twig
Meldiron: What was the final deciison here?

templates/node/src/Payload.ts.twig
```diff
@@ -0,0 +1,58 @@
...
+import { File } from "node-fetch-native-with-agent";
+import { basename } from "path";
+
+export class Payload extends File {
```
Meldiron, line 5: Do we need to `extend File` here?

templates/node/src/Payload.ts.twig
```diff
@@ -0,0 +1,58 @@
...
+  private data: Buffer;
+  private fileName: string;
+
+  constructor(data: Buffer, name?: string) {
+    super([data], name || 'unnamed');
+    this.data = data;
+    this.fileName = name || 'unnamed';
```
Meldiron, line 12: I would avoid doing "unnamed". It should allow null value, and then in endpoints, throw exception if name is needed and is not provided - like createStorage

templates/node/src/Payload.ts.twig
```diff
@@ -0,0 +1,58 @@
...
+    const data = Buffer.from(JSON.stringify(object), "utf-8");
+    return new Payload(data, name);
+  }
+
+  public static fromString(text: string, name?: string): Payload {
+    const data = Buffer.from(text, "utf-8");
+    return new Payload(data, name);
+  }
+
+  public static fromFile(filePath: string, name?: string): Payload {
```
Meldiron, line 53: I would call it just `path`, it's already in namespace of `file` since its in `fromFile`

templates/node/src/client.ts.twig
```diff
@@ -1,8 +1,11 @@
 import { fetch, FormData, File } from 'node-fetch-native-with-agent';
 import { createAgent } from 'node-fetch-native-with-agent/agent';
 import { Models } from './models';
+import { Payload } from './payload';
+import * as multipart from 'parse-multipart-data';
+const { buffer } = require('node:stream/consumers');
```
Meldiron, line 6: Is this nessessary? Cant we do it without library? DOes it convert from X to buffer? what is X?

templates/node/src/client.ts.twig
```diff
@@ -271,6 +275,39 @@ class Client {
...
+        } else if (response.headers.get('content-type')?.includes('multipart/form-data')) {
+            const body = await buffer(response.body);
+            const boundary = multipart.getBoundary(
+                response.headers.get("content-type") || ""
+            );
+            const parts = multipart.parse(body, boundary);
+            const partsObject: { [key: string]: any } = {};
+            
+            for (const part of parts) {
+                if (part.name) {
```
Meldiron, line 287: is this nessessary? If so, lets avoid nesting. we can do opposite of that, and `continue`

templates/node/src/client.ts.twig
```diff
@@ -271,6 +275,39 @@ class Client {
...
+                    if (part.name === "responseBody") {
+                        partsObject[part.name] = new Payload(part.data, part.filename || undefined);
```
Meldiron, line 289:
```suggestion
                        partsObject[part.name] = new Payload(part.data, part.filename);
```
I think its gonna be undefined by default

templates/node/src/client.ts.twig
```diff
@@ -271,6 +275,39 @@ class Client {
...
+                        partsObject[part.name] = parseFloat(part.data.toString());
+                    } else if (part.type === 'application/json') {
+                        try {
+                            let jsonData = JSON.parse(part.data.toString());
+                            if (Array.isArray(jsonData) && jsonData.length > 0 && 'name' in jsonData[0] && 'value' in jsonData[0]) {
+                                jsonData = Object.fromEntries(jsonData.map(item => [item.name, item.value]));
+                            }
+                            partsObject[part.name] = jsonData;
+                        } catch (e) {
+                            console.error('Error parsing JSON:', e);
```
Meldiron, line 302: Woah, no. Just let the exception throw. Hiding it from developer with out custom logging logic is bad idea.

templates/node/src/client.ts.twig
```diff
@@ -271,6 +275,39 @@ class Client {
...
+                    if (part.name === "responseBody") {
+                        partsObject[part.name] = new Payload(part.data, part.filename || undefined);
+                    } else if (part.name === "responseStatusCode") {
+                        partsObject[part.name] = parseInt(part.data.toString(), 10);
+                    } else if (part.name === "duration") {
+                        partsObject[part.name] = parseFloat(part.data.toString());
+                    } else if (part.type === 'application/json') {
+                        try {
+                            let jsonData = JSON.parse(part.data.toString());
+                            if (Array.isArray(jsonData) && jsonData.length > 0 && 'name' in jsonData[0] && 'value' in jsonData[0]) {
```
Meldiron, line 297: Can you explain to me what exactly does this do with nested JSON data? I thought we would just JSON.parse, and that's it.

Meldiron: If we actually need this, lets separate it into a helper function. There is too much nesting here and it's hard to follow.

templates/node/src/Payload.ts.twig
```diff
@@ -0,0 +1,58 @@
...
+    super([data], name || 'unnamed');
+    this.data = data;
+    this.fileName = name || 'unnamed';
+  }
+
+  public getData(): Buffer {
+    return this.data;
+  }
+
+  public getName(): string {
```
Meldiron, line 19: I dont see this method being used. Can you link me where you used it?

templates/node/src/services/template.ts.twig
```diff
@@ -47,15 +48,19 @@ export class {{ service.name | caseUcfirst }} {
         {%~ endif %}
         {%~ endfor %}
         const apiPath = '{{ method.path }}'{% for parameter in method.parameters.path %}.replace('{{ '{' }}{{ parameter.name | caseCamel | escapeKeyword }}{{ '}' }}', {{ parameter.name | caseCamel | escapeKeyword }}){% endfor %};
-        const payload: Payload = {};
+        const originalPayload: Params = {};
```
Meldiron, line 51: Can we rename this back to just `payload`? Cant clearly see reason why it was renamed

templates/node/src/Payload.ts.twig
vermakhushboo: Even I don't have enough clarity. @loks0n please advise.

templates/node/src/Payload.ts.twig
vermakhushboo: It was added because `upload` method in this example expects a `File` object
`response = await general.upload('string', 123, ['string in array'], Payload.fromFile(__dirname + '/../../resources/large_file.mp4', 'large_file.mp4'));`

templates/node/src/Payload.ts.twig
vermakhushboo: @christyjacob4 @loks0n, there is a lot of back-and-forth in this. Can we finalise it?

templates/node/src/services/template.ts.twig
vermakhushboo: Because if we don't, then the compiler complains of naming conflicts

templates/node/src/Payload.ts.twig
vermakhushboo: It is used in chunkedUpload

templates/node/src/Payload.ts.twig
vermakhushboo: I need to do it like this because of strict type checking in ts

templates/node/src/Payload.ts.twig
vermakhushboo: We decided to use Blob in `fromFile` instead of path for node

src/SDK/Language/Node.php
```diff
@@ -30,7 +30,9 @@ public function getTypeName(array $parameter, array $method = []): string
                 }
                 return 'string[]';
             case self::TYPE_FILE:
-                return "File";
+                return 'Payload';
+            case self::TYPE_PAYLOAD:
+                return 'Payload';
```
Meldiron, line 35: This can become:

```php
            case self::TYPE_FILE:
            case self::TYPE_PAYLOAD:
                return 'Payload';
```

src/SDK/Language/Node.php
```diff
@@ -120,7 +122,8 @@ public function getParamExample(array $param): string
                     $output .= '{}';
                     break;
                 case self::TYPE_FILE:
-                    $output .= "InputFile.fromPath('/path/to/file', 'filename')";
+                    $blob = file_get_contents('path/to/file');
+                    $output .= "Payload.fromFile(" . $blob . ", 'filename')";
```
Meldiron, line 126: This is for docs I believe. Let's update it with proper example once we have final decision from Christy

src/SDK/Language/Node.php
```diff
@@ -138,7 +141,8 @@ public function getParamExample(array $param): string
                     $output .= "'{$example}'";
                     break;
                 case self::TYPE_FILE:
-                    $output .= "InputFile.fromPath('/path/to/file', 'filename')";
+                    $blob = file_get_contents('path/to/file');
+                    $output .= "Payload.fromFile(" . $blob . ", 'filename')";
```
Meldiron, line 145: This is for docs I believe. Let's update it with proper example once we have final decision from Christy

templates/node/src/client.ts.twig
```diff
@@ -271,6 +284,35 @@ class Client {
...
+                    continue;
+                }
+                if (part.name === "responseBody") {
+                    partsObject[part.name] = Payload.fromBinary(part.data, part.filename);
+                } else if (part.name === "responseStatusCode") {
+                    partsObject[part.name] = parseInt(part.data.toString());
+                } else if (part.name === "duration") {
+                    partsObject[part.name] = parseFloat(part.data.toString());
+                } else if (part.type === 'application/json') {
+                    try {
```
Meldiron, line 306: Note for myself to see how other PRs do it

templates/node/src/client.ts.twig
```diff
@@ -1,8 +1,12 @@
-import { fetch, FormData, File } from 'node-fetch-native-with-agent';
+import { fetch, FormData } from 'node-fetch-native-with-agent';
 import { createAgent } from 'node-fetch-native-with-agent/agent';
 import { Models } from './models';
+import { Payload } from './payload';
+import { Blob } from 'node:buffer';
+import * as multipart from 'parse-multipart-data';
+import { buffer } from 'node:stream/consumers';
```
Meldiron, line 7: Note for self why blob and buffer needs an import

templates/node/src/payload.ts.twig
```diff
@@ -0,0 +1,49 @@
+export class Payload {
+  private data: Buffer;
+  private fileName: string;
+  public size: number;
+
+  constructor(data: Buffer, name?: string) {
+    this.data = data;
+    this.fileName = name || 'unnamed';
```
Meldiron, line 8: Note for self to compare with other SDKs

Meldiron:
Merging as Node tests are passing

PR #965, opened 2024-08-28
feat: web multipart support

loks0n:


byawitz (approved):

.github/workflows/tests.yml
```diff
@@ -26,8 +26,7 @@ jobs:
           DotNet80,
           FlutterStable,
           FlutterBeta,
-          Go112,
-          Go118,
```
Meldiron, line 30: why not keep 1.18 tests? did they fail?

templates/web/src/index.ts.twig
```diff
@@ -9,8 +9,9 @@ export { Client, Query, {{spec.title | caseUcfirst}}Exception } from './client';
 {% for service in spec.services %}
 export { {{service.name | caseUcfirst}} } from './services/{{service.name | caseDash}}';
 {% endfor %}
-export type { Models, Payload, RealtimeResponseEvent, UploadProgress } from './client';
```
Meldiron, line 12: We could export `Params` here instead. Not sure why it was exported in the first place, but seems reasonable to keep it, just under new name

templates/web/src/payload.ts.twig
```diff
@@ -0,0 +1,40 @@
+export class Payload extends Blob {
```
Meldiron, line 1: Does it need to extend blob? Would prefer to keep it simple if possible

templates/web/src/payload.ts.twig
```diff
@@ -0,0 +1,40 @@
...
+    public name: string | null = null;
```
Meldiron, line 2: In Node we do it private, and have getName method. No idea which is better, lets just be consistent since they are both same language

templates/web/src/payload.ts.twig
```diff
@@ -0,0 +1,40 @@
...
+
+    constructor(data: BlobPart[], name?: string) {
```
Meldiron, line 4: Whats difference between Buffer and BlobPart[]? We use Buffer in Node

templates/web/src/payload.ts.twig
```diff
@@ -0,0 +1,40 @@
...
+
+    public async toBinary(): Promise<Uint8Array> {
+        return new Uint8Array(await this.arrayBuffer());
+    }
+
+    public async toFile(): Promise<File> {
+        return new File([this], this.name || "file");
+    }
+
+    public static fromFile(file: File): Payload {
```
Meldiron, line 25: Here we should have optional name too. If not provided, take it from file name

templates/web/src/payload.ts.twig
```diff
@@ -0,0 +1,40 @@
...
+    public async toJson<T = unknown>(): Promise<T> {
+        return JSON.parse(await this.text());
+    }
+
+    public async toBinary(): Promise<Uint8Array> {
+        return new Uint8Array(await this.arrayBuffer());
+    }
+
+    public async toFile(): Promise<File> {
+        return new File([this], this.name || "file");
```
Meldiron, line 22: We should not have default file names. Let it be null/udefined

templates/web/src/payload.ts.twig
```diff
@@ -0,0 +1,40 @@
...
+
+    public static fromString(data: string, name?: string): Payload {
+        return new Payload([data], name);
+    }
+
+    public static fromJson(data: unknown, name?: string): Payload {
+        return new Payload([JSON.stringify(data)], name);
+    }
+
+    public static fromBinary(data: Uint8Array, name?: string): Payload {
```
Meldiron, line 37: Why not Buffer?

templates/web/src/payload.ts.twig
loks0n: File objects must have a name

templates/web/src/payload.ts.twig
loks0n: It's idiomatic to just have a public field as an accessor, imo

having a `getName` linked to a private field is very enterprise java

templates/web/src/payload.ts.twig
loks0n: BlobPart can be string, uintarray, arraybuffer ...

templates/web/src/payload.ts.twig
```diff
@@ -0,0 +1,40 @@
+export class Payload extends Blob {
+    public name: string | null = null;
+
+    constructor(data: BlobPart[], name?: string) {
+        super(data);
+        this.name = name || null;
+    }
+
+    public async toString() {
```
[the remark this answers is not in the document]
loks0n: Not sure why this is better, the advantage of async is that the parsing is deferred

templates/web/src/client.ts.twig
```diff
@@ -569,56 +571,59 @@ class Client {
                 case 'multipart/form-data':
                     const formData = new FormData();
 
-                    for (const [key, value] of Object.entries(params)) {
-                        if (value instanceof File) {
-                            formData.append(key, value, value.name);
+                    for (const [name, value] of Object.entries(params)) {
+                        if (value instanceof Payload) {
+                            formData.append(name, await value.toFile(), value.filename);
```
loks0n, line 576: Is it possible to remove the await?

templates/web/src/client.ts.twig
```diff
@@ -539,7 +541,7 @@ class Client {
         }
     }
 
-    prepareRequest(method: string, url: URL, headers: Headers = {}, params: Payload = {}): { uri: string, options: RequestInit } {
+    async prepareRequest(method: string, url: URL, headers: Headers = {}, params: Params = {}): Promise<{ uri: string, options: RequestInit }> {
```
loks0n, line 544: Is it possible to remove the await?

templates/web/src/client.ts.twig
```diff
@@ -569,56 +571,59 @@ class Client {
...
-            }
+        while (start < payload.size) {
+            const end = Math.min(start + Client.CHUNK_SIZE, payload.size);
 
-            headers['content-range'] = `bytes ${start}-${end-1}/${file.size}`;
-            const chunk = file.slice(start, end);
+            headers['content-range'] = `bytes ${start}-${end-1}/${payload.size}`;
 
-            let payload = { ...originalPayload, file: new File([chunk], file.name)};
+            const buffer = await payload.toBinary(start, end - start);
```
loks0n, line 616: Is it possible to remove the await?

templates/web/src/client.ts.twig
```diff
@@ -633,8 +638,8 @@ class Client {
         return response;
     }
 
-    async call(method: string, url: URL, headers: Headers = {}, params: Payload = {}, responseType = 'json'): Promise<any> {
-        const { uri, options } = this.prepareRequest(method, url, headers, params);
+    async call(method: string, url: URL, headers: Headers = {}, params: Params = {}, responseType = 'json'): Promise<any> {
+        const { uri, options } = await this.prepareRequest(method, url, headers, params);
```
loks0n, line 642: Is it possible to remove the await?

Meldiron (approved):

loks0n:
LGTM

Issue #963, opened 2024-08-27
🐛 Bug Report: Error in Storage API createFile function


abhishek-visitnow:
### 👟 Reproduction steps

I'm trying to upload a file greater than 4 Mb. Following code is working when the size of file is less than 4 mb.

Using dotnet sdk v0.8.3

```
File myfile = await storage.CreateFile(
                 bucketId: "bucket-id",
                 fileId: ID.Unique(),
                 file: InputFile.FromBytes(
                     Encoding.UTF8.GetBytes(json),
                     "filename.json,
                     "application/json"
                     ),
                 permissions: permissions! // optional
             );
```

### 👍 Expected behavior

It should upload the file.

### 👎 Actual Behavior

Getting following error:
```
Appwrite.AppwriteException: The requested route was not found. Please refer to the API docs and try again.
   at Appwrite.Client.Call[T](String method, String path, Dictionary`2 headers, Dictionary`2 parameters, Func`2 convert)
   at Appwrite.Client.ChunkedUpload[T](String path, Dictionary`2 headers, Dictionary`2 parameters, Func`2 converter, String paramName, String idParamName, Action`1 onProgress)
   at DotNetRuntime.Handler.Main(RuntimeContext Context) in /usr/local/server/src/Index.cs:line 67
   at DotNetRuntime.Program.Action(HttpRequest request) in /usr/local/server/src/Program.cs:line 194
```

### 🎲 Appwrite version

Version 1.5.x

### 💻 Operating system

Linux

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

abhishek-visitnow:
@stnguyen90 Does this mean it will work if I add a string Id to `fileId`?

abhishek-visitnow:
I see the variable name is mismatched `path: "$path/${params[idParamName]}",`

abhishek-visitnow:
@stnguyen90 Any update on this?

dev-p26:
Any plan for when this fix is going to be released?

abhishek-visitnow:
New version of DotNet SDK is released
https://github.com/appwrite/sdk-for-dotnet/releases/tag/0.10.1

With a release note: `fix: chunk upload by @byawitz in #52`

[...]

PR #969, opened 2024-09-02
Fixed file upload for .NET SDK

jtrcarlos:
## What does this PR do?

This PR includes fixes to the 'CreateFile' flow while using the .NET SDK which was not working.
This changes were applied and tested directly to the SDK and then replicated to this template.

What was done:
  - ignored parameters with no value, which was causing a null object reference
  - fixed the logic regarding the GET method that checks for an already existing file with the same 'fileId'
  - fixed a malformed string (path for the GET method mentioned above)
  - fixed a wrong re-assignment of the 'parameters' variable (used in the GET method mentioned above)

## Test Plan

This changes were tested manually by using a locally fixed version of the SDK, using both self-hosted and cloud endpoints.
From what I found, if this is to be tested, it is using PHP, which I don't know.

Files used to test:
  - some .pdf (various sizes)
  - some .png (various sizes)

## Related PRs and Issues

Issue #963 

Yes.

### DIFF
--- a/templates/dotnet/Package/Client.cs.twig
+++ b/templates/dotnet/Package/Client.cs.twig
@@ -154,6 +154,12 @@ namespace {{ spec.title | caseUcfirst }}
 
                 foreach (var parameter in parameters)
                 {
+                    // Skip parameters with no value
+                    if (parameter.Value == null)
+                    {
+                        continue;
+                    }
+
                     if (parameter.Key == "file")
                     {
                         form.Add(((MultipartFormDataContent)parameters["file"]).First()!);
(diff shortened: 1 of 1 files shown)


jtrcarlos:
Issue #963 mentions removing "unique()" check that is missing from this PR.

dev-p26:
Any plan for when this fix is going to be released?

stnguyen90:
@jtrcarlos, thanks for creating this PR! 🙏🏼 Unfortunately, this isn't the approach we want to take as it doesn't match the other SDKs. We'll take it from here.

Also, in the future, please make sure you comment on the issue and ask to be assigned before submitting a PR to avoid wasted work.

PR #993, opened 2024-10-02
Fix chunk uploads

byawitz:
## What does this PR do?

Fix chunk uploads with predefined ID for
- Kotlin (android)
- Ruby

## Test Plan

Running these code snippets.
### Kotlin
```kotlin
val client = Client()
    .setEndpoint("https://cloud.appwrite.io/v1") // Your API Endpoint
    .setProject("...")
    .setKey("...");

val s = Storage(client)
val response = s.createFile(
    bucketId = "...",
    fileId = ID.unique(),
    file = InputFile.fromPath("aaa.mp4"),
)
```

### Ruby
```ruby
client = Client.new
               .set_endpoint('https://cloud.appwrite.io/v1') # Your API Endpoint
               .set_project('...') # Your project ID
               .set_key('...') # The user session to authenticate with

storage = Storage.new(client)

result = storage.create_file(
  bucket_id: '...',
  file_id: ID.unique,
  file: InputFile.from_path('aaa.mp4'),
)
```
`aaa.mp4` is bigger than 10MB.

## Related PRs and Issues

#963 

### Have you read the [Contributing Guidelines on issues]
Yes

### DIFF
--- a/templates/android/library/src/main/java/io/package/Client.kt.twig
+++ b/templates/android/library/src/main/java/io/package/Client.kt.twig
@@ -60,7 +60,7 @@ class Client @JvmOverloads constructor(
     internal lateinit var http: OkHttpClient
 
     internal val headers: MutableMap<String, String>
-    
+
     val config: MutableMap<String, String>
 
     internal val cookieJar = ListenableCookieJar(CookieManager(
@@ -87,14 +87,14 @@ class Client @JvmOverloads constructor(
             "x-sdk-platform" to "{{ sdk.platform }}",
             "x-sdk-language" to "{{ language.name | caseLower }}",
             "x-sdk-version" to "{{ sdk.version }}"{% if spec.global.defaultHeaders | length > 0 %},{% endif %}
-            
+
 {% for key,header in spec.global.defaultHeaders %}
             "{{ key | caseLower }}" to "{{ header }}"{% if not loop.last %},{% endif %}
 {% endfor %}
 
         )
         config = mutableMapOf()
-        
+
         setSelfSigned(selfSigned)
     }
 
@@ -119,10 +119,10 @@ class Client @JvmOverloads constructor(
 {% endfor %}
     /**
      * Set self Signed
-     * 
+     *
      * @param status
      *
-     * @return this     
+     * @return this
      */
     fun setSelfSigned(status: Boolean): Client {
         selfSigned = status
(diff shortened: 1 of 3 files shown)


stnguyen90 (commented):
can we rely on the error type instead of the error message?

templates/android/library/src/main/java/io/package/Client.kt.twig
```diff
@@ -364,16 +364,22 @@ class Client @JvmOverloads constructor(
...
+                    method = "GET",
+                    path = "$path/${params[idParamName]}",
+                    headers = headers,
+                    params = emptyMap(),
+                    responseType = Map::class.java,
+                )
+                val chunksUploaded = current["chunksUploaded"] as Long
+                offset = chunksUploaded * CHUNK_SIZE
+            } catch (e: Exception) {
+                if (e.message != null && !e.message!!.contains("file could not be found")) {
```
loks0n, line 379: Agree with Steven, this seems fragile...

templates/android/library/src/main/java/io/package/Client.kt.twig
```diff
@@ -364,16 +364,22 @@ class Client @JvmOverloads constructor(
...
+                    throw e
```
loks0n, line 380: Doesn't `call()` methods already throw these exceptions, is there any need to catch and throw here?

templates/ruby/lib/container/client.rb.twig
```diff
@@ -304,13 +308,13 @@ module {{ spec.title | caseUcfirst }}
             when Hash  then value.map { |k,v| encode(v, append_key(key,k)) }.join('&')
             when Array then value.map { |v| encode(v, "#{key}[]") }.join('&')
             when nil   then ''
```
loks0n, line 310: Unrelated, but can we fix the spacing here? Looks weird

templates/android/library/src/main/java/io/package/Client.kt.twig
```diff
@@ -364,16 +364,22 @@ class Client @JvmOverloads constructor(
...
+            try {
+                // Make a request to check if a file already exists
+                val current = call(
+                    method = "GET",
+                    path = "$path/${params[idParamName]}",
+                    headers = headers,
+                    params = emptyMap(),
+                    responseType = Map::class.java,
+                )
+                val chunksUploaded = current["chunksUploaded"] as Long
```
loks0n, line 376: Is `Long` type the best choice?

loks0n (changes requested):

PR #999, opened 2024-10-09
Fix: Deno commands

Meldiron:
## What does this PR do?

- Applies proper build command for starters used in Appwrite CLI
- Uses proper command to start Deno during local devleopment

## Test Plan

- [x] Manual QA in Gitpod:

Before:

CleanShot 2024-10-09 at 20 54 31@2x

AFter:

CleanShot 2024-10-09 at 20 54 10@2x

## Related PRs and Issues

x

Yes

### DIFF
--- a/templates/cli/lib/emulation/utils.js.twig
+++ b/templates/cli/lib/emulation/utils.js.twig
@@ -50,7 +50,7 @@ const systemTools = {
     },
     'deno': {
         isCompiled: false,
-        startCommand: "deno start",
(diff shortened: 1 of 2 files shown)


stnguyen90 (commented):
These still work for older versions of deno?

Meldiron:
> These still work for older versions of deno?

Yes, `cache` has been supported for long time. the `deno install` is actually the new one that wouldn't work. Plus it would still want entrypoint, or `deno.json`

Start command should work in all versions too

stnguyen90 (approved):

PR #1018, opened 2024-12-18
feat: realtime heartbeat for react native

loks0n:


lohanidamodar (approved):
Looks good to me. This is same as what we have in Web right?

lohanidamodar (approved):
Looks good to me. This is same as what we have in Web right?

loks0n:
> Looks good to me. This is same as what we have in Web right?

Yeah it's pretty much identical, except for a small difference with types

christyjacob4 (approved):

PR #1033, opened 2025-01-29
fix: dotnet & dart models

loks0n:
- **fix(dotnet): models formatting**
- **fix(dart): models empty sub_schema array scenario**

### DIFF
--- a/templates/dart/lib/src/models/model.dart.twig
+++ b/templates/dart/lib/src/models/model.dart.twig
@@ -6,11 +6,12 @@ class {{ definition.name | caseUcfirst | overrideIdentifier }} implements Model 
 {% for property in definition.properties %}
     /// {{ property.description }}
     final {% if not property.required %}{{_self.sub_schema(property)}}? {{ property.name | escapeKeyword }}{% else %}{{_self.sub_schema(property)}} {{ property.name | escapeKeyword }}{% endif %};
-{% endfor %}
-{% if definition.additionalProperties %}
-    final Map<String, dynamic> data;
-{% endif %}
 
+{% endfor %}
+{%~ if definition.additionalProperties %}
+    final Map<String, dynamic> data;
+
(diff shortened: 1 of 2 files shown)


templates/dart/lib/src/models/model.dart.twig
```diff
@@ -33,7 +34,7 @@ class {{ definition.name | caseUcfirst | overrideIdentifier }} implements Model
                 {%- endif -%}
             {%- else -%}
                 {%- if property.type == 'array' -%}
-                    List<{{ property.items.type | caseUcfirst }}>.from(map['{{property.name | escapeDollarSign }}']?.map((x) => x{% if property.items.type == "string" %}.toString(){% endif %}) ?? [])
```
loks0n, line 36: Not sure where this `property.items.type` property came from, it's empty.

templates/dart/lib/src/models/model.dart.twig
```diff
@@ -33,7 +34,7 @@ class {{ definition.name | caseUcfirst | overrideIdentifier }} implements Model
...
+                    List.from(map['{{property.name | escapeDollarSign }}'] ?? [])
```
abnegate, line 37: Shouldn't we use property.sub_schema for the list type here?

ChiragAgg5k (approved):

templates/dart/lib/src/models/model.dart.twig
loks0n: This else branch is handling when `property.sub_schema` is empty. We handle the normal case above.

The old code case it to generate List<> for AttributeList, which doesn't have property.sub_schema set, I guess because it would require a complicated union of all the different attribute types.

abnegate: I see it now :+1:

abnegate (approved):

PR #1044, opened 2025-03-07
chore: remove unnecessary async and spacing

ChiragAgg5k:
## What does this PR do?

1. fix spacing in sdk for web (there was one extra in front of method names).
2. remove unnecessary async in sdk for node.
3. make the return for `webAuth` be string or void instead of promise. this previously did not give an error since methods were async.
4. remove unnecessary adding of payload in web auth. it would just add project in the payload, which would already be added in headers:

Screenshot 2025-03-07 at 2 10 23 PM

yes.

### DIFF
--- a/templates/web/src/services/template.ts.twig
+++ b/templates/web/src/services/template.ts.twig
@@ -85,10 +85,10 @@ export class {{ service.name | caseUcfirst }} {
             window.location.href = uri.toString();
             return Promise.resolve();
         } else {
-            return Promise.resolve(uri.toString());
+            return uri.toString();
         }
         {%~ elseif method.type == 'location' %}
-        return Promise.resolve(uri.toString());
+        return uri.toString();
         {%~ elseif 'multipart/form-data' in method.consumes %}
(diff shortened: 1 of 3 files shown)


templates/web/src/services/template.ts.twig
```diff
@@ -68,12 +68,6 @@ export class {{ service.name | caseUcfirst }} {
             {%~ endfor %}
         }
 
-        {%~ for node in method.auth %}
-        {%~ for key,header in node|keys %}
-        payload['{{header|caseLower}}'] = this.client.config.{{header|caseLower}};
-        {%~ endfor %}
-        {%~ endfor %}
-
```
loks0n, line 76: Pretty sure this has be added so that we set the project param in webAuth requests, can you confirm it still works?

ChiragAgg5k: ok so we can just enclose it inside `if method.type == 'webAuth`?

loks0n (approved):

PR #1050, opened 2025-03-14
fix: appwrite exception already read

ChiragAgg5k:
## What does this PR do?

A response stream can only be read once. That's why reading it for text and then reading it again for json gave error "Already read". The error was caught and fixed in web and node, but left in react-native. This PR makes the logic consistent across all three.

## Related PRs and Issues

* https://github.com/appwrite/sdk-for-react-native/issues/49
* https://github.com/appwrite/sdk-for-react-native/issues/50

Yes.

### DIFF
--- a/templates/react-native/src/client.ts.twig
+++ b/templates/react-native/src/client.ts.twig
@@ -414,7 +414,6 @@ class Client {
 
         try {
             let data = null;
-            let text: string = '';
 
             const response = await fetch(url.toString(), options);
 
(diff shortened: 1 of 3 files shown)


templates/react-native/src/client.ts.twig
```diff
@@ -424,7 +425,9 @@ class Client {
 
             if (response.headers.get('content-type')?.includes('application/json')) {
                 data = await response.json();
+                text = JSON.stringify(data);
```
stnguyen90, line 428: This may needlessly stringify data when not needed, right?

ChiragAgg5k: we return it as "response" like this:

```
{{spec.title | caseUcfirst}}Exception(data?.message, response.status, data?.type, text);
```

stnguyen90: Only if there is an exception, right? What if there isn't? And what happens more often, exception or no exception?

ChiragAgg5k: ok yeah that makes sense 😅

stnguyen90 (approved):

PR #1057, opened 2025-03-23
chore: update setting of realtime endpoint

ChiragAgg5k:
## What does this PR do?

Due to frequent endpoint changes with `multi-region`, the sdk context changes quite often and hence we need update the realtime endpoint regardless if its been set before.

NOW throws an `AppwriteException` if URL is not valid. breaking for apple and swift.

Yes.

### DIFF
--- a/templates/android/library/src/main/java/io/package/Client.kt.twig
+++ b/templates/android/library/src/main/java/io/package/Client.kt.twig
@@ -179,7 +179,7 @@ class Client @JvmOverloads constructor(
     fun setEndpoint(endpoint: String): Client {
         this.endpoint = endpoint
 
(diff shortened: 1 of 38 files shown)


ItzNotABug (approved):

templates/android/library/src/main/java/io/package/Client.kt.twig
```diff
@@ -176,10 +176,10 @@ class Client @JvmOverloads constructor(
      *
      * @return this
      */
-     @Throws({{ spec.title | caseUcfirst }}Exception::class)
+    @Throws(IllegalArgumentException::class)
```
abnegate, line 179: We'll need to add an import as well

abnegate (approved):

PR #1060, opened 2025-03-29
fix: hot swapping error in python ml function

ChiragAgg5k:
## What does this PR do?

In python-ml function, we copy over files to `runtime-env` here - https://github.com/open-runtimes/open-runtimes/blob/088ef387b097cc7bb29aff6602372ac1196c5925/runtimes/python/versions/latest/helpers/prepare-start.sh#L4

This folder contains various files, even including symlinks.

Symlinks when copied over won't be valid files anymore and hence the `getAllFiles` function in CLI throws an error trying to fetch absolute paths for it.

So the PR makes the function just skip any invalid symlinks.

## Test Plan

Before:

Screenshot 2025-03-29 at 5 44 07 PM

After:

Screenshot 2025-03-29 at 5 44 36 PM

## Related PRs and Issues

- https://github.com/appwrite/sdk-for-cli/issues/150

Yes.

### DIFF
--- a/templates/cli/lib/utils.js.twig
+++ b/templates/cli/lib/utils.js.twig
@@ -9,9 +9,10 @@ function getAllFiles(folder) {
     const files = [];
     for (const pathDir of fs.readdirSync(folder)) {
         const pathAbsolute = path.join(folder, pathDir);
-        if (fs.statSync(pathAbsolute).isDirectory()) {
+        const stats = fs.lstatSync(pathAbsolute);
+        if (stats.isDirectory() && !stats.isSymbolicLink()) {
             files.push(...getAllFiles(pathAbsolute));
-        } else {
+        } else if (stats.isFile() && !stats.isSymbolicLink()) {
             files.push(pathAbsolute);
         }
     }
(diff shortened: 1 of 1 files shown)


stnguyen90 (commented):
Can you make sure some of the other runtimes still work as expected? Particularly:

* node
* python
* dart
* bun

Also, it's possible to deploy a function where certain folders are symlinks. This helps to re-use code. For example, I have a symlink [here](https://github.com/stnguyen90/places/blob/feature/1.3-upgrade/appwrite-functions/create-comment/src/common) to re-use this folder of [common](https://github.com/stnguyen90/places/tree/feature/1.3-upgrade/appwrite-functions/common) modules. Will this setup work locally with the CLI?

ChiragAgg5k:
@stnguyen90 Yeah you are right, the approach i used would fail if any symlinks were used.

I have changed it to just skip symlinks that are not valid and throw an error. It will work the same for all the other files.

stnguyen90:
And you've tested with multiple runtimes like:

* node
* python
* dart
* bun

ChiragAgg5k:
Screenshot 2025-03-31 at 10 32 04 PM

@stnguyen90 yep i just did doing so. lgtm

stnguyen90 (approved):

PR #1072, opened 2025-05-02
chore: improve route not found error message

ChiragAgg5k:
## What does this PR do?

Improved error message.

yes.

### DIFF
--- a/mock-server/src/Utopia/Exception.php
+++ b/mock-server/src/Utopia/Exception.php
@@ -69,7 +69,7 @@ const errorConfig = [
     ],
     Exception::GENERAL_ROUTE_NOT_FOUND => [
         'name' => Exception::GENERAL_ROUTE_NOT_FOUND,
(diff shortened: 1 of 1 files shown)


mock-server/src/Utopia/Exception.php
```diff
@@ -69,7 +69,7 @@
     ],
     Exception::GENERAL_ROUTE_NOT_FOUND => [
         'name' => Exception::GENERAL_ROUTE_NOT_FOUND,
-        'description' => 'The requested route was not found. Please refer to the API docs and try again.',
+        'description' => 'Route not found. Please ensure the endpoint is configured correctly and that the API route is valid for this SDK version. Refer to the API docs for more details.',
```
stnguyen90, line 72: Isn't this only used in tests?

ChiragAgg5k: right! i will update it in appwrite repo as well. if this looks good we should keep both consistent

stnguyen90 (approved):

Issue #1064, opened 2025-04-13
🐛 Bug Report: Appwrite packages still uses flutter_web_auth_2-3.1.2 which uses V1 embedding causing compile errors


FrancisIje:
### 👟 Reproduction steps

This causes a compilation errors in the newer flutter versions. 

Launching lib/main.dart on sdk gphone x86 64 in debug mode...
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
e: file:///Users/mac/.pub-cache/hosted/pub.dev/flutter_web_auth_2-3.1.2/android/src/main/kotlin/com/linusu/flutter_web_auth_2/FlutterWebAuth2Plugin.kt:15:48 Unresolved reference 'Registrar'.
e: file:///Users/mac/.pub-cache/hosted/pub.dev/flutter_web_auth_2-3.1.2/android/src/main/kotlin/com/linusu/flutter_web_auth_2/FlutterWebAuth2Plugin.kt:22:33 Unresolved reference 'Registrar'.
e: file:///Users/mac/.pub-cache/hosted/pub.dev/flutter_web_auth_2-3.1.2/android/src/main/kotlin/com/linusu/flutter_web_auth_2/FlutterWebAuth2Plugin.kt:24:39 Unresolved reference 'messenger'.
e: file:///Users/mac/.pub-cache/hosted/pub.dev/flutter_web_auth_2-3.1.2/android/src/main/kotlin/com/linusu/flutter_web_auth_2/FlutterWebAuth2Plugin.kt:24:62 Unresolved reference 'context'.

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':flutter_web_auth_2:compileDebugKotlin'.
> A failure occurred while executing org.jetbrains.kotlin.compilerRunner.GradleCompilerRunnerWithWorkers$GradleKotlinCompilerWorkAction
   > Compilation error. See log for more details

flutter doctor:
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 3.29.0, on macOS 14.5 23F79 darwin-x64, locale en-NG)
[✓] Android toolchain - develop for Android devices (Android SDK version 35.0.0-rc1)
[✓] Xcode - develop for iOS and macOS (Xcode 15.4)
[✓] Chrome - develop for the web
[✓] Android Studio (version 2022.2)
[✓] VS Code (version 1.99.0)
[✓] Connected device (3 available)
[...]

PR #1080, opened 2025-05-22
chore: update flutter_web_auth_2 to 4.x

ChiragAgg5k:
## What does this PR do?

Migration guide followed - https://pub.dev/packages/flutter_web_auth_2#upgrading-to-4x

For now we are keeping the old behaviour and disabling usage of `webView` for linux and windows platforms. But in future we can introduce an option to enable it.

## Related PRs and Issues

- https://github.com/appwrite/sdk-generator/issues/1064
- https://github.com/appwrite/sdk-for-flutter/issues/238
- https://github.com/appwrite/appwrite/pull/9858

yes.

### DIFF
--- a/templates/flutter/lib/src/client_browser.dart.twig
+++ b/templates/flutter/lib/src/client_browser.dart.twig
@@ -224,6 +224,7 @@ class ClientBrowser extends ClientBase with ClientMixin {
   return FlutterWebAuth2.authenticate(
       url: url.toString(),
       callbackUrlScheme: "{{spec.title | caseLower}}-callback-" + config['project']!,
+      options: const FlutterWebAuth2Options(useWebview: false),
     );
   }
 }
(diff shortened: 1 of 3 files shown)


rong6767:
As mentioned in the [issue](https://github.com/ThexXTURBOXx/flutter_web_auth_2/issues/133#issuecomment-2400098993), this upgrade will raise the minimum supported iOS version for the Appwrite SDK to iOS 17.4, which I think might be a potential issue for our clients. To work around this limitation, I forked flutter_web_auth_2 and used this[3.x-without-v1](https://github.com/rong6767/flutter_web_auth_2/tree/3.x-without-v1) branch and made a small modification so I can still use 3.x but don't have any compile errors.

ChiragAgg5k:
@rong6767 thats great that you have found a workaround for your use case, but we can't use the fork with appwrite 😅
this change will lead to a major bump in the sdk so you can keep on using the older appwrite version, and deal with the warnings by using this approach by Steven - https://github.com/appwrite/sdk-for-flutter/issues/238#issuecomment-2878040145

stnguyen90 (approved):

PR #1084, opened 2025-05-31
fix: devKeys support in flutter by conditionally including credentials

ChiragAgg5k:
## What does this PR do?

Fixes devKeys support in flutter.

When devKeys is set, we treat the authentication method as "keys" based, and thus we should not include "credentials: include". This change was also implemented in web and react native, was left in flutter.

the following error is thrown otherwise:
image

yes.


stnguyen90 (approved):
Looks good. To make it better, can you add more context about why this is needed in the PR description?

ChiragAgg5k:
@stnguyen90 done

PR #1095, opened 2025-06-23
fix: add unsupported attribute error

ChiragAgg5k:
## What does this PR do?

1. fix unsupported attribute error.
2. fix duplication of error messages:
	Screenshot 2025-06-23 at 8 35 37 AM
3. fix strict comparision of floating points: `min changed from -1.7976931348623157e+308 to -1.7976931348623157e+308`

## Test Plan
1. Screenshot 2025-06-23 at 9 11 44 AM
---

2. Screenshot 2025-06-23 at 9 12 04 AM

yes.

### DIFF
--- a/templates/cli/lib/commands/init.js.twig
+++ b/templates/cli/lib/commands/init.js.twig
@@ -94,6 +94,11 @@ const initProject = async ({ organizationId, projectId, projectName } = {}) => {
         }
     }
 
+    if (answers.region) {
+        const endpoint = `https://${answers.region}.cloud.appwrite.io/v1`;
+        localConfig.setEndpoint(endpoint);
+    }
+
     if (answers.start === 'new') {
         response = await projectsCreate({
             projectId: answers.id,
--- a/templates/cli/lib/commands/init.js.twig
+++ b/templates/cli/lib/commands/init.js.twig
@@ -95,8 +95,10 @@ const initProject = async ({ organizationId, projectId, projectName } = {}) => {
         }
     }
 
+    localConfig.clear(); // Clear the config to avoid any conflicts
+    const url = new URL("{{ spec.endpoint }}");
+    
     if (answers.start === 'new') {
-        localConfig.clear();
         response = await projectsCreate({
             projectId: answers.id,
             name: answers.project,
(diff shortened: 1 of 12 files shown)


templates/cli/lib/commands/push.js.twig
```diff
@@ -752,17 +783,17 @@ const checkAttributeChanges = (remote, local, collection, recraeting = true) =>
         }
 
         if (changeableKeys.includes(key)) {
-            if (!recraeting) {
-                reason += compareAttribute(remote[key], local[key], reason, key)
+            if (!recreating) {
+                reason = compareAttribute(remote[key], local[key], reason, key)
```
stnguyen90, line 787: don't we want to concat so we keep all the reasons?

templates/cli/lib/utils.js.twig
```diff
@@ -274,10 +274,17 @@ function getUsersPath(action, ids) {
     return path;
 }
 
+function isCloud() {
+    const endpoint = globalConfig.getEndpoint() || "https://cloud.appwrite.io/v1";
```
stnguyen90: we typically try not to use `appwrite.io` in this repo because we don't necessarily want to couple it with appwrite.  Can we use [spec.endpoint](https://github.com/appwrite/appwrite/blob/e56c03a007ba65345b73892f90b0a968cd58a595/src/Appwrite/Platform/Tasks/Specs.php#L304) like we do here?

https://github.com/appwrite/sdk-generator/blob/93ffb24b25b376ca4423e3a5caf6f916673af4b2/templates/web/src/client.ts.twig#L304

templates/cli/lib/commands/init.js.twig
```diff
@@ -95,6 +96,7 @@ const initProject = async ({ organizationId, projectId, projectName } = {}) => {
     }
 
     if (answers.start === 'new') {
+        localConfig.clear();
```
stnguyen90: why do we need this?

Also, if we do this, should we do it for existing project too?

ChiragAgg5k: I saw when when you are overriding a project, the previous config is not cleared so it leads to a mismatch...

and yeah we should probably do it for existing project too.

templates/cli/lib/commands/push.js.twig
ChiragAgg5k: we are already concatenating the reasons here - https://github.com/appwrite/sdk-generator/blob/a6bcc45a90bbd8105692d14673c724c5cda6d081/templates/cli/lib/commands/push.js.twig#L750-L754

concatenating here as well leads to double concat, as shown in the pr screenshot

christyjacob4 (approved):

PR #1096, opened 2025-06-24
add deprecation handling in templates

ChiragAgg5k:
## What does this PR do?

adds warnings and appropriate docs to deprecated methods.

## Related PRs and Issues

- https://github.com/appwrite/appwrite/pull/10047

yes.


templates/android/library/src/main/java/io/package/services/Service.kt.twig
```diff
@@ -38,6 +38,19 @@ class {{ service.name | caseUcfirst }}(client: Client) : Service(client) {
      * @return [{{ method | returnType(spec, sdk.namespace | caseDot) | raw }}]
     {%~ endif %}
      */
+    {%~ if method.deprecated %}
+    {%~ if method.since and method.replaceWith %}
+    @Deprecated(
+        message = "This API is deprecated since {{ method.since }} and uses outdated terminologies. Please use `{{ method.replaceWith | capitalizeFirst }}` instead.",
```
ItzNotABug: Not necessary that every deprecation would be due to outdated terminology. We can use a generic message. Use the sdk label if needed so every method/endpoint has a specific message.

ItzNotABug: Goes for all other templates as well.

templates/cli/lib/commands/command.js.twig
```diff
@@ -59,7 +59,7 @@ const {{ service.name | caseLower }} = new Command("{{ service.name | caseLower
 /**
  * @param {{ "{" }}{{ service.name | caseUcfirst }}{{ method.name | caseUcfirst }}RequestParams{{ "}" }} params
  */
-{%~ block decleration -%}
+{% block decleration %}
```
ItzNotABug: nitpick: has this always been `decleration`? or is it supposed to be `declaration`?

ItzNotABug (changes requested):

templates/cli/lib/commands/command.js.twig
ChiragAgg5k: fixed

src/SDK/SDK.php
```diff
@@ -207,6 +207,22 @@ public function __construct(Language $language, Spec $spec)
             }
             return $value;
         }));
+        $this->twig->addFilter(new TwigFilter('capitalizeFirst', function ($value) {
+            return ucfirst($value);
+        }));
```
abnegate, line 212: This filter already exists, `caseUcfirst`

templates/android/library/src/main/java/io/package/services/Service.kt.twig
abnegate: @ChiragAgg5k Agree, let's remove ` and uses outdated terminologies`

src/SDK/SDK.php
ChiragAgg5k: that removes stuff like `.` from the word

ChiragAgg5k: Screenshot 2025-07-22 at 3 58 37 PM

it made `Tables.listRows` just `TablesListRows` since it also converts to camel

abnegate: Got it :+1: let's rename it to be more clear later

abnegate (approved):

PR #1109, opened 2025-07-16
fix: CLI env variables not pushing

ChiragAgg5k:
## What does this PR do?

1. use `parse` function from dotenv package to get the environment variables
2. add `--with-variables` option to sites command as well
3. verify that when `--with-variables` option is not passed, a function / site deployment uses the old variables already defined, and not overwrite it with blank values

yes.

### DIFF
--- a/templates/cli/lib/commands/push.js.twig
+++ b/templates/cli/lib/commands/push.js.twig
@@ -152,68 +152,6 @@ const awaitPools = {
             iteration + 1
         );
     },
-    wipeFunctionVariables: async (functionId, iteration = 1) => {
-        if (iteration > pollMaxDebounces) {
-            return false;
-        }
-
-        const { total } = await functionsListVariables({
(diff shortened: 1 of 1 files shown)


templates/cli/lib/commands/push.js.twig
```diff
@@ -1213,16 +1246,49 @@ const pushSite = async({ siteId, async, code } = { returnOnZero: false }) => {
...
+
+            await Promise.all(variables.map(async variable => {
+                await sitesDeleteVariable({
+                    siteId: site['$id'],
+                    variableId: variable['$id'],
+                    parseOutput: false
+                });
+            }));
+
+            let result = await awaitPools.wipeSiteVariables(site['$id']);
```
Meldiron: Lets see if we can remove the pooling here, and inf unctions

Meldiron (approved):

PR #1112, opened 2025-07-20
fix: type generation templates to include relative path and strict flag

ChiragAgg5k:
## What does this PR do?

* Type Generation fixes:
	* Fix comments to display user-relative paths instead of absolute paths
	* Fix generation of types of relationships by using relatedCollection's name instead of `$id`
	* Add `--strict` flag to comments when types are generated in strict mode

## Related Issues

- https://github.com/appwrite/sdk-for-cli/issues/182

yes.


templates/cli/lib/commands/types.js.twig
```diff
@@ -117,8 +117,9 @@ const typesCommand = actionRunner(async (rawOutputDirectory, {language, strict})
     const content = templater({
       collections,
       strict,
+      path: rawOutputDirectory ?? '',
       ...templateHelpers,
-      getType: meta.getType
+      getType: meta.getType,
```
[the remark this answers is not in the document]
ChiragAgg5k: it is consistent now

templates/cli/lib/commands/types.js.twig
```diff
@@ -59,19 +59,23 @@ const typesLanguageOption = new Option(
...
 
 const typesStrictOption = new Option(
-  "-s, --strict",
-  "Enable strict mode to automatically convert field names to follow language conventions"
+  "-s, --strict <value>",
+  "Disable strict mode to not automatically convert field names to follow language conventions"
 )
-  .default(false);
+  .choices(["true", "false"])
+  .default("true");
```
Meldiron: I dont think this is common pattern. When true is default, and option for false is given, it's usually `no-X`, like `no-cache`, `no-cache-filter`, or from kubernetes, `no-headers`, `no-wait`, `no-verbose`.

templates/cli/lib/commands/types.js.twig
```diff
@@ -59,19 +59,23 @@ const typesLanguageOption = new Option(
...
   if (language === "auto") {
     language = detectLanguage();
     log(`Detected language: ${language}`);
   }
 
+  strict = strict === "true" || strict === true;
   if (strict) {
     log(`Strict mode enabled: Field names will be converted to follow ${language} conventions`);
+  } else {
+    warn(`Strict mode disabled: Field names will not be converted to follow ${language} conventions`);
```
Meldiron: Hmm also, how does strict even work? If in my collection attribute is `game_id`, and I do `-l ts --strict`, it becomes `gameId` in type, right?

But do we generate functions to translate it? If I do `.listDocuments<Article>()`, and `Article` has `gameId`, how would it know to have value from `game_id` from the response? Also, how would that work with `Query.select()`?

Meldiron: Also, is there some better word for this than "strict"?
Strict type generation suggest to me there is loose type generation as opposide, and for that, I would probably imagine something like NoSQL - when my types are included, but wildcard is allowed.
Which is not the case.

It might be better to allow `--casing="camelCase` `--casing="snake_case"` instead, no matter the language. And for automatic, `--casing="auto"` or `--casing="automatic"`.

OR it could also be `--casing-camel` and `--casing-snake` and `--casing-auto`.
I would look into what others do.

templates/cli/lib/type-generation/languages/dart.js.twig
```diff
@@ -78,19 +79,15 @@ class Dart extends LanguageMeta {
...
     return `<% for (const attribute of collection.attributes) { -%>
 <% if (attribute.type === 'relationship') { -%>
-import '<%- toSnakeCase(attribute.relatedCollection) %>.dart';
+import '<%- toSnakeCase(collections.find(c => c.$id === attribute.relatedCollection).name) %>.dart';
 
 <% } -%>
 <% } -%>
 /// This file is auto-generated by the Appwrite CLI. 
-/// You can regenerate it by running \`appwrite types -l dart ${this.getCurrentDirectory()}\`.
+/// You can regenerate it by running \`appwrite types -l dart <%- strict ? '' : '--strict false ' %><%- path %>\`.
```
Meldiron: It might be better to try and get raw command they ran 🤔 Would that be possible? Just `appwrite ${...rawParms.join(' ')}`, something like that

templates/cli/lib/commands/types.js.twig
ChiragAgg5k: the problem is we already released `strict`... i did not want to remove it and introduce a `no-strict` option 🤔

templates/cli/lib/commands/types.js.twig
ChiragAgg5k: Jake told me today that some languages like ruby, swift etc. do offer mapping of response. so we can in future do the conversion.

but in languages like typescript, that isnt really possible, and user will have to do manual work to make sure typing is mapped correctly

templates/cli/lib/type-generation/languages/dart.js.twig
ChiragAgg5k: i dont think so... 👀

Meldiron: Should be `process.argv`, lets give it 1 try. either it works like a charm, or totally out of expectations.

ChiragAgg5k: that did not work as expected 😅
Screenshot 2025-07-22 at 11 40 47 AM

ChiragAgg5k: donee!

templates/cli/lib/type-generation/languages/typescript.js.twig
```diff
@@ -29,7 +29,8 @@ class TypeScript extends LanguageMeta {
         type = "boolean";
         break;
       case AttributeType.RELATIONSHIP:
-        type = LanguageMeta.toPascalCase(attribute.relatedCollection);
+        const relatedCollection = collections.find(c => c.$id === attribute.relatedCollection);
```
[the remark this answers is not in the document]
ChiragAgg5k: api will throw the error

PR #1116, opened 2025-07-21
fix: update localConfig from appwrite.json to appwrite.config.json

ChiragAgg5k:
## What does this PR do?

since frameworks like vite allow direct import of json files, `appwrite.json` can be confused with the `appwrite` package import. this PR shifts default config to be `appwrite.config.json` while keeping backwards compatibility with appwrite.json

## Related PRs and Issues

- https://github.com/appwrite/sdk-for-cli/issues/172

yes.

### DIFF
--- a/templates/cli/lib/config.js.twig
+++ b/templates/cli/lib/config.js.twig
@@ -125,27 +125,39 @@ class Config {
 }
 
 class Local extends Config {
-    static CONFIG_FILE_PATH = "{{ spec.title|caseLower }}.json";
+    static CONFIG_FILE_PATH = "{{ spec.title|caseLower }}.config.json";
+    static CONFIG_FILE_PATH_LEGACY = "{{ spec.title|caseLower }}.json";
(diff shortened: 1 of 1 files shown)


templates/cli/lib/config.js.twig
```diff
@@ -125,27 +125,39 @@ class Config {
 }
 
 class Local extends Config {
-    static CONFIG_FILE_PATH = "{{ spec.title|caseLower }}.json";
+    static CONFIG_FILE_PATH = "{{ spec.title|caseLower }}.config.json";
+    static CONFIG_FILE_PATH_LEGACY = "{{ spec.title|caseLower }}.json";
```
Meldiron, line 129: As per torsten suggestion, it might be better to use library for this: https://github.com/unjs/c12

that will aslo support .js out of the box, which would be great addition for those wanting a bit of complexity.

I would never try to load such JS/TS configs on my own, trusting lib with 10M monthly downloads is better

PR #1119, opened 2025-07-23
fix: example sdks to use correct platform

ChiragAgg5k:
## What does this PR do?

use 1.7.x for now and use correct platforms.

yes.


example.php
```diff
@@ -53,7 +51,7 @@ function getSSLPage($url) {
     $php
         ->setComposerVendor('appwrite')
         ->setComposerPackage('appwrite');
-    $sdk  = new SDK($php, new Swagger2($spec));
+    $sdk  = new SDK($php, new Swagger2($serverSpec));
```
abnegate: We should not pin languages to specific platforms, the generator is agnostic. PHP being a server-side SDK is an Appwrite specific concept

abnegate (approved):

PR #1147, opened 2025-08-10
fix(dart/templates): tolerant int deserialization in models; make \$s…

sadji-zigadi:
## What does this PR do?
- Fixes Dart model deserialization for integer fields.
- For Dart templates, parses ints as `(map['$field'] as num?)?.toInt()` instead of assigning raw values.
- Special-cases `"$sequence"` to be nullable in generated models to tolerate missing values.

Why:
- Some servers return `"$sequence"` as a string or omit it. Previous code caused `String → int` and `Null → int` runtime errors in generated SDKs.

## Test Plan
- Regenerated the Dart/Flutter SDK from this template.
- Verified generated `Document.fromMap` contains:
  - `$sequence: (map['\$sequence'] as num?)?.toInt(),` (nullable)
  - Other required ints use `?? 0` default when applicable.
- Ran:
  - `dart format .`
  - `dart analyze`
  - `flutter test` on generated SDK
- Manual check: Created documents and listed documents with/without `"$sequence"` and with `"$sequence": "3"`; confirmed no crashes and correct values.

## Related PRs and Issues
- appwrite/sdk-for-flutter#261
- appwrite/appwrite#10253

### Have you read the Contributing Guidelines on issues?
Yes.

### DIFF
--- a/templates/dart/lib/src/models/model.dart.twig
+++ b/templates/dart/lib/src/models/model.dart.twig
@@ -5,7 +5,7 @@ part of '../../models.dart';
 class {{ definition.name | caseUcfirst | overrideIdentifier }} implements Model {
 {% for property in definition.properties %}
     /// {{ property.description }}
-    final {% if not property.required %}{{_self.sub_schema(property)}}? {{ property.name | escapeKeyword }}{% else %}{{_self.sub_schema(property)}} {{ property.name | escapeKeyword }}{% endif %};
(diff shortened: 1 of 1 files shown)


istornz:
IMO it's a migration issue which do not create the required/mandatory new `$sequence` field for existing documents in DB.
In my case I migrated from `1.6.2` to latest `1.7.4`, running migration with no issue.

What did you think @stnguyen90 ?

PR #1210, opened 2025-10-04
  fix: 🐛 Handle null values for integer properties in Dart models

### LINKED ISSUE #1150: 🐛 Bug Report: Appwrite Flutter SDK TypeError: null: type 'Null' is not a subtype of type 'int' in 17.1.0
### 👟 Reproduction steps

I just updated my Flutter app to appwrite 17.1.0 and when I startet it a fetching function caused a TypeError: null: type 'Null' is not a subtype of type 'int'.
After some investigation I couldn't not find an issue in my code. The documents it is fetching don't even have int attributes.
I downgraded back to appwrite 17.0.1 and everything works again.

The affected function just calls `Appwrite().databases.listDocuments(...` and maps the results. But like I said, the objects don't have int attributes.

### 👍 Expected behavior

It should fetch my documents and map the attributes like in appwrite 17.0.1

### 👎 Actual Behavior

Error TypeError: null: type 'Null' is not a subtype of type 'int' being thrown.

### 🎲 Appwrite version

Different version (specify in environment)

### 💻 Operating system

MacOS

### 🧱 Your Environment

I selfhost Appwrite. Currently using 1.6.2. The Flutter app targets web.

### 👀 Have you spent some time to check if this issue has been raised before?

- [x] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

iamngoni:
## What does this PR do?

fix: 🐛 Handle null values for integer properties in Dart models
---
Fixes #1150 - Type 'Null' is not a subtype of type 'int'  by adding explicit type casting for integer properties to properly handle null values for optional & required fields.

## Test Plan

[x] Ran both Dart & Flutter tests (both passed)

## Related PRs and Issues

https://github.com/appwrite/sdk-generator/issues/1150

Yes

### DIFF
--- a/templates/dart/lib/src/models/model.dart.twig
+++ b/templates/dart/lib/src/models/model.dart.twig
@@ -44,6 +44,9 @@ class {{ definition.name | caseUcfirst | overrideIdentifier }} implements Model 
                     List.from(map['{{property.name | escapeDollarSign }}'] ?? [])
                 {%- else -%}
                     map['{{property.name | escapeDollarSign }}']
+                    {%- if property.type == "integer" -%}
+                        {%- if not property.required %}?{% endif %}.toInt()
+                    {%- endif -%}
                     {%- if property.type == "number" -%}
                         {%- if not property.required %}?{% endif %}.toDouble()
                     {%- endif -%}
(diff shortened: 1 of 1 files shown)


ChiragAgg5k:
@iamngoni this issue as well seems to be related to $sequence, can you please verify its an actual issue now? we have reports that the issue has been fixed already since it was a server side change

iamngoni:
Hi @ChiragAgg5k, 

You're right that this is related to $sequence. However, I believe this PR is still valuable as a defensive measure because:
1. The server-side fix only helps users on latest Appwrite instances
2. Users on older self-hosted versions still experience this issue
3. This SDK-side fix would handle null values gracefully for all integer properties

PR #1431, opened 2026-04-08
Ship native CLI binaries via Homebrew and automate formula publishing

ChiragAgg5k:
## Summary
- install platform-specific native CLI binaries via the generated Homebrew formula instead of `npm install`
- keep `--HEAD` installs building from source and tighten Homebrew install detection in `appwrite update`
- replace `sha256 :no_check` with per-platform checksum fields and automate formula version/checksum updates during CLI release publishing

## Why
The CLI already ships native binaries, but Homebrew was still installing the Node/shebang entrypoint. Moving Homebrew to native binaries is the right packaging model, but it also requires the formula to contain real per-asset SHA256 values. This PR updates the generated formula and teaches the publish workflow to write the final release tag and checksums back into the formula after the binaries are uploaded.

## Impact
- `brew install appwrite` installs the native binary for the current macOS or Linux target
- `brew install --HEAD appwrite` still builds a native binary from source with Bun
- published formulas contain real platform-specific SHA256 values instead of skipping integrity checks
- `appwrite update` only treats actual Homebrew-managed paths as Homebrew installs

## Validation
- `php example.php cli console`
- `composer lint-twig`
- `php -l example.php`
- `php -l src/SDK/Language/CLI.php`
- `ruby -c examples/cli/Formula/appwrite.rb`
- `ruby -e 'require "yaml"; YAML.load_file("examples/cli/.github/workflows/publish.yml"); puts "yaml ok"'`
- `cd examples/cli && npm ci && npm run build`
- `cd examples/cli && node dist/cli.cjs --help`
- `cd examples/cli && node dist/cli.cjs update --manual`
- `cd examples/cli && bun run mac-arm64`

### DIFF
--- a/src/SDK/Language/CLI.php
+++ b/src/SDK/Language/CLI.php
@@ -155,6 +155,9 @@ class CLI extends Node
     }
 
     /**
+     * Override a generated Homebrew formula checksum placeholder when a release
+     * build already knows the target binary SHA256.
+     *
      * @param string $key
      * @param string $sha256
      * @return $this
--- a/src/SDK/Language/CLI.php
+++ b/src/SDK/Language/CLI.php
@@ -70,6 +70,10 @@ class CLI extends Node
         'executableName' => 'executable',
         'logo' => '',
         'logoUnescaped' => '',
+        'homebrewMacArm64Sha256' => '0000000000000000000000000000000000000000000000000000000000000000',
+        'homebrewMacX64Sha256' => '0000000000000000000000000000000000000000000000000000000000000000',
+        'homebrewLinuxArm64Sha256' => '0000000000000000000000000000000000000000000000000000000000000000',
+        'homebrewLinuxX64Sha256' => '0000000000000000000000000000000000000000000000000000000000000000',
     ];
 
     /**
@@ -146,6 +150,18 @@ class CLI extends Node
     public function setLogoUnescaped(string $logo): self
     {
         $this->setParam('logoUnescaped', $logo);
+
+        return $this;
+    }
+
+    /**
+     * @param string $key
+     * @param string $sha256
+     * @return $this
+     */
+    public function setHomebrewSha256(string $key, string $sha256): self
+    {
+        $this->setParam($key, $sha256);
 
         return $this;
     }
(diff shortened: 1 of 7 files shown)


templates/cli/.github/workflows/publish.yml
```diff
@@ -66,3 +68,81 @@ jobs:
...
+                  rf'({re.escape(artifact)}"\n\s+sha256 ")([0-9a-f]{{64}})(")'
+              )
+              text, count = pattern.subn(rf'\g<1>{checksum}\g<3>', text, count=1)
+              if count != 1:
+                  raise SystemExit(f"Failed to update checksum for {artifact}")
+
+          formula_path.write_text(text)
+          PY
+
+          ruby -c "$FORMULA_PATH"
```
abnegate, line 127: If we need ruby anyway, better to write the inline script with that instead of pulling in python too

templates/cli/Formula/formula.rb.twig
```diff
@@ -1,19 +1,62 @@
...
   license "{{ sdk.license }}"
-  head "https://github.com/{{ sdk.gitUserName }}/{{ sdk.gitRepoName | caseDash }}.git", branch: "master"
+  version "{{ sdk.version }}"
+
+  def self.binary_arch
+    Hardware::CPU.arm? ? "arm64" : "x64"
+  end
+
+  def self.binary_os
+    OS.mac? ? "darwin" : "linux"
```
abnegate: What about windows?

src/SDK/Language/CLI.php
```diff
@@ -150,6 +154,18 @@ public function setLogoUnescaped(string $logo): self
         return $this;
     }
 
+    /**
+     * @param string $key
+     * @param string $sha256
+     * @return $this
+     */
+    public function setHomebrewSha256(string $key, string $sha256): self
```
abnegate, line 165: How does this work? Don't we need to generate first to build to get the SHA?

templates/cli/Formula/formula.rb.twig
ChiragAgg5k: Handled this explicitly in `4b12fe032`. The Homebrew formula now only maps supported Homebrew targets (`macOS` and `Linux`) and raises on any unsupported OS instead of implicitly falling through to Linux.

src/SDK/Language/CLI.php
ChiragAgg5k: Clarified the flow in `4b12fe032`. The generator keeps placeholder SHA256 values by default, `example.php` only overrides them when env vars are already available, and the generated publish workflow rewrites the formula after the release binaries are built, which is when the real SHAs exist.

templates/cli/.github/workflows/publish.yml
ChiragAgg5k: Updated this in `4b12fe032`. The checksum updater now uses Ruby inline, which keeps the workflow on the tooling it already needs for `ruby -c` instead of pulling in Python as well.

abnegate (approved):

PR #1436, opened 2026-04-11
fix: Flutter Realtime compatibility with Appwrite 1.9

SticksDev:
## What does this PR do?

Fixes Flutter SDK Realtime compatibility with Appwrite 1.9 and improves error diagnostics.

1. **Removed forced `Query.select(["*"])` injection** - The SDK was automatically appending `{"method":"select","values":["*"]}` to every subscription URL when no explicit queries were provided. `select` is not a supported Realtime filter type on Appwrite 1.9, causing the server to immediately close the connection. Now skips query params entirely when none are provided.

2. **Fixed subscriptions map cast crash** - The SDK expected the server's `connected` message to contain `subscriptions` as a `Map<String, dynamic>`. Appwrite 1.9 sends it as an empty `List` (`[]`), causing a hard cast crash. Now guarded with an `is Map` type check.

3. **Added channel-based event routing fallback** - The SDK routes incoming events by matching `subscriptionId` to slot via a map populated from the `connected` message. Since Appwrite 1.9 does not assign subscription IDs, that map is always empty and every event was silently dropped. Added a fallback that routes events by matching the event's `channels` array against each subscription's registered channels.

4. **Added disconnect/error logging** - onDone` now prints the WebSocket close code and reason. `onError` prints the stream error. `handleError` prints the server error frame code and message.

## Test Plan

- Subscribe to a Realtime channel against an Appwrite 1.9 server
- Verify the WebSocket connection is not immediately closed (fix 1)
- Verify no cast exception on the `connected` message (fix 2)
- Verify events are received by the subscription callback (fix 3)
- Disconnect the server or trigger an error and verify diagnostic output appears in debug console (fix 4)

## Related PRs and Issues

- Related to Appwrite 1.9 Realtime compatibility

Yes

### DIFF
--- a/templates/flutter/lib/src/realtime_mixin.dart.twig
+++ b/templates/flutter/lib/src/realtime_mixin.dart.twig
@@ -100,9 +100,11 @@ mixin RealtimeMixin {
             
             // Store subscription ID mappings from backend
             // Format: { "0": "sub_a1f9", "1": "sub_b83c", ... }
+            // Appwrite 1.9 sends subscriptions as an empty List instead of a Map,
+            // so guard with a type check to avoid a cast crash.
             _slotToSubscriptionId.clear();
             _subscriptionIdToSlot.clear();
-            if (data.data['subscriptions'] != null) {
+            if (data.data['subscriptions'] is Map) {
               final subscriptions = data.data['subscriptions'] as Map<String, dynamic>;
               subscriptions.forEach((slotStr, subscriptionId) {
                 final slot = int.tryParse(slotStr);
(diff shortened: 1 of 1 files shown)


SticksDev:
#2 and #3 might be an upstream issue, but had to fix these nonetheless incase the server didn't return them.

SticksDev:
thank you GitHub for not co-authoring @JakeyPrime - he helped a tooooone with this, including fighting with the server-side and client side issues and problems. This PR would not be possible with out him.

ArnabChatterjee20k:
hey @SticksDev can you confirm are you using 1.8.0 or 1.9.0?
In 1.9.0 -> 
queries are mandatory for slotting
Subscriptions are always sent as a map

Can you try the flutter sdk against cloud to be sure?
Thanks for the help

SticksDev:
Our server is on 1.9.0 (not 1.9.1 yet) and we can confirm these issues exist. I didn't have these issues on cloud though when testing, so likely something upstream? Maybe 1.9.1 fixes it?

ChiragAgg5k:
@SticksDev yes likely the case the issue was fixed after 1.9.0 release. 1.9.1 is coming right up, please wait that to release

PR #1458, opened 2026-04-17
Strip markdown syntax from CLI help descriptions

ChiragAgg5k:
## Summary

- CLI commands render method/parameter descriptions as help text, but those descriptions come from the OpenAPI spec and contain markdown (links, bold) that terminals don't render — so users see raw syntax like `[Learn more about queries](https://appwrite.io/docs/queries)` and `**userId**` in `appwrite <service> -h` output.
- Added a `stripMarkdown` Twig filter in `src/SDK/SDK.php` that converts `[text](url)` → `text` and removes `**bold**` / `__bold__` markers.
- Applied the filter to `method.description` and `parameter.description` in `templates/cli/lib/commands/services/services.ts.twig` so CLI help output becomes plain text.

### Before

```
list-templates [options]   List available site templates. You can use template details in [createSite](/docs/references/cloud/server-nodejs/sites#create) method.
--queries [queries...]     Array of query strings ... [Learn more about queries](https://appwrite.io/docs/queries). Maximum of 100 queries ...
```

### After

```
list-templates [options]   List available site templates. You can use template details in createSite method.
--queries [queries...]     Array of query strings ... Learn more about queries. Maximum of 100 queries ...
```

The filter only touches the CLI template — other SDKs keep markdown in descriptions since they render it in docs/comments where markdown is fine.

## Test plan

- [x] Regenerate CLI SDK: `docker run --rm -v $(pwd):/app -w /app php:8.3-cli php example.php cli`
- [x] Verify no remaining `[text](url)` or `**bold**` in `examples/cli/lib/commands/services/*.ts`
- [x] Twig lint: `uvx djlint templates/cli/lib/commands/services/services.ts.twig --lint` (0 errors)
- [ ] Spot-check `appwrite sites -h`, `appwrite account -h`, `appwrite databases -h` in the built CLI

### DIFF
--- a/src/SDK/SDK.php
+++ b/src/SDK/SDK.php
@@ -256,8 +256,22 @@ class SDK
             if ($value === null) {
                 return '';
             }
-            // Convert markdown links [text](url) -> text
-            $value = preg_replace('/\[([^\]]+)\]\([^)]+\)/', '$1', $value);
+            // Convert markdown links.
+            // Absolute URLs (http/https) are preserved as "text (url)" so users
+            // can copy or click them; relative links like "/docs/..." are
+            // useless in a terminal, so we drop the URL and keep just the text.
+            $value = preg_replace_callback(
+                '/\[([^\]]+)\]\(([^)]+)\)/',
+                function ($m) {
+                    $text = $m[1];
+                    $url = trim($m[2]);
+                    if (preg_match('/^https?:\/\//i', $url)) {
+                        return $text . ' (' . $url . ')';
+                    }
+                    return $text;
+                },
+                $value
+            );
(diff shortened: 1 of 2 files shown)


ChiragAgg5k:
Good catch — pushed 1333afe33. Switched both bold regexes from `[^*]+`/`[^_]+` to lazy `(.+?)` so inner characters (e.g. `__user_id__`) match correctly. Added unit-style checks against a handful of inputs (`__user_id__`, `**with*star**`, adjacent spans like `__a__ and __b__`, relative vs. absolute links, null, empty); all pass. Regenerated CLI output has zero remaining `**..**` / `__..__` / `[..](..)` residue.

PR #1484, opened 2026-04-28
Fix CLI Homebrew tap updates

ChiragAgg5k:
## Summary

- Detect the actually installed Homebrew formula and use it consistently for latest-version checks, automatic updates, and manual instructions.
- Warn when the installed formula is not the official Appwrite tap, recommend the tap for faster native binaries, and keep updating the currently installed formula.
- Avoid automatic formula migration; users get an explicit migration command instead.

## Verification

- `docker run --rm -v $(pwd):/app -w /app php:8.3-cli php example.php cli`
- `composer lint-twig`
- `bun run build:types` in `examples/cli`

### DIFF
--- a/templates/cli/lib/commands/update.ts
+++ b/templates/cli/lib/commands/update.ts
@@ -11,6 +11,7 @@ import {
   getLatestVersionForInstallation,
   compareVersions,
   getErrorMessage,
+  getInstalledHomebrewFormula,
 } from "../utils.js";
 import {
   EXECUTABLE_NAME,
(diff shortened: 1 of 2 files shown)


ChiragAgg5k:
Addressed review comments:\n\n- The destructive uninstall/install migration path is no longer present. For non-Appwrite-tap Homebrew installs, the CLI now warns, suggests the explicit tap install command, and continues with `brew upgrade <installed-formula>`.\n- Added bounded timeouts for Homebrew `execFileSync` calls, including `brew list --formula --full-name`, while preserving shorter update-check timeouts when provided.\n\nRe-verified with CLI regeneration, `composer lint-twig`, and `bun run build:types` in `examples/cli`.

PR #1489, opened 2026-04-29
added presence api update

ArnabChatterjee20k:
## What does this PR do?

(Provide a description of what this PR does.)

## Test Plan

## Related PRs and Issues

(Write your answer here.)


mock-server/app/http.php
```diff
@@ -39,23 +38,24 @@
...
-    host: '0.0.0.0',
-    port: App::getEnv('PORT', 80),
-    mode: SWOOLE_PROCESS
-);
 $payloadSize = 6 * (1024 * 1024); // 6MB
 $workerNumber = swoole_cpu_num() * intval(App::getEnv('_APP_WORKER_PER_CORE', 6));
 
-$http->set([
-    'worker_num' => $workerNumber,
+$adapter = new Adapter\Swoole(host: '0.0.0.0', port: (int) App::getEnv('PORT', 80));
```
ArnabChatterjee20k: using http and websocket server in a single file. Forwarding request to the http App via the onRequest

mock-server/src/Utopia/Realtime/Protocol.php
```diff
@@ -0,0 +1,186 @@
+<?php
+
+namespace Utopia\MockServer\Utopia\Realtime;
+
+use Utopia\MockServer\Utopia\Exception;
+use Utopia\WebSocket\Server;
+
```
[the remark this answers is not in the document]
ArnabChatterjee20k: its already fixed

templates/web/src/services/realtime.ts.twig
```diff
@@ -562,9 +627,6 @@ export class Realtime {
             case this.TYPE_PONG:
                 // Handle pong response if needed
                 break;
-            case this.TYPE_RESPONSE:
```
ArnabChatterjee20k, line 565: removed this cause the handleResponseAction was doing nothing. Not needed as on type response we dont store any internal state

ArnabChatterjee20k:
@greptile `The mock server's handlePresence sends a type: 'response' acknowledgment frame that every SDK client now silently discards — the TYPE_RESPONSE case was removed from all switch statements in this same PR `
They are not out of sync. rather the mock server is trying to replicate the original appwrite realtime server as much as possible. for the `response` type message, the sdks doesn't perform anything -> no change in internal state. so nothing to perform here. From the beginning, the handler was a function with empty body

ArnabChatterjee20k:
@greptile 
> tests/languages/web/node.js (presence test always prints passed without executing)
nodejs a backend one, we dont support presence realtime test there

ArnabChatterjee20k:
@greptile 
> server/src/Utopia/Realtime/Protocol.php (query-filtered subscriptions never receive the test event)

That's actually intentional — and it's what makes the 'Realtime failed!' expected output in Base::REALTIME_RESPONSES work. Walking through the contract:

  Subscription set up by the tests:

  ┌─────────────────────────┬───────────────────────────────────────────────┬─────────────────────────────────────────────────────────┐
  │         Handle          │                    Queries                    │                   What it's asserting                   │
  ├─────────────────────────┼───────────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
  │ rtsub                   │ none                                          │ "no filter → receives the event"                        │
  ├─────────────────────────┼───────────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
  │ rtsubWithQueries        │ equal('response', ['WS:/v1/realtime:passed']) │ "query that matches the payload → receives the event"   │
  ├─────────────────────────┼───────────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
  │ rtsubWithQueriesFailure │ equal('response', ['failed'])                 │ "query that doesn't match → does NOT receive the event" │
  └─────────────────────────┴───────────────────────────────────────────────┴─────────────────────────────────────────────────────────┘

  What the mock does today (Protocol.php):

  $eventPayload = ['response' => 'WS:/v1/realtime:passed'];
  if (!$this->subscriptionMatchesPayload($queries, $eventPayload)) {
      continue;   // ← this is the "never receives" you flagged
  }
  $this->send($server, $connection->fd, 'event', [...]);

  For the failure subscription, subscriptionMatchesPayload(['{"method":"equal","attribute":"response","values":["failed"]}'], ['response' => 'WS:/v1/realtime:passed']) → equal('response', 
  ['failed']) → in_array('WS:/v1/realtime:passed', ['failed'], true) → false. So the continue fires and no event is sent.

  That's by design:
  - The test fixtures explicitly want the failure subscription to time out — both the Flutter Completer-based read and Web/Android/Apple's pre-initialized "Realtime failed!" variable
  depend on the callback not firing.
  - Base::REALTIME_RESPONSES order is ['passed', 'passed', 'failed', ...], asserting the failure subscription prints the un-changed initial string.
  - It matches real Appwrite realtime semantics — queries filter events, so a subscription whose query doesn't match an event simply doesn't see it.

  If the filter were inverted (failure sub also got the event), here's what would break in each SDK:

  - Web/Android/Apple: their callback would fire and overwrite responseRealtimeWithQueriesFailure to 'WS:/v1/realtime:passed', then the test would print that instead of 'Realtime failed!'
  → mismatch at the failure-output index.
  - Flutter: the rtsubWithQueriesFailureFirst Completer would complete with the message → the try branch's print("Realtime failed!") would still fire (both branches print the same string,
  so this one wouldn't observably regress — but the TimeoutException path would never run, which defeats the intent of the test).

  So the filter-out behavior is load-bearing for the test's "failed" assertion. It looks like a missing feature but it's actually what proves query filtering works.

ArnabChatterjee20k:
@greptile
> Multiple template files still generate syntactically invalid code for $-prefixed additionalPropertiesKey values in Python, Ruby, Rust, PHP, and .NET — issues raised in earlier review rounds that remain unresolved.

The core presence logic is sound and consistently implemented across platforms. However, additionalPropertiesKey changes in five language templates do not apply removeDollarSign in identifier positions, so any spec using a $-prefixed key will produce uncompilable generated code. The composer.lock also locks utopia-php/query at PHP 8.4+, stricter than every other dependency.

templates/python/package/models/model.py.twig, templates/ruby/lib/container/models/model.rb.twig, templates/rust/src/models/model.rs.twig, templates/php/src/Models/Model.php.twig, templates/dotnet/Package/Models/Model.cs.twig — all generate invalid identifiers when additionalPropertiesKey starts with $
removeDollarSign is already applied in every identifier position in all five templates:

  - templates/python/package/models/model.py.twig — every reference to additionalPropertiesKey used as an identifier (lines 73, 97, 100,
  101, 103, 104, 105, 109, 110, 112, 115) is piped through | caseSnake | removeDollarSign. The three usages at lines 111, 113, 115 that
  omit the filter are string keys in result['…'] for the outgoing JSON dict — they must keep the $ because that's the wire format.
  - templates/ruby/lib/container/models/model.rb.twig — every identifier usage (lines 11, 20, 35, 46, 58, 65) chains | caseSnake | 
  escapeKeyword | removeDollarSign. Lines 46 and 58 use the raw form only in "…": JSON string-key positions.
  - templates/rust/src/models/model.rs.twig — all four usages (lines 29, 60, 64, 65) chain | caseSnake | removeDollarSign. There is no
  non-identifier usage.
  - templates/php/src/Models/Model.php.twig — every usage including the @param ${{…}} docblock, the constructor property, the named
  argument, and the property access (lines 55, 66, 143, 164) chains | caseCamel | removeDollarSign.
  - templates/dotnet/Package/Models/Model.cs.twig — every identifier usage (lines 20, 29, 37, 84, 97, 103) chains | caseUcfirst | 
  escapeKeyword | removeDollarSign or | caseCamel | escapeKeyword | removeDollarSign. Line 97's bare form is inside a "…" string key.

src/Spec/Swagger2.php
```diff
@@ -598,7 +598,8 @@ public function getDefinitions(): array
                 'properties' => $schema['properties'] ?? [],
                 'description' => $schema['description'] ?? '',
                 'required' => $schema['required'] ?? [],
-                'additionalProperties' => $schema['additionalProperties'] ?? []
+                'additionalProperties' => $schema['additionalProperties'] ?? [],
```
ArnabChatterjee20k, line 601: to switch between 'data' and 'metadata' via the specs. eg; database service uses data and presences use metadata

templates/android/library/src/main/java/io/package/services/Realtime.kt.twig
```diff
@@ -34,43 +34,47 @@ class Realtime(client: Client) : Service(client), CoroutineScope {
         private const val TYPE_ERROR = "error"
         private const val TYPE_EVENT = "event"
         private const val TYPE_PONG = "pong"
-        private const val TYPE_RESPONSE = "response"
```
ArnabChatterjee20k, line 37: removed this, as we dont handle the response types directly and from the start an empty function with no body was present. so removed it

abnegate (approved):

PR #1530, opened 2026-05-12
Fix CLI init project config resolution

ChiragAgg5k:
## What

Fixes CLI local config resolution so `appwrite init project` links the current directory instead of reusing and overwriting an ancestor config.

## Why

`init project` is a creation/linking command for the directory where it is run. It should not treat an ancestor `appwrite.config.json` as the current directory's link target. At the same time, existing operational commands should keep working from project subdirectories.

## How

- Adds `Local.useCwdConfig()` to retarget the shared `localConfig` singleton to only `./appwrite.config.json` or local legacy `./appwrite.json`.
- Calls `useCwdConfig()` at the start of `initProject()` before prompts inspect existing project state.
- Keeps ancestor lookup for other commands, but stops traversal before `$HOME` unless the command is run directly in `$HOME`, preventing stale home-dir configs from hijacking unrelated projects.

## Testing

- `docker run --rm -v $(pwd):/app -w /app php:8.3-cli php example.php cli`
- `composer lint-twig`
- `git diff --check`
- `npm ci && npm run build:types` in `examples/cli`

### DIFF
--- a/templates/cli/lib/config.ts
+++ b/templates/cli/lib/config.ts
@@ -509,6 +509,10 @@ class Config<T extends ConfigData = ConfigData> {
     }
   }
 
+  protected usePath(path: string): void {
+    (this as { path: string }).path = path;
+  }
+
   read(): void {
     try {
       const file = fs.readFileSync(this.path).toString();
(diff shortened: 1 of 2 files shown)


ChiragAgg5k:
Reviewed Greptile's latest P1. The global removal is intentional for this PR: local config resolution should only check the directory where the command is run, including for push/deploy/init resource commands. The older inline comments about `useCwdConfig()` and mutable `Config.path` are outdated after the branch was revised.

ChiragAgg5k:
@greptileai

ChiragAgg5k:
@greptileai

lohanidamodar (approved):

templates/cli/lib/config.ts
```diff
@@ -509,6 +509,10 @@ class Config<T extends ConfigData = ConfigData> {
     }
   }
 
+  protected usePath(path: string): void {
+    (this as { path: string }).path = path;
+  }
```
TorstenDittmann, line 514: Why this type annotation?

PR #1553, opened 2026-05-21
Fix missing body property defaults in Swagger parser

ChiragAgg5k:
## What changed

This fixes a PHP warning in `Swagger2::parseMethod()` when parsing body schema properties that do not define a `default` value:

```text
Warning: Undefined array key "default" in src/Spec/Swagger2.php
```

Swagger/OpenAPI schema properties are valid without a `default` key. The parser already handles missing defaults for top-level parameters using `?? null`, but the body-property path accessed `$value['default']` directly in two places.

## Fix details

- Read the body property default into a local `$default` variable using `$value['default'] ?? null`.
- Preserve the existing object-default behavior by converting empty object defaults to `stdClass` before JSON encoding.
- Use a safe type check with `($value['type'] ?? null)` so properties missing `type` do not trigger similar notices.
- Continue setting parsed parameter defaults to `null` when no default is provided, matching the behavior used for non-body parameters.

## Validation

- `vendor/bin/phpcs src/Spec/Swagger2.php`
- `git --no-pager diff --check`

### DIFF
--- a/src/Spec/Swagger2.php
+++ b/src/Spec/Swagger2.php
@@ -292,8 +292,9 @@ class Swagger2 extends Spec
                             'type' => $value['items']['type'] ?? '',
                             'model' => isset($value['items']['$ref']) ? str_replace('#/definitions/', '', $value['items']['$ref']) : null,
                         ];
-                        if ($value['type'] === 'object' && is_array($value['default'])) {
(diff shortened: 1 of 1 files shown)


ChiragAgg5k:
Addressed Greptile's inline suggestion: the object-default check now uses the already-normalized temp type value instead of re-reading the raw value type.

PR #1570, opened 2026-06-04
Support ProjectQuery security in generated SDKs

ChiragAgg5k:
## What changed

This updates the SDK generator to support security schemes that should be sent as URL query parameters, specifically the new `ProjectQuery` scheme used by OAuth approve/authorize/reject methods.

The parser now resolves method security into two flattened method-level collections:

- `method.securityHeaders` for method-scoped security headers such as `Project`
- `method.securityQueries` for method-scoped security query params such as `ProjectQuery`

The old parsed `method.security` output has been removed from template context so SDK templates no longer need nested security-node loops or scheme-specific checks.

## Why

OAuth methods can require the project ID in the request URL rather than as `X-Appwrite-Project`. Previously SDKs treated project as a global header through `setProject`, which meant these methods could not follow specs that declare only `ProjectQuery` security.

With this change:

- `setProject` stores the project ID in client config
- methods declaring `Project` send `X-Appwrite-Project`
- methods declaring `ProjectQuery` append `project=...` to the request URL
- if a future method declares both schemes, both transports are generated from the spec

## SDK/template impact

Updated request generation templates across the supported clients to use the flattened security fields:

- Web, Node, React Native, Deno
- PHP, Python, Ruby
- Dart, Flutter
- Go
- Swift, Apple
- Kotlin, Android
- .NET, Unity

Client setters now keep project in config without registering it as a global header. Other global headers keep the existing behavior.

Request clients that append GET params were also adjusted where needed so existing URL query strings are preserved instead of overwritten.

## Docs impact

REST and GraphQL HTTP example templates now render security according to transport:

- query security appears in the request URL, e.g. `?project=<YOUR_PROJECT_ID>`
- header security appears as an HTTP header, e.g. `X-Appwrite-Project: <YOUR_PROJECT_ID>`

Generated REST OAuth examples now show project in the URL instead of as a fake `project:` header.

## Validation

- Ran `composer lint-twig`
  - `565` Twig files linted
  - `0` errors
- Regenerated affected SDK examples from the saved client spec:
  - web, node, php, python, ruby, dart, flutter, react-native, go, swift, apple, dotnet, android, kotlin, unity
  - REST, GraphQL, Markdown docs
- Verified generated OAuth methods append `project` as a query parameter.
- Verified normal secured methods still send `X-Appwrite-Project`.
- Verified no remaining `method.security` references in `src` or `templates`.

## Notes

`global` is internal generator metadata. `Project` is marked non-global because the project ID is stored on the client, but each method's security declaration decides whether it is sent as `X-Appwrite-Project` or as a `ProjectQuery` URL parameter.

### DIFF
--- a/src/Spec/Swagger2.php
+++ b/src/Spec/Swagger2.php
@@ -161,10 +161,21 @@ class Swagger2 extends Spec
         $methodSecurity = $method['security'][0] ?? [];
 
         foreach ($methodAuth as $i => $node) {
-            $methodAuth[$i] = (array_key_exists($i, $security)) ? $security[$i] : [];
+            $methodAuth[$i] = (array_key_exists($i, $security)) ? [...$security[$i], 'global' => $i !== 'Project'] : [];
         }
         foreach ($methodSecurity as $i => $node) {
-            $methodSecurity[$i] = (array_key_exists($i, $security)) ? $security[$i] : [];
+            $methodSecurity[$i] = (array_key_exists($i, $security)) ? [...$security[$i], 'global' => $i !== 'Project'] : [];
+        }
+
+        $methodSecurityHeaders = [];
+        $methodSecurityQueries = [];
+        foreach ($methodSecurity as $i => $node) {
+            if (($node['in'] ?? '') === 'header' && !($node['global'] ?? false)) {
+                $methodSecurityHeaders[$i] = $node;
+            }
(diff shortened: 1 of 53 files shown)


ChiragAgg5k:
Closing this fork-based PR so the branch can be recreated from appwrite/sdk-generator and run CI with repository secrets available.

PR #1576, opened 2026-06-06
fix(unity): ship SDK as a UPM package with .meta files

ChiragAgg5k:
## Problem

The Unity SDK installs via UPM but is completely non-functional: every asset is ignored with

> Asset Packages/io.appwrite.unity/… has no meta file, but it's in an immutable folder. The asset will be ignored.

Two root causes:

1. **Missing `.meta` files.** A UPM package consumed via git URL is **immutable**, and Unity does **not** generate `.meta` files for immutable packages — they must already exist in the repo. The generator only emitted metas for the `.asmdef` and bundled `.dll` files (8 of ~70 assets), so every `.cs` file, every folder, `package.json`, etc. was ignored.
2. **Project layout, not package layout.** `package.json` shipped under `Assets/`, forcing the awkward `?path=Assets` install URL.

This was never caught because the Unity PlayMode test generates a **mutable** project (under `Assets/`), where Unity auto-generates metas — so the test environment structurally can't reproduce the immutable-package failure.

## Fix

- Add a generic `postGenerate(string $target)` hook on `Language` (no-op) invoked after generation in `SDK::generate()`.
- `Unity::postGenerate()` sweeps the generated package and writes a deterministic `.meta` for every file/folder missing one:
  - GUIDs derived from the package-relative path → stable across regenerations (clean diffs).
  - Correct importer per type (`MonoImporter` for `.cs`, `folderAsset` for dirs, WebGL-enabled `PluginImporter` for `.jslib`, `DefaultImporter` otherwise).
  - Skips `~` folders (`Samples~`, `docs~`) and existing metas (asmdef/dll GUIDs + import settings preserved).
- Ship the package at the **repo root** in publish mode (strip the `Assets/` prefix) so it installs from the bare git URL. **Test mode keeps the `Assets/` project layout**, so the PlayMode test is unchanged.
- Drop `?path=Assets` from the README install instructions.

## Verification

Local generation was checked against every invariant Unity enforces at import (25/25):

- All importable files **and** folders have a sibling `.meta` (72/72).
- GUIDs valid 32-hex and globally unique; every `.meta` is valid YAML with `fileFormatVersion: 2`.
- Importer type correct per asset; `WebGLCookies.jslib` enabled on WebGL.
- Hand-authored asmdef/dll metas preserved byte-for-byte.
- Fully deterministic across regenerations.
- Test-mode output byte-for-byte unchanged (Assets/ layout + ProjectSettings/Packages intact, sweep skipped).
- PHP and Go generation unaffected by the new hook.

## Follow-ups (outside this repo)

- The publish pipeline copies `Assets/` today — it must copy the **repo root** for the new layout to reach the published `sdk-for-unity` repo.
- `appwrite/sdk-for-unity` needs regeneration + republish; the currently-published package stays broken until then.

### DIFF
--- a/src/SDK/Language.php
+++ b/src/SDK/Language.php
@@ -57,6 +57,20 @@ abstract class Language
     abstract public function getFiles(): array;
 
     /**
+     * Hook invoked once after all files have been generated.
+     *
+     * Languages can override this to run post-processing over the
+     * generated output (e.g. emitting sidecar files that depend on the
+     * full file tree). Default implementation is a no-op.
+     *
+     * @param string $target Absolute path the SDK was generated into.
+     * @return void
+     */
+    public function postGenerate(string $target): void
+    {
+    }
+
+    /**
      * @param array $parameter
      * @return string
      */
--- a/src/SDK/Language/Unity.php
+++ b/src/SDK/Language/Unity.php
@@ -586,6 +586,13 @@ class Unity extends DotNet
             return implode("\n", $lines) . "\n";
         }
 
+        // Plugins/Android/AndroidManifest.xml must import as an Android
+        // plugin so Unity merges it into the build manifest; a plain
+        // DefaultImporter would import it as an inert text asset.
+        if (str_ends_with($relativePath, 'Plugins/Android/AndroidManifest.xml')) {
+            return implode("\n", $this->pluginImporterMeta($guid, 'Android')) . "\n";
+        }
+
         $extension = strtolower(pathinfo($relativePath, PATHINFO_EXTENSION));
 
         $lines = match ($extension) {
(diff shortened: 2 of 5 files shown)


ChiragAgg5k:
Thanks for the review — both addressed in `8ca0acb7b`:

- **P1 (silent write failure):** `postGenerate()` now throws a `RuntimeException` if `file_put_contents()` returns `false`, so an incomplete package can't be reported as a success.
- **P2 (path robustness):** the target is normalized once (`rtrim(str_replace('\\\\', '/', $target), '/')`) and the asset path is slash-normalized before the `substr`, so the relative path / GUID is correct regardless of separator style or a trailing slash.

Re-verified locally: 0 missing metas and GUIDs unchanged (normalization didn't shift any). `build (8.3, Unity2021)` is green.

PR #1581, opened 2026-06-09
Fix JS language override property fatal

ChiragAgg5k:
## What

- Remove the invalid `#[Override]` attribute from `JS::$params`, which PHP rejects because `Override` only targets methods in the runtime used by CI.
- Skip Rector’s `AddOverrideAttributeToOverriddenPropertiesRector` rule so `composer refactor:check` does not reintroduce the fatal attribute.

## Why

Web SDK generation with `VERSION="1.9.x"` failed during class loading before generation could start:

```
Attribute "Override" cannot target property (allowed targets: method)
```

## Test Plan

- `php -l src/SDK/Language/JS.php`
- `php -l rector.php`
- `composer refactor:check`
- `composer lint-twig`
- `VERSION="1.9.x" php example.php web`

### DIFF
--- a/rector.php
+++ b/rector.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 use Rector\CodeQuality\Rector\Catch_\ThrowWithPreviousExceptionRector;
 use Rector\Config\RectorConfig;
 use Rector\Php84\Rector\MethodCall\NewMethodCallWithoutParenthesesRector;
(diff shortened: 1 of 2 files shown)


ChiragAgg5k:
The `use Override;` import is still required: `JS::getFilters()` is annotated with `#[Override]` later in the same file. `composer refactor:check` also passes without producing a pending diff.

PR #1653, opened 2026-07-10
fix: pin CLI template @appwrite.io/console to 15.3.0 and dedupe generated query flags

ChiragAgg5k:
This PR carries the fixes needed for the full validation matrix to pass against the current 1.9.x specs (synced in appwrite/specs#86). Four independent issues, all surfaced by the new spec:

## 1. Pin `@appwrite.io/console` to `15.3.0` in the CLI template

The CLI depends on the console web SDK. The template pinned a caret range:

```diff
-    "@appwrite.io/console": "^15.1.1",
+    "@appwrite.io/console": "15.3.0",
```

`sdk-for-cli` is already pinned to exactly `15.3.0` ([appwrite/sdk-for-cli@f232e97](https://github.com/appwrite/sdk-for-cli/commit/f232e97)), so every regeneration reverted it back to `^15.1.1` — which can resolve to console SDKs missing the new APIs (organization memberships, OAuth2 PAR, TablesDB migrations, addons) and fail the build with errors like:

```
error: No matching export in "node_modules/@appwrite.io/console/dist/esm/sdk.js" for import "Mongo"
```

`package-lock.json.twig` and `bun.lock.twig` were regenerated via `scripts/update-lockfiles.sh cli` (with npm 11 — an earlier regeneration with npm 10 dropped the `libc` fields on optional native-binary entries, caught by Greptile and fixed).

An exact pin keeps the template byte-identical with the generated repo and makes console bumps an explicit, reviewed change here.

## 2. CLI: generated query flags collided with spec-declared parameters

The CLI generator adds convenience flags (`--filter`, `--sort-*`, `--limit`, `--offset`, `--cursor-*`, `--select`) to any command whose method has a `queries` array parameter. The new cloud `usage` endpoints (`listEvents`, `listGauges`) declare **literal** `limit`/`offset` parameters *alongside* `queries`, so the generated command registered the same option twice and Commander threw at startup:

```
error: "limit" cannot be bound multiple times in the same parameter list
```

The generated flag groups are now tracked in a constant and a group is skipped when any of its option names collides with a parameter the method already declares — the spec-declared options win:

```php
private const array QUERY_FLAG_PARAMS = [
    'filtering' => ['filter', 'where', 'sortAsc', 'sortDesc', 'cursorAfter', 'cursorBefore'],
    'pagination' => ['limit', 'offset'],
    'select' => ['select'],
];
```

```php
$collides = fn (string $group): bool => array_intersect(self::QUERY_FLAG_PARAMS[$group], $parameterNames) !== [];
...
$hasPaginationQueries = $hasQueries && !$hasSelectionOnlyQueries && !$collides('pagination');
```

The same constant now also feeds the `buildQueries` builder params, removing the previously duplicated hardcoded lists.

## 3. Python: service/model name collision broke generic responses

The new `organization` **service** returns the `organization` **model**. The service template aliases the import on collision:

```python
from ..models.organization import Organization as OrganizationModel
```

…but the generic-response paths still emitted the bare name in the return annotation and the response parsing:

```python
def get(self, model_type: Type[T] = dict) -> Organization[T]:   # NameError
    ...
    return Organization.with_data(response, model_type)          # resolves to the service class
```

On Python ≤ 3.13, return annotations are evaluated at function-definition time, so importing the module raised:

```
NameError: name 'Organization' is not defined
```

(this was the `python (server)` validation failure — it doesn't reproduce on Python 3.14, which defers annotation evaluation). The collision alias is now applied in all three places (return annotation, docstring, `.with_data()` call):

```python
def get(self, model_type: Type[T] = dict) -> OrganizationModel[T]:
    ...
    return OrganizationModel.with_data(response, model_type)
```

## 4. Dart: enum properties in nested model-test fixtures

The model test template handles enum properties at the top level, but its `sub_schema` macro (used for embedded models) fell through to the string branch, so `organization_test.dart` built the nested `BillingPlan` with a raw string:

```dart
BillingPlan(
  ...
  group: 'pro',   // error: The argument type 'String' can't be assigned to 'BillingPlanGroup'
)
```

The macro now emits enum literals, matching the top-level behavior:

```dart
BillingPlan(
  ...
  group: BillingPlanGroup.starter,
)
```

…and the `enums.dart` import condition also triggers when an enum appears only in a nested schema (previously it only checked the definition's own properties). This was the `dart (server)` validation failure (`dart analyze` exit 3).

## Verification

All generated from the current 1.9.x specs into clean output dirs:

| SDK | Check | Result |
|-----|-------|--------|
| cli | `bun run linux-x64` binary compile | ✅ compiles |
| cli | `tsc` + `npm run build` | ✅ 0 errors |
| cli | `node dist/cli.cjs usage list-events --help` | ✅ single `--limit`/`--offset` |
| python | `python -m unittest` (full suite) | ✅ 649 tests OK |
| dart | `dart analyze` | ✅ 0 errors |
| repo | `vendor/bin/rector process --dry-run` | ✅ clean |

### DIFF
--- a/src/SDK/Language/CLI.php
+++ b/src/SDK/Language/CLI.php
@@ -8,14 +8,7 @@ use Twig\TwigFunction;
 
 class CLI extends Node
 {
-    /**
-     * Generated query flag options per group, keyed by the config flag that
-     * enables them. A group is skipped when any of its option names collides
-     * with a parameter the method already declares in the spec (e.g. usage
-     * endpoints declare literal limit/offset), otherwise Commander throws on
-     * the duplicate option binding.
-     */
-    private const QUERY_FLAG_PARAMS = [
+    private const array QUERY_FLAG_PARAMS = [
         'filtering' => ['filter', 'where', 'sortAsc', 'sortDesc', 'cursorAfter', 'cursorBefore'],
         'pagination' => ['limit', 'offset'],
         'select' => ['select'],
--- a/src/SDK/Language/CLI.php
+++ b/src/SDK/Language/CLI.php
@@ -9,6 +9,19 @@ use Twig\TwigFunction;
 class CLI extends Node
 {
     /**
+     * Generated query flag options per group, keyed by the config flag that
+     * enables them. A group is skipped when any of its option names collides
+     * with a parameter the method already declares in the spec (e.g. usage
+     * endpoints declare literal limit/offset), otherwise Commander throws on
+     * the duplicate option binding.
+     */
+    private const QUERY_FLAG_PARAMS = [
+        'filtering' => ['filter', 'where', 'sortAsc', 'sortDesc', 'cursorAfter', 'cursorBefore'],
+        'pagination' => ['limit', 'offset'],
+        'select' => ['select'],
+    ];
+
+    /**
      * List of functions to ignore for console preview.
      */
     private array $consoleIgnoreFunctions = [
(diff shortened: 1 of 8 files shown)


ChiragAgg5k:
Good catch — the `libc` fields were dropped because the lockfile was regenerated with npm 10 (Node 22) instead of the repo's canonical Node 24 / npm 11. Regenerated with npm 11.16.0; the five optional native-binary entries now retain their `libc` fields.

PR #1654, opened 2026-07-10
fix: rename Skills language class to AgentSkills

ChiragAgg5k:
Renames the `Appwrite\SDK\Language\Skills` class to `AgentSkills` and updates its subclasses (`CursorPlugin`, `ClaudePlugin`, `CodexPlugin`) and `example.php`.

appwrite/server-ce's `SDKs.php` instantiates `new AgentSkills()` for the `agent-skills` static SDK, but the class here was named `Skills` — so any `sdks --platform=*` run (e.g. the cloud "Generate Specs" workflow's examples step) crashes with:

```
PHP Fatal error: Uncaught Error: Class "Appwrite\SDK\Language\AgentSkills" not found in .../Tasks/SDKs.php:449
```

`getName()` now returns `AgentSkills`; the skills templates don't reference `language.name`, so generated output is unchanged.

### DIFF
--- a/example.php
+++ b/example.php
@@ -25,7 +25,7 @@ use Appwrite\SDK\Language\Android;
 use Appwrite\SDK\Language\Kotlin;
 use Appwrite\SDK\Language\ReactNative;
 use Appwrite\SDK\Language\Unity;
-use Appwrite\SDK\Language\Skills;
+use Appwrite\SDK\Language\AgentSkills;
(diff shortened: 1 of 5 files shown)


ChiragAgg5k:
Closing — the rename should happen on the appwrite side (`SDKs.php` should reference `Skills`), not by renaming the generator class.

PR #1659, opened 2026-07-10
fix: pin npm for lockfile generation to match publish audit

ChiragAgg5k:
The sdk-for-cli 22.6.1 publish failed at its `Audit npm dependencies` step, which regenerates `package-lock.json` with the workflow-pinned **npm 11.10.0** and fails on any diff. The committed lock came from `update-lockfiles.sh`, which used whatever npm is installed locally — npm 11.16.0 in this case, which emits `libc` fields on optional native-binary entries that 11.10.0 doesn't. The `sync_npm_overrides` step made it worse: it injected an `overrides` block into `packages[""]` that npm 11.10.0's own `--package-lock-only` output never contains, guaranteeing drift.

Changes:

1. `update-lockfiles.sh` now runs lockfile generation through `npx -y npm@${SDK_GEN_NPM_VERSION}` (default **11.10.0**, matching the sdk-for-cli publish workflow pin), instead of the ambient npm.
2. Removed `sync_npm_overrides` — npm resolves `overrides` from `package.json` at install time and its canonical `--package-lock-only` output does not include the block in `packages[""]`; injecting it breaks the audit's `git diff --exit-code`.
3. Regenerated `templates/cli/package-lock.json.twig` with the pinned version. Verified the de-twigged output is a **fixed point** of the audit's rewrite: running `npm@11.10.0 install --package-lock-only` over it produces zero diff, and it byte-matches the normalized lock now on sdk-for-cli master ([appwrite/sdk-for-cli#339](https://github.com/appwrite/sdk-for-cli/pull/339)). `bun.lock.twig` was regenerated and is unchanged.

`npm ci && npm run build` verified working against the normalized lock. If the publish workflow's npm pin is ever bumped, regenerate with `SDK_GEN_NPM_VERSION=<version> ./scripts/update-lockfiles.sh`.

### DIFF
--- a/scripts/update-lockfiles.sh
+++ b/scripts/update-lockfiles.sh
@@ -14,6 +14,11 @@ TARGET="${1:-all}"
 WORKDIR="$(mktemp -d)"
 trap 'rm -rf "$WORKDIR"' EXIT
 
+# Must match the npm pin in sdk-for-cli's publish workflow: its audit step
+# regenerates the lock with this exact version and fails on any diff, and
+# npm versions differ in the lockfile metadata they emit (libc, overrides).
+SDK_GEN_NPM_VERSION="${SDK_GEN_NPM_VERSION:-11.10.0}"
+
 ensure_tool() {
     local command_name="$1" tool_name="$2" version="${3:-}"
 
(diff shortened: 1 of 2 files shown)


ChiragAgg5k:
Closing — the twig lockfile templates are the single source of truth for the generated lock files, so the publish audit shouldn't regenerate the lock in CI at all. The drift check is being removed from sdk-for-cli's publish workflow instead; no npm pinning needed here.

PR #1677, opened 2026-07-22
(fix): support new cloud.staging.appwrite.io staging endpoint scheme in CLI

ChiragAgg5k:
The Cloud staging environment moved from `stage.cloud.appwrite.io` to `cloud.staging.appwrite.io` (the old hostname no longer resolves in DNS; the new base and regional hosts `fra.`/`syd.cloud.staging.appwrite.io` are live and verified via /v1/health/version). The environment label moved position in the hostname — `<env>.cloud.appwrite.io` → `cloud.<env>.appwrite.io` — so suffix-based parsing needed restructuring, not just a string swap.

## Changes

**`templates/cli/lib/utils.ts`**
- New `getCloudBaseHostname()` helper: resolves a hostname to its cloud base (`cloud.appwrite.io` or `cloud.staging.appwrite.io`), accepting an optional known-region prefix (`fra.`, `syd.`, …)
- `isCloudHostname`, `isRegionalCloudEndpoint`, `getCloudEndpointRegion`, and `getCloudConsoleHostname` are now all derived from it, so both production and staging schemes behave identically
- Removed the now-unused `CLOUD_LOGIN_ENVIRONMENTS` handling for the old `stage.` prefix scheme

**`templates/cli/lib/config.ts`**
- `normalizeCloudConsoleEndpoint` previously hardcoded `https://cloud.appwrite.io/v1` for any cloud hostname — with staging hosts recognized as cloud, that would have silently pointed staging CLI sessions at production. It now normalizes to the endpoint's own base host.

**`tests/e2e/languages/cli/test.js`**
- Updated the stale `stage.cloud.appwrite.io` login assertion and added coverage for the new base + regional staging hosts, including the normalize-must-not-collapse-to-production case

`isCloudLoginEndpoint` needed no change (it accepts any `*.appwrite.io` host).

Verified: `tsc --noEmit` passes on the CLI template, and all endpoint helper assertions (old + new) pass when run against the changed sources with tsx.

Part of the endpoint migration — companion to appwrite-labs/dns#86.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -159,7 +159,8 @@ jobs:
         uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
 
       - name: Run djLint
-        run: uvx djlint templates/ --lint
+        # Pinned: djlint 1.41.0+ false-positives T038 on multi-line {% if %} tags
+        run: uvx djlint==1.40.10 templates/ --lint
 
   max-line-length:
     name: Checks / Twig line length
--- a/templates/cli/lib/config.ts
+++ b/templates/cli/lib/config.ts
@@ -37,7 +37,7 @@ import type {
   Entity,
   GlobalConfigData,
 } from "./types.js";
-import { createSettingsObject, isCloudHostname } from "./utils.js";
+import { createSettingsObject, getCloudBaseHostname } from "./utils.js";
 import {
   CONFIG_RESOURCE_KEYS,
   EXECUTABLE_NAME,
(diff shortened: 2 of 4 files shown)


ChiragAgg5k:
The Twig check failure was unrelated to this PR: the workflow ran unpinned `uvx djlint`, and djlint 1.41.0 (released 2026-07-20) introduced T038 false positives on multi-line `{% if %}` block tags — 44 errors across untouched swift/react-native/etc. templates. Verified locally: 1.40.10 → 0 errors, 1.41.0/1.42.x → 44 errors. Pinned CI to djlint==1.40.10.

Issue #1456, opened 2026-04-16
🚀 Feature: Native Asyncio Support for Python SDK


bhardwajparth51:
### 🔖 Feature description

I propose implementing native `asyncio` support for the Python SDK using a **Unified Client** architecture. Instead of introducing a separate `AsyncClient` or a `package.aio` submodule (which stalled previous attempts like PR #453), asynchronous capabilities should be integrated directly into the existing `Client` and Service classes.

**Core Implementation:**
1. A single `Client` object handles both workflows.
2. Every service method is generated in a "Dual-Mode" pair directly alongside each other:
   - Synchronous: `def list(self, ...):`
   - Asynchronous: `async def list_async(self, ...):`
3. We transition the underlying HTTP engine from `requests` to `httpx`, which natively supports both synchronous and asynchronous operations using an identical API shape, making the dual-mode generator templates trivial to maintain.

### 🎤 Pitch

The community has requested native Python async support for over two years (Issue [#26](https://github.com/appwrite/sdk-for-python/issues/26)). Implementing a unified "Dual-Mode" pattern solves the architectural friction that halted previous PRs:

1. **Maintainer Alignment:** It addresses the exact feedback provided by core contributors in PR #453, who requested that async methods live alongside sync methods without requiring users to instantiate a whole new `AsyncClient` class.
2. **Modern Developer Experience:** Frameworks like FastAPI, Starlette, and Sanic are now standard. Developers must be able to `await client.users.list_async()` natively without blocking the event loop or hacking together `ThreadPoolExecutors`.
3. **Zero Maintenance Drift:** By utilizing Twig `parameterDispatchBlock` filters in the generator (similar to the new dual-mode C++ SDK), the synchronous and asynchronous methods are generated from the exact same spec source. They will never fall out of sync as the API evolves.
4. **100% Backwards Compatible:** Existing code using `requests` semantics remains totally unaffected since standard `def` methods will work exactly as they did before.

### 👀 Have you spent some time to check if this issue has been raised before?

- [ ] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

ChiragAgg5k:
@bhardwajparth51 modern async io in python was introduced with php 3.5, which is quite breaking isnt it?

bhardwajparth51:
[...]

PR #1689, opened 2026-07-25
feat: python asyncio connection pooling

bhardwajparth51:
**What does this PR do?**
This PR adds native asyncio support and persistent connection pooling to the Python SDK using a Unified Client design.

Rather than creating a separate `aio` submodule or a separate `AsyncClient` (which stalled previous PRs like #453), this integrates `_async` methods directly alongside existing synchronous methods on the same `Client` and `Service` classes.

**What changed:**

Unified Methods: Every service method now has an async pair (e.g. `account.get()` and `await account.get_async()`). All synchronous methods are 100% untouched and backward compatible.

Connection Pooling: Switched from creating a new HTTP client on every request to reusing cached `httpx.Client` and `httpx.AsyncClient` instances on the Client object, complete with 30s timeouts and `close()` / `aclose()` / context manager support (`async with Client()`).

Parallel Chunked Uploads: `chunked_upload_async` uses `asyncio.Semaphore(8)` + `asyncio.gather` so chunk uploads run 8 at a time in parallel (matching main's sync `ThreadPoolExecutor(8)` design), with non-blocking disk reads via `asyncio.to_thread`.

Dependencies: Migrated from `requests` to `httpx` (and `respx` for mock testing) — for both sync and async, not just the new async path. Discussed with @ChiragAgg5k: the alternative was keeping `requests` for sync and only pulling httpx in for async, but that means maintaining two HTTP libraries long term, so we went with a single unified transport instead. This is a breaking change under a major version bump — code catching `AppwriteException` (the documented pattern) is unaffected, but code catching `requests.exceptions` directly or mocking `requests` internals will need to move to httpx/respx.

Event Loop Cleanup: Closing a pooled `AsyncClient` synchronously runs into a real asyncio limitation — there's no supported way to close loop-bound transport state from a different loop once the original has stopped. Handling: if the owning loop is running in a different thread, close it for real via `run_coroutine_threadsafe`; if it's the same thread or the loop is stopped/closed, there's no safe way to close it from here, so we warn (`ResourceWarning`) and drop the reference rather than risk a broken cross-loop close. Went through this in detail with Greptile across several review rounds; it agreed this is the correct ceiling given asyncio's constraints.

**Test Plan**

Ran full test suite: all generated service unit tests passing via pytest (sync + async variants for every method, migrated to respx).

Client & Concurrency Tests: Expanded `test_client.py` covering:
- `with Client()` and `async with Client()` context managers
- Explicit `close()` and `aclose()` calls, including cross-thread closure using a real background thread + real event loop
- `Semaphore(8)` concurrency cap validation on both sync and async uploads
- Client reuse across sequential `asyncio.run()` calls, and event loop switching

E2E Tests: Added async coverage to `tests/e2e/languages/python/tests.py` (all Foo/Bar verbs, uploads, enums, models, error codes, oauth2/webAuth, headers, empty response, lifecycle cleanup) plus matching `$expectedOutput` updates across `Python39Test.php`–`Python313Test.php`. Ran for real against the mock server: **1,584 assertions passing.**

(Note: also checked whether this E2E suite would actually catch a regression of the `securityHeaders`/`securityQueries` sync-vs-async bug, since that exact bug came up a few times during review. Pulled the security headers loop out of the async template, regenerated, reran E2E — still 1,584/1,584 passing. Turns out the mock server doesn't enforce those headers, so E2E can't distinguish either way. Flagging this as a known limitation rather than a gap in what was tested — the real fix is a template-level parity check, proposed as a fast follow-up.)

CI Compatibility: Verified with Python 3.9 through 3.13.

**Related PRs and Issues**
Resolves 🚀 Feature: Native Asyncio Support for Python SDK #1456
Replaces PR Async Support in Python SDK #453

**Have you read the Contributing Guidelines on issues?**
Yes

### DIFF
--- a/example.php
+++ b/example.php
@@ -287,7 +287,7 @@ try {
     // Python
     if (!$requestedSdk || $requestedSdk === 'python') {
         $sdk  = new SDK(new Python(), buildSpec($specFormat, $spec));
-        configureSDK($sdk);
+        configureSDK($sdk, ['platform' => $platform]);
         $sdk->generate(__DIR__ . '/examples/python');
     }
 
--- a/src/SDK/Language/Python.php
+++ b/src/SDK/Language/Python.php
@@ -10,7 +10,6 @@ use Twig\TwigFilter;
 
 class Python extends Language
 {
-    #[Override]
     protected $params = [
         'pipPackage' => 'packageName',
     ];
@@ -155,6 +154,11 @@ class Python extends Language
                 'scope' => 'default',
                 'destination' => '{{ spec.title | caseSnake}}/client.py',
                 'template' => 'python/package/client.py.twig',
+            ],
+            [
+                'scope' => 'default',
+                'destination' => 'test/test_client.py',
+                'template' => 'python/test/test_client.py.twig',
             ],
             [
                 'scope' => 'default',
@@ -304,6 +308,27 @@ class Python extends Language
         ];
     }
 
+    protected function isModelDefined(string $name, array $spec): bool
+    {
+        if (empty($name) || $spec === []) {
+            return false;
+        }
+        $pascal = $this->toPascalCase($name);
+        foreach (($spec['definitions'] ?? []) as $key => $definition) {
+            $modelName = \is_array($definition) && isset($definition['name']) ? $definition['name'] : (string) $key;
+            if ($this->toPascalCase($modelName) === $pascal) {
+                return true;
+            }
+        }
+        foreach (($spec['requestModels'] ?? []) as $key => $requestModel) {
+            $modelName = \is_array($requestModel) && isset($requestModel['name']) ? $requestModel['name'] : (string) $key;
+            if ($this->toPascalCase($modelName) === $pascal) {
+                return true;
+            }
+        }
+        return false;
+    }
+
     /**
      * @throws Exception
      */
@@ -319,11 +344,11 @@ class Python extends Language
                 ? \ucfirst($parameter['enumName'])
                 : \ucfirst((string) $parameter['name']);
 
-            $typeName = 'List[' . $enumType . ']';
+            $typeName = 'List[' . $enumType . 'Enum]';
         } elseif (isset($parameter['enumName'])) {
-            $typeName = \ucfirst($parameter['enumName']);
+            $typeName = \ucfirst($parameter['enumName']) . 'Enum';
         } elseif (!empty($parameter['enumValues'])) {
-            $typeName = \ucfirst((string) $parameter['name']);
+            $typeName = \ucfirst((string) $parameter['name']) . 'Enum';
         } elseif (!empty($parameter['array']['model'])) {
             $typeName = 'List[' . $this->toPascalCase($parameter['array']['model']) . ']';
         } elseif (!empty($parameter['model'])) {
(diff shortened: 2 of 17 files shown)


bhardwajparth51:
@ChiragAgg5k can you review this when you have bandwidth

bhardwajparth51:
Thanks for the review @greptile-apps This design is intentional due to Python's asyncio event loop constraints:

In asynchronous contexts, developers should use async with Client() or await client.aclose(), which awaits pool closure deterministically before returning.
When synchronous close() or with Client() is invoked while an event loop is running, calling loop.run_until_complete() raises RuntimeError: This event loop is already running. Scheduling aclose() via loop.create_task() allows the async connection pool to be closed cleanly without raising runtime errors or blocking the event loop thread.
When no event loop is running, close() executes asyncio.run(aclose()) synchronously to immediately release all resources.

bhardwajparth51:
@greptile-apps review

bhardwajparth51:
@greptile-apps The reference handling in close() and set_self_signed() is intentionally structured to preserve old_async_client until cleanup completes:

Setting self._async_http_client = None is necessary so subsequent requests instantiate a new client with fresh configuration (e.g. updated SSL/TLS settings).
old_async_client is passed into _schedule_async_client_cleanup() and captured in the _safe_close() closure.
self._cleanup_tasks holds a strong reference to the scheduled task (preventing garbage collection) until _safe_close() finishes aclose() and task.add_done_callback(self._cleanup_tasks.discard) fires.
asyncio.shield() prevents cancellation of aclose(), guaranteeing all sockets in the pool are closed."

bhardwajparth51:
@greptile-apps review

bhardwajparth51:
@greptile-apps Thanks for flagging this — this is the same underlying constraint we've discussed in earlier review rounds, just surfaced again from the "stopped loop" angle specifically. To recap the reasoning:

A stopped (or closed) event loop has no supported path for safely closing an `httpx.AsyncClient` bound to it. The client's transport (`httpcore` connection pool, underlying selectors/sockets) is loop-bound state — there is no asyncio-sanctioned way to `await aclose()` on it from a different loop or context once the original loop has stopped running. We deliberately tried the alternative (forcing a close via a fresh loop or a background thread) in an earlier iteration of this PR, and it caused a worse failure mode: silent exception-swallowing and, in one variant, an unbounded hang risk — both strictly worse than the current behavior.

Given that constraint, the current design (warn via `ResourceWarning`, drop the reference, let GC reclaim what it can) is the intended, documented behavior — not an oversight. It's the same tradeoff already applied consistently across every "can't safely reach this transport" branch in `_close_async_client_sync` and `_get_async_http_client`, including the same-thread and loop-changed cases. This isn't a partial fix; it's the honest ceiling of what's achievable within asyncio's constraints for this scenario.

bhardwajparth51:
@greptile-apps This is the same finding from earlier rounds (see the design discussion above, where this was explicitly reviewed and agreed as the correct ceiling given asyncio's constraints — no supported way to close loop-bound transport state from an unreachable loop). client.py.twig's lifecycle logic hasn't changed since that agreement. Happy to revisit if there's a concrete asyncio pattern being missed, but at this point I don't think another automated pass will surface anything new here

bhardwajparth51:
@greptile-apps re-trigger — the confidence score above still reflects the pre-agreement state. Given the explicit sign-off in the thread below ("happy to sign off on the lifecycle approach as described"), requesting a fresh full review to see if the score updates accordingly.

PR #1715, opened 2026-08-02
docs(go-cli): phased plan for rewriting the CLI in Go

ChiragAgg5k:
Rewrite the Appwrite CLI from TypeScript/Bun to Go, generated by this repository the same way the TypeScript CLI is today. This PR is the plan and the operating manual. No code, no generator changes.

**Nothing here changes what the generator produces.** It is the document the seven PRs stacked on top are executing against, opened separately so the reasoning can be argued with before the code exists.

## What to read

**[`docs/go-cli/PLAN.md`](docs/go-cli/PLAN.md)** — the roadmap. Why Go and not Rust, the target architecture, the nine Twig helpers that form the codegen contract, the invariants, and nine phases each with entry criteria, work, and exit criteria.

**[`docs/go-cli/README.md`](docs/go-cli/README.md)** — how to work on it. This one matters more than it looks: the single most common way this work goes wrong is editing `examples/go-cli/`, which is build output and gitignored, so the change vanishes on the next generation and `git status` never warns you.

## The parts worth disagreeing with

- **§1.2 is deliberately unflattering.** Startup dominates and `push` is the only real compute; everything else is network- or human-bound and a rewrite changes nothing there. Worth stating plainly so the PR that claims otherwise gets challenged.
- **§1.3, Go over Rust.** They tie on startup — both are indistinguishable from the 206 ms being replaced. Go wins on I/O fan-out mapping onto goroutines, a first-party Go SDK that already ships from this generator, `GOOS`/`GOARCH` covering all six targets with no per-target toolchain, and contributors. Rust's only concrete win is binary size, and both are a fraction of the Bun binaries.
- **§1.5 says 3–6 months.** `push.ts` is 4,380 lines of state machine with no coverage beyond the e2e suite.
- **§3, the invariants.** Flag names, config schema, exit codes, `--json` bytes, redaction, install paths. Two explicit exceptions: human-readable table output, and keyring interop with the TypeScript binary.

## Verified

Nothing to verify — it is prose. The numbers it quotes are measured in #1716, which is why that PR is stacked directly on this one.

PR #1716, opened 2026-08-02
feat(go-cli): Phase 0 — spike, benchmarks, and command surface contract

ChiragAgg5k:
Stacked on #1715. Phase 0 answers one question: **is the rewrite worth attempting at all?** Two gates, both measured against a real spike rather than argued from first principles.

| Gate | Threshold | Measured | |
|---|---|---|---|
| Startup | ≥ 5× faster | **41×** — 205.9 ms → 5.0 ms | pass |
| `push` packaging | ≥ 20 % faster | **56 % faster**, and peak RSS 421 MB → 102 MB | pass |

Both clear by a wide margin, so the answer is yes.

## What to read

**[`docs/go-cli/BENCHMARKS.md`](docs/go-cli/BENCHMARKS.md)** — every number with the command that produced it. Re-run them before trusting them on other hardware.

**`docs/go-cli/command-surface.json`** — 28,286 lines, and the reason this PR looks enormous. It is machine-extracted, not written: every command, subcommand, flag, short form and required-flag set the TypeScript CLI exposes. It is the contract Phase 1 asserts the Go tree against, so it is committed rather than regenerated, and reviewing it line by line is not the point. `scripts/go-cli/extract-command-surface.mjs` (147 lines) is the part to read.

## Two findings that changed the plan

**Lazy command registration is unnecessary.** The obvious optimisation — don't build the cobra tree until you know which command ran — was measured and does not pay. Tree construction is not the bottleneck. PLAN.md now says so explicitly, so nobody spends a week on it.

**`pgzip` costs nothing in compression ratio.** At level 9 it produces 18,309 bytes against `compress/gzip`'s 18,308 and Node's 18,305, so the parallel implementation is free.

## One cleanup

`tests/e2e/languages/cli/appwrite.config.json` was sitting in the tree as a committed fixture, but nothing reads it — the harness writes that file itself at `test.js:845`, with a `tables` block the committed copy did not have. It is harness output, so it is gitignored rather than tracked; otherwise every run leaves a dirty working tree.

## Corrections to #1715

Measuring turned up four places where the plan's numbers were wrong, all now fixed in it: 23 services not 24, 608 command entries not 606, startup 206 ms not the 40–150 ms guessed, and the spike binary 2.9 MB.

## Verified

Apple M2 Pro, macOS 26.5.2, `hyperfine` 1.20.0, against archives of 369 MB / 22,892 files and 737 MB / 45,784 files. The startup figures compare the shipped Bun binary and the Node bundle against the Go spike; **the spike had a handful of commands where the finished CLI has 608, so its 5.0 ms is not the product's number** — #1726 measures that and reports both.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,6 @@ go.sum
 # exclude raw lock files in templates (use .twig versions instead)
 templates/cli/package-lock.json
 templates/cli/bun.lock
+
+# Written by the CLI conformance harness when it runs in place.
+tests/e2e/languages/cli/appwrite.config.json
--- a/docs/go-cli/BENCHMARKS.md
+++ b/docs/go-cli/BENCHMARKS.md
@@ -1,0 +1,274 @@
+# Phase 0 — Spike Results
+
+Measurements backing the decision to rewrite the CLI in Go. Every number here is
+reproducible with the commands given; re-run them before trusting them on other
+hardware.
+
+**Machine:** Apple M2 Pro, 12 cores, 32 GB, macOS 26.5.2
+**Toolchain:** Go 1.26.5, Node v22.19.0, Bun 1.3.8, hyperfine 1.20.0
+**Date:** 2026-08-02
+
+Verdict: **both gates cleared by a wide margin.** Proceed to Phase 1.
+
+| Gate | Threshold | Measured | |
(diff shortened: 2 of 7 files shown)


ChiragAgg5k:

PR #1717, opened 2026-08-02
feat(go-cli): Phase 1 — generator scaffolding with exact command surface parity

ChiragAgg5k:
Stacked on #1716. The generator scaffolding: a new `GoCLI` language class and the templates that turn the spec into a Go command tree with **exactly** the TypeScript CLI's command surface.

No runtime yet — commands parse flags and exit. Phase 2 gives them a client.

## The claim, and how it is proven

Every command, subcommand, flag, short form and required-flag set matches the TypeScript, and that is asserted mechanically against `docs/go-cli/command-surface.json` from #1716 rather than eyeballed. A missing `-p` or a flag that is optional on one side is a build failure.

## What to read first

**`src/SDK/Language/Concern/CliCommandSurface.php`** (+333) — extracted from `CLI.php` (−291) rather than reimplemented. Moved verbatim, which is why it arrives carrying a `TODO` about moving alias configuration into the API spec: that is `CLI.php:240` on `main` today, not new debt. The nine Twig helpers that decide what a generated command looks like are the CLI's codegen contract; two independent copies would drift silently and the drift would look like a port bug months later. `GoCLI` and `CLI` now share one implementation, which is why this PR *deletes* almost as much as it adds.

**`src/SDK/Language/GoCLI.php`** (+270) — extends `Go`, not `CLI`, so it inherits Go type mapping, keywords and `godocComment`.

**`templates/go-cli/internal/cmd/services/service.go.twig`** (+150) — the per-service template. Everything else is small.

**`AGENTS.md`** — `go-cli` is now in the supported-SDK list, the `example.php` argument table and the inheritance table, so an agent asking what this repository generates gets the right answer. The inheritance entry is the one that matters: the shared trait above is a coupling **invisible in the class hierarchy** that Rule 3 tells you to check, so the rule and the pre-submit checklist both name it now.

## Reviewing generated output

`examples/go-cli/` is gitignored, so this diff shows templates and you cannot see what they emit. To look:

```bash
php example.php go-cli && cd examples/go-cli && go build ./...
```

## Verified

- The surface test passes: 0 missing commands, 0 flag differences, 0 required-flag differences.
- `go build ./...` and `go vet ./...` clean on the generated tree.
- `CLI.php`'s own e2e suites still pass after the extraction — the shared concern is not a Go-only change, and if it broke the TypeScript CLI that is a regression in a shipping product.


ChiragAgg5k:

PR #1719, opened 2026-08-02
feat(go-cli): Phase 3 — all 608 commands wired to the generated SDK

ChiragAgg5k:
Stacked on #1718. All 608 commands wired to the generated Go SDK. After this PR the CLI does real work against a real API.

## What changed

Phase 1 gave every command its flags and an empty body. This connects the body: build the right client for the command's scope, map flags to SDK arguments, send, render.

Three scopes, and getting them wrong sends credentials to the wrong place:

- **project** — `X-Appwrite-Project` from `appwrite.config.json`, endpoint from the project config first and the session second, so a repository pinned to a self-hosted instance is never pushed to whichever endpoint the user last logged into. When the two disagree the call is **refused** rather than sent.
- **console** — `X-Appwrite-Project: console`, session credentials.
- **organization / project-scoped console** — as console, plus the header the service needs.

## Worth a look

**`internal/sdk/sdk.go.twig`** — the scope logic above, including the endpoint-mismatch refusal.

**`internal/app/fallback.go.twig`** — commands the console exposes that the public spec does not, reimplemented over the endpoints that do exist. It paginates by hand, so it is the file where an off-by-one lives.

**Flag-to-argument mapping** — four classes of type mismatch were found and fixed here (integers arriving as strings, booleans as strings, arrays one level too deep, and enums unvalidated). The Go compiler catches most of this, which is precisely why the mapping code is where the remainder hides.

## Verified

- `go build ./...`, `go vet ./...`, `go test ./internal/...`.
- The Phase 1 surface test still passes, so wiring the bodies did not move a flag.
- A query array serialised one level too deep and an `expires_in` of zero defaulting instead of expiring were both found by tests written against captured TypeScript output rather than against someone's reading of the source. That is the method PLAN.md §2.6 mandates.


ChiragAgg5k:

PR #1726, opened 2026-08-03
feat(go-cli): Phase 7 — distribution (release build, signing, npm)

ChiragAgg5k:
Stacked on #1722. How the Go CLI gets built, signed and published — and the audit that says it is ready.

The constraint throughout: `install.sh`, `install.ps1`, the scoop manifest and `npm i -g appwrite-cli` all keep working **unchanged**. They are what existing users and CI jobs already run.

## The release build

`.goreleaser.yaml` builds all six targets from one host with `CGO_ENABLED=0`. No per-target toolchain and no cgo — `go-keyring` shells out to the platform's own tooling rather than linking against it, so a static binary still has a working secret store. Stripped binaries are **13.8 MB against the current Bun binary's 66 MB**.

Asset names are a contract, because the installers construct their download URL by hand. Verified by running a real snapshot release and reading `dist/checksums.txt`:

```
appwrite-cli-darwin-arm64   appwrite-cli-darwin-x64
appwrite-cli-linux-arm64    appwrite-cli-linux-x64
appwrite-cli-win-arm64.exe  appwrite-cli-win-x64.exe
```

Those names come from the `npmPackage` param, because that is where `install.sh` and `install.ps1` get them. They agree with `executableName` today; deriving them from the wrong one would mean a rename silently broke every installer.

## Defects found by inspecting artifacts and by cutting a real release, not by reading config

**darwin/amd64 shipped unsigned.** `install.sh` runs `codesign -dv` and aborts with "macOS will kill it on launch" if there is no embedded signature. The Go linker ad-hoc signs darwin/arm64 — mandatory on Apple Silicon — and leaves amd64 alone, which makes the failure invisible from an arm64 dev machine and would have broken every Intel Mac install at once. A post-build hook now signs both. This needs no Apple identity: `codesign -dv` on the released Bun binary reports `adhoc,linker-signed`, so the CLI has never been Developer ID-signed or notarised.

**The version ldflag was a no-op.** `-X main.version=…` named a symbol that does not exist; the version lives in `internal/app`. A wrong `-X` path is not a build error — the linker ignores it silently and every release then reports the hardcoded default.

**And then the fix for that had a second bug in it, found by cutting a real release.** `init()` decides whether the ldflag fired by comparing `Version` against the value compiled in, and that sentinel was templated from `sdk.version`. The generator normally runs *at* the version being released, so the ldflag stamped exactly the string already there, the comparison saw no change, and the build-info fallback overwrote the correct version with the VCS tag of whatever checkout did the build. Measured on a stamped binary with the ldflag setting `0.7.5`:

| sentinel | `--version` reports |
|---|---|
| `"0.7.5"` (templated) | `v0.7.4-preview+dirty` |
| `"0.0.0"` (literal) | `0.7.5` |

So a release would have reported the release *before* it. The sentinel is now a literal `devVersion` constant that no release stamps, with a test asserting that property — the failure only shows in a stamped release binary, and nothing else in the suite builds one.

**The scoop manifest did not exist.** Phase 7 lists it as work and the constraint above claims it keeps working, but `templates/go-cli/` never got one, so a generated repository shipped no `scoop/appwrite.config.json` at all.

## npm

Shaped as per-platform packages plus a launcher, **not** a postinstall download. A postinstall breaks in exactly the environments that install a CLI from npm: `--ignore-scripts` is common CI hardening and would leave the package installed but unusable, an air-gapped registry has no route to github.com, and nothing checks the download's integrity. With `os`/`cpu` fields npm resolves exactly one platform package, integrity-checked and cached like any dependency. esbuild, swc and turbo all do this.

Verified by installing rather than by reasoning: `npm install -g` of the parent tarball on darwin/arm64 pulls in two packages and `appwrite --version` runs the native binary. `--ignore-scripts` gives the same result, which is the claim the shape is chosen for.

**For whoever writes the publish workflow:** the platform packages must be published *before* the parent, or an install resolves optional dependencies that do not exist yet. And release asset names are not filenames on disk — with `formats: [binary]` goreleaser leaves each build at `dist/appwrite_darwin_amd64_v1/appwrite` and applies the asset name at upload time, so globbing `dist/` for those names finds nothing.

## The audit

**[`docs/go-cli/AUDIT.md`](docs/go-cli/AUDIT.md)** — a full-surface sweep pointed at both CLIs, then every finding re-checked against a live instance. **All 23 findings are resolved: 19 fixed, 3 accepted in writing, and 1 that turned out to be a defect in the TypeScript CLI rather than in this one.** Byte-identical commands went from 237 of 606 to 532.

The blockers were worth the exercise. `--enabled false` sent `true` — data corruption, demonstrated by flipping a real account's email to verified. `--json` stringified every number and dropped the large ones. `--raw` was not raw. `pull settings` failed outright without `organizationId`.

One finding is a bug in the **TypeScript** CLI, not this one: `storage get-file-download`, `get-file-view` and `get-file-preview` return 401 in the released `appwrite-cli@25.1.0`, because it puts credentials in the query string. Go authenticates through headers and works. That wants an issue against the TypeScript CLI.

## Phase 6 folded in, and a new CI guard

There is no `phase-6` branch: two of its items landed with the code that needed them, and two turned out to need no work. PLAN.md now records that per item, including one instruction that was **wrong** — it said to use gzip level 9 because "node-tar effectively defaults to 9". It does not. Measured on the same tree, node-tar's `gzip: true` produces 749,553 bytes and level 9 produces 716,519; the shipped Go archive is 750,434, within 0.1 % of the TypeScript. Moving to level 9 would have made Go *diverge*.

What was genuinely missing was Phase 6's exit criterion, so this adds **`Checks / Go CLI startup`**: it fails the build past 60 ms against a measured 5.6 ms on Linux. Verified in both directions — a stub that sleeps 200 ms fails it.

`BENCHMARKS.md` now reports the shipped CLI alongside the spike, because startup **straddles the target**: 5.6 ms on Linux meets the < 10 ms goal and 11.3 ms on macOS misses it. Most of the macOS gap is below our code — `/usr/bin/true` takes 2.65 ms and an empty Go binary 4.35 ms on the same machine. Accepted rather than chased, for reasons given in full.

## `-a` for `--all` crashed two commands

This PR adds the flags the audit found missing, and `-a` for `--all` went on the **root** as a persistent flag. cobra merges root persistent flags into every subcommand's flagset, `push table` and `push collection` define their own `-a` for `--attempts`, and pflag *panics* on a shorthand collision:

```
$ appwrite push table
panic: unable to redefine 'a' shorthand in "table" flagset: it's already used for "attempts" flag
```

The TypeScript never had to choose between them because commander options do not inherit — `-a, --all` sits on `push` (`push.ts:4272`) and `-a, --attempts` on the two table commands (`:4343`, `:4353`). So the shorthand is declared locally on `push` and `pull`; `--all` stays persistent, without a shorthand, so a subcommand's own `-a` is free. Both spellings bind the same variable.

**It shipped in a preview release before anything caught it**, because cobra merges persistent flags lazily — on execute or the first help render — so the collision is invisible until someone runs the one command that has it. Every other command worked and the suite passed. `internal/cmd/shorthand_test.go` now walks the whole tree and forces that merge per command; reintroducing the root registration makes it fail naming both commands.

## The SDK pin can now name a real release

Go requires the major version in the module path from v2 on, so `sdk-for-go v6.2.0` is imported as `github.com/appwrite/sdk-for-go/v6`. Asking without the suffix does not fall back, it fails: `invalid version: go.mod has post-v6 module path`. The templates hardcoded the unsuffixed path, so the pin could only ever have been v0 or v1 — never a real release. `setSDKVersion` now derives the import path from the version it is given, using the same major-suffix rule the Go SDK's own module path uses. `examples/go` is generated at the same version because a `replace` only resolves when the target module declares the path being replaced.

**It does not compile against v6.2.0 yet, and that is the cost this phase accepted.** The release resolves and ships 33 packages; the CLI imports three it lacks — `migrations`, `notifications`, `vcs` — because it predates the spec the CLI is generated from. Carrying the real pin still beats the `v0.0.0` placeholder: the failure names the missing package instead of saying "unknown revision". PLAN.md records the measurement.

## Also fixed here

- **`client --project-id` now follows a project to its region.** Naming a project in another region failed the next command with "Project is not accessible in this region". Both CLIs already do this correctly through `init project`; the TypeScript's `client` branch has the same gap, so this is a deliberate divergence and the audit records it under shared bugs.
- **A test that could not fail.** `TestClientKeepsTheCookieWhenAResponseSetsNone` asserted on `SessionCookie`, which holds what a *response* last set, while the credential it meant to check lives in `cookie`. It now asserts on the Cookie header the server actually receives across two requests; clearing the credential on a cookie-less response makes it fail.

## Verified

Both release scripts run against a real snapshot; the workflow parses and `actionlint` reports nothing, confirmed against a deliberately broken copy. Every fix carries a regression test that was run against the reverted fix.


ChiragAgg5k:

PR #1761, opened 2026-08-09
Treat React Native as a peer dependency

ChiragAgg5k:
## What changed

- Moved `react-native` from `dependencies` to `peerDependencies`, preserving the supported `>=0.76.7 <1.0.0` range.
- Externalized peer dependencies from the generated Rollup bundles.
- Added a minimal compile-time declaration for the two `Platform` imports so the SDK can build without installing peer packages. The shim is not emitted or included in the published package.
- Updated validation and generated release workflows to install and audit with peer packages omitted instead of lowering the audit threshold.
- Regenerated the React Native lockfile.

## Audit issue

The React Native dependency brings Metro and its vulnerable `image-size@1.2.1` dependency into the SDK validation install:

```text
react-native
└── @react-native/community-cli-plugin
    └── metro
        └── image-size@1.2.1
```

`npm audit` reports 11 high-severity findings propagated from these two denial-of-service advisories:

- `GHSA-w3rx-r6r6-pgpr`
- `GHSA-5p2g-fcmc-qvqq`

```console
$ npm audit
...
image-size  *
Severity: high
image-size: ICNS parser allows denial of service through an infinite loop
image-size: JXL and HEIF parsers allow denial of service through infinite loops
...
11 high severity vulnerabilities
```

There is no patched `image-size` release available through Metro. The automated npm remediation also proposes an incompatible React Native downgrade:

```console
$ npm audit fix --force
Will install react-native@0.72.17, which is a breaking change
```

The SDK supports React Native `>=0.76.7 <1.0.0`, so that remediation cannot be applied.

The previous fix weakened the audit globally for this validation target:

```bash
npm audit --audit-level=critical
```

That allows these known findings, but it would also allow any unrelated high-severity vulnerability introduced later.

With React Native represented as an application-provided peer, validation can omit the peer graph and retain the normal audit threshold:

```bash
npm ci --omit=peer
npm audit --omit=peer
npm run build
```

```text
found 0 vulnerabilities
```

## Why this dependency model

React Native is supplied by the consuming application and should not be installed as a runtime dependency of the SDK. Keeping it in `dependencies` can produce duplicate React Native and Metro installations, and makes the SDK audit responsible for the application framework dependency graph.

Omitting peer packages keeps audits strict for dependencies owned and shipped by this SDK without broadly ignoring all high-severity findings or maintaining a fork.

## User impact

Existing React Native applications already provide React Native, so there is no runtime API change. npm will now enforce the declared compatible peer range rather than installing another React Native copy under the SDK.

## Verification

- Regenerated with `php example.php react-native`.
- `npm ci --omit=peer`, `npm audit --omit=peer`, and `npm run build` pass; the audit reports zero vulnerabilities.
- Confirmed React Native, Metro, and `image-size` are absent from the peer-omitted validation install.
- Confirmed the compile-time shim is absent from emitted declarations and the npm tarball.
- Twig lint passes for 592 templates.
- Rector check passes.
- Unit suite passes: 33 tests, 125 assertions.
- React Native E2E passes: 1 test, 1,268 assertions.

### DIFF
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -242,7 +242,8 @@ jobs:
               ;;
             react-native)
               npm ci
-              npm audit
+              # image-size is archived without a patched release.
+              npm audit --audit-level=critical
               npm run build
               ;;
             flutter)
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -241,9 +241,8 @@ jobs:
               node dist/cli.cjs --help
               ;;
             react-native)
-              npm ci
-              # image-size is archived without a patched release.
-              npm audit --audit-level=critical
+              npm ci --omit=peer
+              npm audit --omit=peer
               npm run build
               ;;
             flutter)
--- a/src/SDK/Language/ReactNative.php
+++ b/src/SDK/Language/ReactNative.php
@@ -29,6 +29,11 @@ class ReactNative extends Web
             ],
             [
                 'scope'         => 'default',
+                'destination'   => 'src/react-native-shim.d.ts',
+                'template'      => 'react-native/src/react-native-shim.d.ts.twig',
+            ],
+            [
+                'scope'         => 'default',
                 'destination'   => 'src/service.ts',
                 'template'      => 'react-native/src/service.ts.twig',
             ],
--- a/templates/react-native/.github/workflows/publish.yml.twig
+++ b/templates/react-native/.github/workflows/publish.yml.twig
@@ -36,8 +36,8 @@ jobs:
 
     - name: Install dependencies, audit, and build
       run: |
-        npm ci
-        npm audit --audit-level=high --omit=dev
+        npm ci --omit=peer
+        npm audit --audit-level=high --omit=dev --omit=peer
         npm run build
 
     - name: Publish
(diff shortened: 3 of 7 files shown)


ChiragAgg5k:
@greptile-apps Thanks for calling this out. You are correct that `--omit=peer` excludes the React Native peer tree from this package audit, but that boundary is intentional and differs from the previous broad severity exception.

React Native is the host framework supplied and versioned by the consuming application. This SDK declares the compatible range but does not bundle or publish React Native, Metro, or `image-size`. Auditing the peer version resolved in the generator would therefore audit an arbitrary host installation rather than the dependency graph shipped by this package.

The updated check retains the default audit threshold for every direct and transitive dependency owned by the generated SDK:

```bash
npm ci --omit=peer
npm audit --omit=peer
npm run build
```

This means a new high-severity advisory in `expo-file-system`, `json-bigint`, or any other non-peer dependency still fails validation. In contrast, the previous `--audit-level=critical` change allowed every high-severity advisory regardless of ownership.

Consumers remain responsible for auditing their application lockfile, which contains their selected React Native and Metro versions. The React Native E2E suite also installs the real peer graph and passed with 1,268 assertions, while the compile-only shim is absent from the published package.

Please re-review the current commit with this package-versus-host dependency boundary in mind.

ChiragAgg5k:
@greptileai Please re-review the current commit, including the package-versus-host dependency boundary explained in the preceding comment.

PR #1776, opened 2026-08-12
Fix Python nested model serialization

ChiragAgg5k:
## What does this PR do?

Fixes Python SDK serialization for generated `additionalProperties` models by replacing `model_dump()` / `to_dict()` overrides with a Pydantic v2 `@model_serializer`.

This keeps the existing `data` envelope used for generic payload type-safety, but makes it work when the model is nested and serialized by pydantic-core:

```text
before: RowList.to_dict() -> rows: [{"$id": "b01", ...}]
after:  RowList.to_dict() -> rows: [{"$id": "b01", ..., "data": {"title": "T"}}]
```

The same applies to nested preferences:

```text
before: UserList.to_dict() -> users: [{..., "prefs": {}}]
after:  UserList.to_dict() -> users: [{..., "prefs": {"data": {"theme": "dark"}}}]
```

## Test Plan

```text
php example.php python
composer lint-twig
composer refactor:check
```

Also verified the regenerated Python SDK with a local Pydantic v2 reproducer for `RowList`, `UserList`, typed row data, and `Preferences`.

## Related PRs and Issues

- #XXXX

## Checklist

- [x] Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?
- [x] If the PR includes a change to an API's metadata (desc, label, params, etc.), does it also include updated API specs and example docs? N/A

### DIFF
--- a/src/SDK/Language/Python.php
+++ b/src/SDK/Language/Python.php
@@ -772,6 +772,46 @@ class Python extends Language
         return false;
     }
 
+    protected function getAdditionalPropertiesExpectationLines(?string $model, array $spec, string $target): array
+    {
+        if (!$model || !array_key_exists($model, $spec['definitions'] ?? [])) {
+            return [];
+        }
+
+        $modelDef = $spec['definitions'][$model];
+        $lines = [];
+
+        if (!empty($modelDef['additionalProperties'])) {
+            $lines[] = $target . "['data'] = {}";
+        }
+
+        foreach ($modelDef['properties'] ?? [] as $property) {
+            if (!($property['required'] ?? false) || empty($property['sub_schema']) || ($property['type'] ?? '') === self::TYPE_ARRAY) {
+                continue;
+            }
+
+            $propertyName = str_replace(["\\", "'"], ["\\\\", "\\'"], (string) $property['name']);
+            $lines = [
+                ...$lines,
+                ...$this->getAdditionalPropertiesExpectationLines(
+                    $property['sub_schema'],
+                    $spec,
+                    $target . "['" . $propertyName . "']"
+                ),
+            ];
+        }
+
+        return $lines;
+    }
+
+    protected function getAdditionalPropertiesExpectations(?string $model, array $spec, string $target): string
+    {
+        return implode("\n", array_map(
+            fn(string $line): string => '        ' . $line,
+            $this->getAdditionalPropertiesExpectationLines($model, $spec, $target)
+        ));
+    }
+
     /**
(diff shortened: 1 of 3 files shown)


ChiragAgg5k:
`@model_serializer(mode='wrap')` is the right shape — the old override could never work for nested models because pydantic-core does not call subclass Python methods when serializing a parent. Confirmed the nested case against a local reproducer of the new template logic:

```python
RowList.to_dict()
# {'total': 1.0, 'rows': [{'$id': 'b01', 'data': {'title': 'Dune', 'rating': 4.5}}]}   ✅
```

Worth noting the blast radius is wider than TablesDB, in case it helps the test plan: the same `with_data`/`_data` pattern is in `row.py`, `document.py`, `user.py`, `team.py`, `organization.py`, `preferences.py` and all five `*_list.py` siblings, so `list_documents`, `list_teams` and `list_organizations` were dropping payloads too.

Two things I think are worth settling before this merges.

### 1. Payload-only models gain an envelope the API does not have

The serializer sets `result['data']` unconditionally for anything with `additionalProperties`. That is right where the model also has declared fields to separate the payload from, and wrong where it has none. `Preferences` declares no fields:

```python
account_get_prefs()  -> {'data': {'theme': 'dark'}}
User.to_dict()       -> {'$id': 'u1', 'prefs': {'data': {'theme': 'dark'}}}
# REST returns          {'$id': 'u1', 'prefs': {'theme': 'dark'}}
```

The `data` key separates nothing here — preferences *are* the payload — and no other SDK wraps `prefs`. The double wrap is pre-existing for direct `getPrefs()` calls so it is not a regression, but this PR is the natural place to fix it: skip the envelope when the model has no declared fields, e.g. gate on `definition.properties` being empty in the template, or on `type(self).model_fields` at runtime.

### 2. The nested case is a silent behaviour change

Independent of (1): anyone reading `user['prefs']` gets `{}` today and `{'data': {...}}` after this change. Code that currently works around the empty dict will not start working — it will read the wrong key. Worth a changelog note whichever way (1) lands.

---

Context on why this matters beyond the SDK: this was the headline defect in a [recent MCP benchmark](https://6a79b8b74e1f6bf83f68.appwrite.network/) — `list_rows` returned row IDs with no column values, so reading one page of 10 rows cost 10 extra `get_row` calls. It was attributed there to MCP response serialization; it is this. We deliberately removed the client-side workaround from `appwrite/mcp` so the fix lands once, here.

PR #1779, opened 2026-08-12
fix(python): honour include and exclude in generic model serializers

ChiragAgg5k:
Follow-up to #1776/#1777, from review on appwrite/sdk-for-console-python#7.

Models with additional properties serialize their payload from a `@model_serializer(mode='wrap')`, because the payload lives in a private attribute rather than a declared field. The serializer wrote that key *after* pydantic had applied `include`/`exclude`, so field selection could not see it. On the published SDK:

```python
doc.model_dump(exclude={'data'})
# {'id': 'a', ..., 'data': {'title': 'hello', 'body': 'x'}}   <- exclude ignored

doc.model_dump(include={'id'})
# {'id': 'a', 'data': {'title': 'hello', 'body': 'x'}}        <- data not asked for
```

Nested selection could not reach the payload either: `include={'data': {'title'}}` returned the whole payload, and for a typed payload nothing was propagated into its own `model_dump`.

The serializer now resolves the selector against the synthetic key before writing it, and passes any nested selector down — to `model_dump` when the payload is a model, or by filtering keys when it is a plain dict. `SerializationInfo` exposes `include`/`exclude`, so this needs no override of `model_dump`.

For payload-only models such as `Preferences`, whose keys merge at the top level, the selector applies to those keys directly.

## After

```
plain             {'id': 'a', ..., 'data': {'title': 'hello', 'body': 'x'}}
exclude={'data'}  data absent
include={'id'}    {'id': 'a'}
include={'data'}  {'data': {'title': 'hello', 'body': 'x'}}
include={'data': {'title'}}   {'data': {'title': 'hello'}}
exclude={'data': {'title'}}   {'data': {'body': 'x'}}
exclude={'id'}    data still present
prefs include={'theme'}       {'theme': 'dark'}
```

Same results whether the payload is a plain dict or a typed model. `to_dict()` is unchanged.

## Verification

Regenerated the Console Python SDK and exercised every case above against the built package, for dict and typed payloads. The 1053 generated tests still pass, `composer lint-twig` is clean, and the only files that change in a regenerated SDK are the three generic models (`document`, `row`, `preferences`).

### DIFF
--- a/templates/python/package/models/model.py.twig
+++ b/templates/python/package/models/model.py.twig
@@ -2,7 +2,7 @@
 {% set isGeneric = definition.additionalProperties or hasGenericProperty %}
 from typing import Any, Dict, List, Optional, Union, cast{% if isGeneric %}, Generic, TypeVar, Type{% endif %}
 
-from pydantic import Field, PrivateAttr{{ definition.additionalProperties ? ', model_serializer' : '' }}
+from pydantic import Field, PrivateAttr{{ definition.additionalProperties ? ', TypeAdapter, model_serializer' : '' }}
 
 from .base_model import AppwriteModel
 {% set added = [] %}
@@ -30,6 +30,10 @@ from .{{ subSchema | caseSnake }} import {{ subSchema | caseUcfirst }}
 {% if isGeneric %}
 
 T = TypeVar('T')
+{% endif %}
+{% if definition.additionalProperties %}
+
+_PAYLOAD_ADAPTER = TypeAdapter(Dict[str, Any])
 {% endif %}
 
 class {{ definition.name | caseUcfirst }}(AppwriteModel{% if isGeneric %}, Generic[T]{% endif %}):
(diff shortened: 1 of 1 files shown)


ChiragAgg5k:
Valid — fixed in 903104d.

Pydantic does apply nested selection to `Dict[str, Any]` values, so filtering by key alone was both incomplete and inconsistent with the contract:

```python
M(d={'profile': {'name': 'n', 'secret': 's'}}).model_dump(include={'d': {'profile': {'name'}}})
# {'d': {'profile': {'name': 'n'}}}     <- pydantic recurses into a plain dict
```

Rather than hand-roll that traversal, the payload now dumps through a module-level `TypeAdapter(Dict[str, Any])`, which reuses pydantic's own selection — nesting at any depth, `__all__`, and the `by_alias`/`exclude_*` flags. The adapter runs only when a selector is present, so the default path is untouched.

```
include={'data': {'profile': {'name'}}}        {'profile': {'name': 'n'}}
exclude={'data': {'profile': {'secret'}}}      {'profile': {'name': 'n'}, 'tags': [...], 'other': 1}
include={'data': {'tags': {'__all__': {'a'}}}} {'tags': [{'a': 1}]}
```

Same results for a typed payload, which routes through the model's own `model_dump`. Plain `model_dump()` and `to_dict()` are unchanged, the 1053 generated tests pass, and `lint-twig` is clean.

PR #1783, opened 2026-08-13
Support portable OpenAPI generation profiles

ChiragAgg5k:
## What does this PR do?

Adds support for generating platform SDKs from one portable OpenAPI document plus a separate SDK generation profile.

```text
open-api3-1.9.x.json
sdk-generation-profile-1.9.x.json
              │
              └── OpenAPI3(spec, platform, profile) → existing generator input
```

Standard OpenAPI fields now work without a profile: services use `tags`, method names use `operationId`, authentication uses operation `security`, examples use `example`, enum names can use schema `title`, and OpenAPI 3.1 nullable schemas are normalized. Legacy platform specs and `x-appwrite` metadata remain supported.

`example.php` prefers the unified spec/profile pair and falls back to the legacy platform-specific URL for older releases.

## Test Plan

```text
composer refactor:check
composer lint
composer lint-twig
```

Additional compatibility validation:

- Built a portable OpenAPI document and profile from the current client, server, and console specs.
- Compared the complete normalized generator input for all three platforms.
- Generated and recursively compared Web client, Node server, Web console, and PHP server SDKs; no output differences.
- Validated the portable OpenAPI document with `swagger-cli`.

## Related PRs and Issues

- Follow-up to appwrite/appwrite#13201; that PR will be reworked to emit the portable document and profile after this lands.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -202,14 +202,17 @@ Pass as first argument to generate only that SDK:
 - **Silent no-op:** A new `.twig` file with no `getFiles()` entry — generation runs successfully but the file is never created
 - **Wrong scope:** Using `default` scope when you need `service` scope means your template can't access `{{ service.name }}`
 - **Copy scope surprises:** A `copy`-scoped file with Twig syntax — the syntax is output literally, not rendered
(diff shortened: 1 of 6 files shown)


ChiragAgg5k:
Closing per design feedback. This approach still projected the official OpenAPI document through a separate JSON generation profile and reconstructed legacy metadata internally. We will replace it with incremental native parsing of official OpenAPI attributes in SDK Generator, without converting the specification into another JSON representation.

PR #1786, opened 2026-08-13
Add Utopia OpenAPI adapter

ChiragAgg5k:
## What does this PR do?

Drafts SDK Generator's consumer boundary for the planned `utopia-php/openapi` contract.

```text
Utopia\OpenAPI\Specification
            ↓
      UtopiaAdapter
            ↓
existing Spec contract and templates
```

- Maps canonical metadata, tags, and operations into the existing `Spec` contract.
- Accepts the current parser as an optional compatibility delegate for methods, schemas, enums, and authentication that have not migrated yet.
- Keeps all Appwrite and SDK compatibility policy outside `utopia-php/openapi`.
- Uses contract fixtures because the Utopia package currently contains only its design plan and is not published yet.

This PR intentionally does not add the Composer dependency or switch generation entrypoints. Those changes follow after the Utopia package publishes the contract represented by the fixtures.

## Test plan

```text
composer refactor:check
composer lint
composer lint-twig
composer validate --no-check-lock
git diff --check
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,22 @@ The SDK Generator uses predefined language settings as [Twig templates](https://
 
 Both OpenAPI 3.0 and Swagger 2.0 specs are supported and produce identical SDKs, including services, methods, models, enums, and union types.
 
+A migration adapter is also available for the planned `utopia-php/openapi` canonical model. It lets SDK Generator consume `Utopia\OpenAPI\Specification` while delegating template fields that have not migrated yet to the existing parser:
+
+```php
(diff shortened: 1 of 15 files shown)


ChiragAgg5k:
Closing because the adapter still depends on the legacy Appwrite\\Spec\\Spec contract. The intended integration will have SDK Generator consume Utopia\\OpenAPI\\Specification directly, with generator-only policy and template projection kept separate from specification parsing.

PR #1838, opened 2026-08-24
fix: preserve JSON object-versus-array distinction in Web doc examples

ChiragAgg5k:
## What does this PR do?

Fixes a review finding on #1836: `getObjectExample` decoded spec examples with `json_decode($example, true)`, which erases the JSON object-versus-array distinction. A nested `"items": []` rendered as `items: {}` in the generated documentation, and an object with sequential numeric keys like `{"0": "a", "1": "b"}` rendered as the array `['a', 'b']`.

Examples are now decoded to `stdClass` so `renderJsValue` can tell JSON objects and arrays apart: empty arrays render as `[]`, empty objects as `{}`, and numeric-key objects keep their quoted keys.

## Test Plan

- Verified the helper directly: `{"items": []}` → `items: []`, `{"prefs": {}}` → `prefs: {}`, `{"0": "a", "1": "b"}` → quoted-key object, top-level arrays and non-JSON strings unchanged.
- Regenerated the Web SDK from both the client and console specs; `format:check` and `lint` pass on both.
- `composer lint` and `composer refactor:check` pass.

## Related PRs and Issues

- #1836

### DIFF
--- a/src/SDK/Language/Web.php
+++ b/src/SDK/Language/Web.php
@@ -10,6 +10,7 @@ use Utopia\OpenAPI\Model\Parameter;
 use Utopia\OpenAPI\Model\Schema;
 use Utopia\OpenAPI\Specification;
 use Override;
+use stdClass;
 use Twig\TwigFilter;
 
 class Web extends JS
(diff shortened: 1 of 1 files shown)


ChiragAgg5k:
Post-merge review of the #1836 finding (`json_decode($example, true)` collapsing `[]` vs `{}` and numeric-key objects).

LGTM for Web. Decoding to `stdClass` and splitting `renderJsValue` on `is_array` vs `stdClass` is the right model: JSON arrays stay PHP arrays, objects stay objects, empty `[]` / `{}` and `{"0":"a"}` survive, and top-level arrays / non-JSON strings are unchanged. The `allComposite` `stdClass` check is necessary so arrays of objects still break the way Prettier wants.

Not covered here, and worth a follow-up when Node/RN get the formatter-clean pass: both override `getParamExample` and still round-trip through `json_encode(json_decode(..., true))`, which turns nested `{}` into `[]` and list-like numeric-key objects into arrays. `getObjectExample` already exists on `Web` — those overrides should call it rather than re-decode. No unit test locks this helper (repo is e2e-only); a spec fixture with both `{"items":[]}` and `{"prefs":{}}` would stop this from regressing silently.

No revert/follow-up needed on the Web path itself.

PR #1839, opened 2026-08-25
fix(cli): resolve wildcard function template tags

ChiragAgg5k:
## Summary

Resolve wildcard starter-template versions returned by Appwrite before cloning function source.

Appwrite 1.9.6 returns starter metadata similar to:

```json
{
  "providerOwner": "appwrite",
  "providerRepositoryId": "templates",
  "providerVersion": "0.3.*",
  "runtimes": [
    {
      "name": "bun-1.3",
      "providerRootDirectory": "./bun/starter"
    }
  ]
}
```

The CLI treated `providerVersion` as a literal branch/tag and generated:

```sh
git clone --single-branch --depth 1 --branch 0.3.* --sparse \
  https://github.com/appwrite/templates .
```

`git clone --branch` only accepts a concrete ref. It does not expand tag patterns, so initialization failed with:

```text
Error: exit status 128
Cloning into '.'...
fatal: Remote branch 0.3.* not found in upstream origin
```

This also explains why searching the CLI or server source for a hard-coded `0.3.*` did not find the cause: the value comes from the `/functions/templates/starter` API response.

## Fix

For tag references containing a Git ref pattern (`*`, `?`, or `[`), the CLI now resolves the newest matching remote tag first:

```sh
git ls-remote --refs --tags --sort=-version:refname \
  https://github.com/appwrite/templates '0.3.*'
```

The current response is version-sorted as:

```text
4eb58cb2... refs/tags/0.3.1
7717e63c... refs/tags/0.3.0
```

The CLI extracts `0.3.1` and passes that concrete tag to `git clone --branch`. Concrete tags and branch references such as `main` retain their existing behavior.

The original wildcard remains in the template metadata used for deployment; resolution is limited to the local Git clone that requires a concrete ref.

## Tests

Added coverage for:

- parsing the Bun 1.3 starter response with `providerVersion: "0.3.*"`
- selecting the first version-sorted concrete tag
- returning an actionable error when no tag matches
- leaving concrete tag references unchanged without a remote lookup

### Live OSS verification

Reproduced with CLI 27.2.0 against `https://oss.appwrite.org/v1` (Appwrite 1.9.6). Before the change, `init function` reached `/functions/templates/starter` and failed exactly as reported:

```text
GET /v1/functions/templates/starter 200
fatal: Remote branch 0.3.* not found in upstream origin
```

Built the regenerated CLI and repeated the same flow against the same project. It resolved `0.3.*` to `0.3.1`, cloned the starter, and completed:

```text
GET /v1/functions/templates/starter 200
Success: Initialized function
```

The OSS project's runtime allowlist currently exposes Node 16, PHP 8, Ruby 3, and Python 3.9, so the end-to-end initialization used `node-16.0`. The failing clone path and starter-template version are shared across runtimes; the mocked API regression test specifically covers the `bun-1.3` metadata returned by Appwrite 1.9.6.

### Commands

```sh
php example.php cli
test -z "$(gofmt -l examples/cli)"
(cd examples/cli && go mod tidy && go build ./... && go vet ./... && go test ./...)
composer lint-twig
composer refactor:check
uvx djlint templates/cli/internal/cmd/initfunction.go \
  templates/cli/internal/cmd/initfunction_test.go \
  templates/cli/internal/cmd/initscaffold.go --lint
```

### DIFF
--- a/templates/cli/internal/cmd/initfunction.go
+++ b/templates/cli/internal/cmd/initfunction.go
@@ -575,8 +575,15 @@ func runInitFunction(command *cobra.Command, options initFunctionOptions) error 
 	repositoryURL := fmt.Sprintf("https://github.com/%s/%s", template.Owner, template.Repository)
 	clone := fmt.Sprintf("git clone --single-branch --depth 1 --sparse %s .", repositoryURL)
 	if template.Reference != "" {
+		reference := template.Reference
+		if template.ReferenceType == "tag" {
+			reference, err = resolveGitTag(templatesDir, repositoryURL, reference)
+			if err != nil {
+				return err
+			}
+		}
 		clone = fmt.Sprintf("git clone --single-branch --depth 1 --branch %s --sparse %s .",
-			template.Reference, repositoryURL)
+			reference, repositoryURL)
 	}
 	if err := runGit(templatesDir, clone); err != nil {
 		return err
--- a/templates/cli/internal/cmd/initfunction.go
+++ b/templates/cli/internal/cmd/initfunction.go
@@ -573,7 +573,7 @@ func runInitFunction(command *cobra.Command, options initFunctionOptions) error 
 	defer os.RemoveAll(templatesDir)
 
 	repositoryURL := fmt.Sprintf("https://github.com/%s/%s", template.Owner, template.Repository)
-	clone := fmt.Sprintf("git clone --single-branch --depth 1 --sparse %s .", repositoryURL)
+	cloneArguments := []string{"clone", "--single-branch", "--depth", "1"}
 	if template.Reference != "" {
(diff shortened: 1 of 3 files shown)


ChiragAgg5k:
Looks good — this is the right fix for 1.9.6 starter metadata.

`git clone --branch` cannot expand `0.3.*`. Resolving with `git ls-remote --refs --tags --sort=-version:refname` via argv (not `runGit`) is the Windows-safe approach, and leaving the wildcard on `templateReference` for the server seed is correct.

CLI Go tests and Twig/lint are green. Optional follow-up: quote `--branch %q` on the clone line, and later reuse `resolveGitTag` in `init site` so that path can drop `tail`/`awk`.

Approving.

Issue #1813, opened 2026-08-19
Generate formatter-clean SDKs directly from templates


ChiragAgg5k:
## Goal

Every SDK generated by this repository should already match its ecosystem's canonical formatter.

```text
before: generate -> run formatter -> commit formatting changes
after:  generate ----------------> formatter-clean SDK
```

The fix belongs in the Twig templates. Do not add a post-generation formatting step, format `examples/` in place, or rely on a workflow in the generated SDK to rewrite files after generation. Generated output is the product, and the templates should be the source of truth for its layout.

Reference implementations:

- Go and CLI: #1800
- Dart and Flutter: #1809
- Python: #1825
- PHP: #1824
- Rust: #1834
- Swift and Apple: #1835
- Kotlin and Android: #1837
- Ruby: #1840
- .NET and Unity: #1842
- JavaScript / TypeScript: #1836, #1843

### Completed

- [x] **Go / CLI** — #1800
- [x] **Dart / Flutter** — #1809
- [x] **Python** — #1825
- [x] **PHP** — #1824
- [x] **Rust** — #1834
- [x] **Swift / Apple** — #1835
- [x] **Kotlin / Android** — #1837
- [x] **JavaScript / TypeScript** — `web`, `node`, `react-native` — #1836, #1843
- [x] **Ruby** — `ruby` — #1840
- [x] **C#** — `dotnet`, `unity` — #1842

### Claimed

_None._

### Available

The related targets are grouped because they share templates or inherit from one another. A PR that changes a parent must regenerate and verify its children too.

_None — all tracked SDK families are complete._

REST and GraphQL are documentation-only targets, so they are outside the source-formatting scope. The skills and editor/agent plugin generators are tooling targets rather than SDKs.

## How to contribute

Comment with the target or family you want to take before starting so we can mark it as claimed. Prefer one checklist item per PR; large families may be split after coordinating here.

For each target:

1. Remove the existing generated output and regenerate it with `php example.php <sdk>`.
2. Run the ecosystem's canonical formatter in **check-only** mode and record the initial failures.
3. Fix the relevant Twig templates so a clean generation passes that same check without changing any generated file.
4. Add or extend automated validation so future formatting regressions fail CI.
5. Regenerate and inspect every affected child SDK when inheritance is involved.
6. Run `composer refactor:check` and `composer lint-twig`.

[...]

PR #1843, opened 2026-08-25
Generate Prettier-clean Node and React Native SDK files

harsh-sikarwar:
## Problem

The generated Node and React Native SDKs did not match Prettier's output, and
neither shipped the ESLint and `tsc` gates the Web SDK gained in #1836.

```text
before: generate -> run formatter -> commit formatting changes
after:  generate ----------------> formatter-clean SDK
```

Completes the **JavaScript / TypeScript** checklist item in #1813. Follows
#1836 (`web`), which introduced the shared helpers and CI gates this builds on.

| Target | Failing before | After |
|---|---|---|
| `node` | 749 / 854 | **0** |
| `react-native` | 219 / 227 | **0** |

## Changes

- add `.prettierrc`, `.prettierignore`, and `eslint.config.mjs` to both SDKs and
  register them in `getFiles()`
- add `prettier`, `eslint`, and `typescript-eslint` devDependencies plus
  `format`, `format:check`, `lint`, `lint:fix`, and `analyse` scripts, and
  regenerate both lock file templates with `./scripts/update-lockfiles.sh`
- update the Node and React Native Twig templates so generated services, models,
  enums, client, tests, docs examples, and configuration are Prettier-clean and
  lint-clean on a clean generation
- run `format:check`, `lint`, and `analyse` for both targets in CI, and document
  the child-target commands in AGENTS.md

## Deduplicating the example renderer

`Node` and `ReactNative` each carried their own copy of `getParamExample()`,
which still had the defects fixed in `Web` during #1836 and #1838: raw string
interpolation with no escaping, and `json_decode($example, true)` collapsing the
JSON object-versus-array distinction.

The three copies differed only in the file example, so that one difference moves
into an overridable `getFileExample()` hook and both duplicates are deleted.
Every JavaScript target now renders examples through the same audited path
rather than drifting apart again.

The same consolidation applies to the Node test templates, which interpolated
spec examples directly into single-quoted strings — they now route through the
shared escaper via a `jsString` filter.

## One shared Web template changed

`templates/web/src/services/realtime.ts.twig` declares `queryParams` as `let`
for the targets that append a JWT to it and `const` for React Native, which does
not. Without this the shared template emits a never-reassigned binding that
`prefer-const` rejects. Web output is unchanged; it was regenerated and verified.

## Note on `formatArgumentList`

The single-argument branch uses an argument-length threshold that was fitted
empirically against Prettier's output rather than derived from its layout
algorithm. It is correct for every call site across all three SDKs and all three
platforms, but it is an approximation and worth a look in review.

## Tests

Verified on all three platforms `example.php` accepts, not only the ones CI runs:

```text
php example.php node <client|console|server>
cd examples/node && npm ci && npm run format:check && npm run lint \
  && npm run analyse && npm run build && npm run test        # 828/828 pass

php example.php react-native <client|console|server>
cd examples/react-native && npm ci --omit=peer && npm run format:check \
  && npm run lint && npm run analyse && npm run build

php example.php web <client|console>                          # regression check
cd examples/web && npm ci && npm run format:check && npm run lint \
  && npm run analyse && npm run build

composer lint
composer lint-twig
composer refactor:check
./.github/scripts/max-line-length.sh . 1200 "*.twig"
```

All green. `npm run format:check` reports zero files on a clean generation for
both targets.


src/SDK/Language/JS.php
```diff
@@ -273,6 +273,55 @@ protected function formatClientChain(array $calls): string
...
+        $lead = $prefixWidth > 0 ? ' ' : '';
+
+        if ($arguments === [] || mb_strlen($oneLine) <= self::PRINT_WIDTH) {
+            return $lead . $callee . '(' . implode(', ', $arguments) . ')' . $suffix;
+        }
+
+        // A single short argument does not earn its own line: Prettier breaks
+        // after the assignment instead and leaves the call intact. It only
+        // explodes the list once an argument is long enough to be worth it.
+        if ($prefixWidth > 0 && count($arguments) === 1 && mb_strlen($arguments[0]) <= 20) {
```
ChiragAgg5k: This `<= 20` is the empirical bit. It only runs for assignment-RHS calls (`prefixWidth > 0`) with a single argument, which matches how Prettier prefers to break after `=` rather than explode a short call.

Fine to ship given it matches every current call site across web/node/RN. I'd hoist `20` to a named constant (same family as `PRINT_WIDTH`) so the next method that misses the threshold isn't a mystery. Not a merge blocker.

ChiragAgg5k (commented):
Went through the Node/RN formatter work, the shared `getFileExample()` split, and `formatArgumentList` specifically.

Generate jobs for node, react-native, and web are green (`format:check` / `lint` / `analyse` / build, plus node tests). The Greptile InputFile finding is fixed: RN examples now emit the URI record instead of `InputFile.fromPath`.

`formatArgumentList` note is on the magic `20` below. Non-blocking otherwise, LGTM.

src/SDK/Language/JS.php
harsh-sikarwar: Done in aba6fd0ab — hoisted to `HUGGABLE_ARGUMENT_WIDTH` next to `PRINT_WIDTH`, with a docblock saying it is fitted to Prettier's observed output rather than derived from its layout algorithm, so a method that later lands near the cutoff is traceable.

Confirmed it is a pure rename: regenerated node and react-native before and after and diffed the output, byte-identical. All gates still green on both targets.

PR #1849, opened 2026-08-27
fix(python): stop breaking short calls across lines

ChiragAgg5k:
## What does this PR do?

The Python templates emitted `Field(...)` and `self._normalize_value(...)` pre-exploded with a magic trailing comma:

```python
    id: str = Field(
        ...,
        alias='$id',
    )
```

```python
        if queries is not None:
            api_params['queries'] = self._normalize_value(
                queries,
            )
```

Black only breaks a call that does not fit the configured line length, and the magic trailing comma then pins it open — so every one-argument model field and query parameter in every generated SDK was split across four lines for no reason. The published diff for a routine release is dominated by that reflow.

The templates now render the call on one line and a new `formatCall` filter explodes the arguments only when the line is longer than Black's configured length (`line-length = 120`, set in `templates/python/pyproject.toml.twig`), so the output stays Black-clean in both directions:

```python
    id: str = Field(..., alias='$id')
    app_vcs_providers_with_repository_creation: List[Any] = Field(
        ...,
        alias='_APP_VCS_PROVIDERS_WITH_REPOSITORY_CREATION',
    )
```

Rule 7 in `AGENTS.md` still holds: the fix is in the templates, with no post-generation formatting step.

## Test Plan

```bash
rm -rf examples/python && php example.php python console
(cd examples/python && find appwrite test -name '*.py' -print0 | xargs -0 python -m black --check)   # 606 files unchanged
(cd examples/python && python -m black --check setup.py)                                             # 1 file unchanged
(cd examples/python && python -m compileall -q appwrite)
composer lint && composer lint-twig && composer refactor:check
```

Verified on the generated console SDK that short calls collapse, the 14 call sites that genuinely exceed 120 columns still explode, and no `Field(` or `_normalize_value(` line exceeds the limit.

## Related PRs and Issues

Noticed while releasing the Console Python SDK 0.6.0 (appwrite/sdk-for-console-python#8), where the reflow accounted for most of the diff.

## Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?

Yes.

### DIFF
--- a/src/SDK/Language/Python.php
+++ b/src/SDK/Language/Python.php
@@ -18,6 +18,12 @@ use Twig\TwigFilter;
 
 class Python extends Language
 {
+    /**
+     * Black's line length for generated SDKs, kept in step with the `[tool.black]`
+     * block in `templates/python/pyproject.toml.twig`.
+     */
+    public const int LINE_LENGTH = 120;
+
     #[Override]
     protected $params = [
         'pipPackage' => 'packageName',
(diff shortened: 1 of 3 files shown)


ChiragAgg5k:
Fixed in ff86542 — the finding was real. Reproduced against the pre-fix code with `alias='name(legacy'` on a declaration past 120 columns:

```python
    weird_and_long...: Optional[str] = Field(default=None, alias='name(
        legacy',
    )
```

The backward scan matched the parenthesis inside the string literal. Both the delimiter matcher and the argument splitter now share a single `quotedPositions` mask and skip anything inside a string, and a line whose last character is itself quoted is left alone. Verified with `alias='name(legacy'` and `alias='a)b(c'` at both short and long widths; regenerated console SDK stays Black-clean (606 files) and `compileall` passes.

PR #1850, opened 2026-08-27
fix(python): alias a model that collides with a service enum

ChiragAgg5k:
## What does this PR do?

`appwrite_console/services/organizations.py` imported `Addon` twice — the response model, then the enum `get_addon_price` takes:

```python
from ..models.addon import Addon
from ..enums.addon import Addon      # shadows the model
```

The second import wins, so four methods passed the enum to `_parse_response()`:

```python
    def get_addon(self, organization_id: str, addon_id: str) -> Addon:
        ...
        return self._parse_response(response, model=Addon)   # AttributeError
```

`AttributeError: type object 'Addon' has no attribute 'model_validate'` on `get_addon`, `create_baa_addon`, `create_premium_geo_db_addon` and `confirm_addon_payment`.

The templates already alias a model whose name collides with its own service (`Project` → `ProjectModel`). This extends that rule to a collision with an enum the same service imports, and moves the decision into one `getServiceModelName` filter — the `(model | caseUcfirst) == (service.name | caseUcfirst)` comparison was duplicated across ten template sites (imports, return annotations, docstring `Returns` blocks, `model=` arguments, `Union[...]` members), which is why the enum case was missed in all of them at once.

After:

```python
from ..models.addon import Addon as AddonModel
from ..enums.addon import Addon

    def get_addon(self, organization_id: str, addon_id: str) -> AddonModel:
        return self._parse_response(response, model=AddonModel)

    def get_addon_price(self, organization_id: str, addon: Addon) -> AddonPrice:
```

## Test Plan

```bash
rm -rf examples/python && php example.php python console
(cd examples/python && find appwrite test -name '*.py' -print0 | xargs -0 python -m black --check)   # 606 files unchanged
(cd examples/python && python -m black --check setup.py && python -m compileall -q appwrite)
composer lint && composer lint-twig && composer refactor:check
php example.php python server    # other platforms still generate
```

Asserted on the generated console SDK:

```
module Addon        -> <enum 'Addon'>
module AddonModel   -> <class 'appwrite.models.addon.Addon'>  model_validate: True
get_addon_price addon annotation -> <enum 'Addon'>
get_addon return                 -> <class 'appwrite.models.addon.Addon'>
```

The pre-existing service-name collision is unchanged: `project.py` still imports `Project as ProjectModel` and annotates `-> ProjectModel`. Docstring `Returns` blocks track the annotations, so the two cannot drift.

## Related PRs and Issues

Found reviewing the regenerated Console Python SDK in appwrite/sdk-for-console-python#8. Follows #1849.

## Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?

Yes.

### DIFF
--- a/src/SDK/Language/Python.php
+++ b/src/SDK/Language/Python.php
@@ -527,9 +527,10 @@ class Python extends Language
      *
      * Must stay identical to the `-> ...` annotation the service template
      * emits: a multi-model response is a `Union[...]`, and a model whose name
-     * collides with its own service is imported under a `Model` suffix.
+     * collides with its own service, or with an enum that service imports, is
+     * imported under a `Model` suffix.
      */
-    protected function getResponseType(Operation $method, string $serviceName = ''): string
+    protected function getResponseType(Operation $method, ?Tag $service = null, ?Specification $spec = null): string
     {
         $models = \array_filter(
             $this->getOperationResponseModels($method),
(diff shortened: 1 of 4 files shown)


ChiragAgg5k:
Closing in favour of fixing this at the source: the collision was one explicit enum name in the cloud spec, renamed `Addon` -> `AddonKey` so the generated module imports `Addon` (model) and `AddonKey` (enum) with no shadowing, and every language benefits rather than Python alone.

The general problem is still worth solving in the generator — a spec is free to name an enum after a model again, and Python is the language where that silently produces working-looking code instead of a compile error. Opening an issue proposing the pattern; this branch is kept if we want the implementation.

PR #1863, opened 2026-08-31
fix: quote CLI string examples with whitespace

ChiragAgg5k:
## Summary

- quote generated CLI string examples when they contain shell-unsafe characters
- escape embedded single quotes using POSIX shell-safe quoting
- fix generated commands such as `--default Hello World` and values such as `don't`

Fixes the Greptile finding on appwrite/sdk-for-cli#362.

## Validation

- `php example.php cli`
- `composer refactor:check`
- `composer lint`
- `composer lint-twig`
- `test -z "$(gofmt -l examples/cli)"`
- `git diff --check`


ChiragAgg5k:
@greptile review

PR #1866, opened 2026-09-02
fix: prevent empty GraphQL SDK examples

ChiragAgg5k:
## Summary

- render standard GraphQL examples independently of request headers
- keep multipart rendering selected by the operation content type
- validate generated client and server GraphQL examples in CI

## Verification

- `php example.php graphql client`
- `php example.php graphql server`
- generated 710 server examples with zero missing operations
- `composer lint-twig`
- `composer refactor:check`

### DIFF
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -70,6 +70,13 @@ jobs:
           - sdk: rust
             platform: server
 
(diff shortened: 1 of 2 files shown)


ChiragAgg5k:
Closing at requester’s direction.

PR #1877, opened 2026-09-05
Resolve request model parameters through schema references

ChiragAgg5k:
## Summary

- replace the non-standard `x-model` extension in the e2e OpenAPI fixture with a standard `$ref` to the request model schema
- remove the `x-model` fallbacks from the shared schema model helpers so model resolution only follows `$ref`, `items`, and composite members

The Appwrite OpenAPI producer never emitted `x-model` (verified against the published 2.0.x client, server, and console specs), so this only affects the fixture and the generator's legacy fallback.

## Output change

Go, PHP, and every array-typed request model parameter already resolved to the request model type. Web, Node, React Native, and Python typed object-valued request model parameters as `object` / `Dict[str, Any]` because their type resolution only followed `$ref`. With the fixture using a reference, those four now emit `Models.Player` / `Player`, matching the other languages and the array variants. Request bodies are unchanged. All other generated files are identical.

Related to [CLO-4377](https://linear.app/appwrite/issue/CLO-4377/openapi3-standards).

## Validation

- `composer lint`
- `composer refactor:check`
- generated every SDK from the fixture for console, server, and client before and after; only the Web, Node, React Native, and Python service files for the two request model methods differ, as described above
- Web, Node, and React Native: `npm run lint` and `npm run analyse` pass; the only Prettier warnings are pre-existing in the baseline output for the same fixture (`(Models.Player)[]` parenthesised arrays), tracked separately
- Python: `black --check` passes

### DIFF
--- a/src/SDK/Language.php
+++ b/src/SDK/Language.php
@@ -401,16 +401,8 @@ abstract class Language
 
     protected function getSchemaModel(Schema|Parameter $value): ?string
     {
-        $schema = $this->getSchema($value);
-        $models = $this->getSchemaModels($schema);
-        if (\count($models) === 1) {
-            return $models[0];
-        }
-        if ($schema instanceof ArraySchema) {
-            return $schema->items->extensions['x-model'] ?? $schema->extensions['x-model'] ?? null;
-        }
-
-        return $schema->extensions['x-model'] ?? null;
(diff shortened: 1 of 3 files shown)


ChiragAgg5k:
@greptile review

PR #1881, opened 2026-09-07
refactor: extension key enums and platform-keyed auth

ChiragAgg5k:
## Summary

Follow-up to #1880, preparing the generator for one canonical OpenAPI document.

- **Extension keys are enums.** `Appwrite\SDK\Extension` holds every top-level vendor extension the generator reads (`x-appwrite`, `x-request-model`, `x-mapping`, and the generator's own `x-sdk-source` / `x-sdk-config` annotations); `Appwrite\SDK\Extension\Appwrite` holds the keys read under `x-appwrite` (`platforms`, `methods`, `auth`, `produces`, `packaging`, `location`, `param`, `config`, `optional`). Every string read in `SDK.php`, `Language.php`, `CLI.php` and `Rust.php` goes through them, so the full set the generator depends on is visible in two files.
- **`x-appwrite.auth` may be keyed by platform.** A per-platform document lists the example credentials flat; the canonical document has to carry all three, so it keys them by platform and the generator reads `auth[platform]`, falling back to the flat form. No behaviour change for existing documents.

```diff
 "x-appwrite": {
-  "auth": { "Project": [], "Session": [] }        // one per document today
+  "auth": {
+    "client":  { "Project": [] },
+    "server":  { "Project": [], "Session": [] },
+    "console": { "Project": [] }
+  }
 }
```

- The `x-oneOf` / `x-anyOf` fallback appears in no fixture or published spec and is removed. The fixture-only `x-appwrite.weight` is dropped from the fixture.

## Tests

- Fixture gains a platform-auth operation with keyed `auth` and a keyed `auth` on the server alias variant.
- `testExampleCredentialsFollowPlatformAuth()` asserts the exact setters in those examples on both platforms: server shows `setKey()` for the operation and `setJWT()` for the alias, client shows `setProject()` alone.

## Validation

- `vendor/bin/phpunit --testsuite Generation`: 82 tests, 1115 assertions.
- `composer lint`, `composer refactor:check`: clean.

Related to [CLO-4377](https://linear.app/appwrite/issue/CLO-4377/openapi3-standards).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -284,6 +284,7 @@ public function getFiles(): array
 | Which files get generated | `src/SDK/Language/<Lang>.php` → `getFiles()` |
 | Type mappings for a language | `src/SDK/Language/<Lang>.php` → `getTypeName()` |
 | Available Twig filters | `src/SDK/SDK.php` (around line 62) |
+| Vendor extensions the generator reads | `src/SDK/Extension.php` and `src/SDK/Extension/Appwrite.php`; add a case before reading a new key |
 | Canonical spec parser and DTOs | `utopia-php/openapi` (VCS dependency) |
 | SDK grouping and filtering | `src/SDK/SDK.php` |
 | Generation orchestration | `src/SDK/SDK.php` → `generate()` |
--- a/src/SDK/Extension.php
+++ b/src/SDK/Extension.php
@@ -1,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Appwrite\SDK;
+
+/**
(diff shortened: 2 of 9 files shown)


ChiragAgg5k:
The two findings describe the previous design of this PR. Scope changed: the canonical document carries `x-appwrite.auth` keyed by platform (see appwrite/appwrite Specs task change), so the generator reads `auth[platform]` and never derives credentials from the security requirement, and aliases carry their own keyed `auth` rather than a separate `security`. Both behaviours are covered by `testExampleCredentialsFollowPlatformAuth()`.

@greptile review

PR #1913, opened 2026-09-17
fix(web): declare a Node floor the lint toolchain can actually run on

ChiragAgg5k:
The Web and React Native templates declare `"node": ">=18.0.0"` in `engines` while their `devDependencies` pull `@eslint/js` 10.0.1 and `eslint` 10.10.0, both of which declare `"node": "^20.19.0 || ^22.13.0 || >=24"`. Every SDK generated from those templates inherits the contradiction.

What it costs a contributor working on a generated SDK repo at the advertised Node version: engine warnings on install, a hard failure under `engine-strict`, and no way to run `npm run lint` or `npm run lint:fix` at all. Consumers are unaffected — npm does not apply devDependency engines to installers of a package — so this is a contributor-facing inconsistency rather than a runtime one.

## The fix

`engines.node` becomes `^20.19.0 || ^22.13.0 || >=24` in both templates, matching what the lint toolchain accepts.

The other direction would be to hold eslint at a release that still supports Node 18. That trades a current toolchain for a promise nobody is keeping: Node 18 reached end of life in April 2025, and 20.19 is the lowest version eslint 10 will run on. This is the smallest range that makes the declaration true.

Only these two templates declare `engines`. The Node template pulls the same eslint 10 without one, so nothing there changes.

## Verification

Generated the Console Web SDK against a patched generator and parsed the result: valid JSON, `engines` renders as `{"node": "^20.19.0 || ^22.13.0 || >=24"}`, and the rest of the package — including the `impersonateuserid` URL builders restored in #1912 — is byte-identical to a stock 5.0.2 generation.

Found by Greptile on appwrite/sdk-for-console#112, where the mismatch first shipped into a generated SDK. That PR is the 16.1.0 Console release; it is not blocked on this, so the corrected `engines` will land in whichever release follows.


ChiragAgg5k:
Closing this — not worth the trade.

The mismatch is real but contributor-facing only: npm does not apply devDependency engines to anyone installing a generated SDK, so nothing reaches consumers. Raising `engines.node` to satisfy the lint toolchain, on the other hand, narrows the support the published packages advertise, which is a bigger change than the problem justifies and reads as a regression rather than a fix.

If this is picked up again, the version that does not touch the consumer contract is holding the lint dependencies at a release that still runs on the advertised floor, rather than moving the floor to match the tooling.

